PR #568, opened 2022-10-27
feat: add `DtPresence`

modialpad:
# Add presence

Add the new `DtPresence` component.

- [x] Feature
- [x] Documentation

## :book: Description

Add the new `DtPresence` component that consumes `d-presence` styles. Adds the `presence`, `theme`, and `srText` props to customize how the presence circle is rendered.

## :bulb: Context

Follows work on adding Presence styles in Dialtone here: https://github.com/dialpad/dialtone/pull/716

## :pencil: Checklist

- [x] I have added tests

## :crystal_ball: Next Steps

Integrating `DtPresence` component as part of `DtAvatar`.

image

## :link: Sources

https://dialpad.design/components/presence.html

### DIFF
--- a/components/presence/presence.mdx
+++ b/components/presence/presence.mdx
@@ -1,0 +1,83 @@
+import { Canvas, Story, Subtitle, ArgsTable, PRIMARY_STORY } from '@storybook/addon-docs/blocks';
+import Presence from './presence.vue';
+
+# Presence
+
+<Subtitle>
+    This component displays a presence circle indicating the current status.
+</Subtitle>
+
+## Base Style
+
+Presence can be initialized by passing `presence` prop that can take one of 4 values: 'away', 'active', 'offline', and 'busy'.
+
+<Canvas>
+  <Story id="components-presence--default" />
+</Canvas>
+
+## Variants
+
+<Canvas>
+  <Story id="components-presence--variants" />
+</Canvas>
+
+## Props, Slots & Events
+
+<ArgsTable story={PRIMARY_STORY} />
+
+## Usage
+
+### Import
+
+```jsx
+import { DtPresence } from '@dialpad/dialtone-vue';
+```
+
+### With Presence Prop
+
+```html
+<dt-presence
+  presence="active"
+/>
+```
+
+### With SR text prop
+
+```html
+<dt-presence
+  presence="active"
+  sr-text="User Presence"
+/>
+```
+
+### With Theme prop
+
+```html
+<dt-presence
+  theme="default"
+/>
+```
+
+### Accessibility
+
+Keyboard shortcuts should be visible to sighted users and made available to assistive technology.
+The Presence component is purely visual by default, and will not read out to a screen reader.
+
+If you'd like for a SR to read out any changes to Presence, you should pass the `srText` prop so it is
+read by AT.
+
+Example:
+
+```html
+<dt-presence
+  presence="active"
+  sr-text="User {{ user }} is active"
+/>
+```
+
+Abbreviations / symbols should be read out in full for voiceover / screen readers.
+
+### References
+
+- [Keyboard Symbols](http://xahlee.info/comp/unicode_computing_symbols.html)
+- [aria-keyshortcuts](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-keyshortcuts)
--- a/components/presence/presence.mdx
+++ b/components/presence/presence.mdx
@@ -64,7 +64,9 @@ import { DtPresence } from '@dialpad/dialtone-vue';
 The Presence component is purely visual by default, and will not be read out to a screen reader.
 
 If you'd like for a SR to read out any changes to Presence, you should pass the `srText` prop so it is
-read by AT. The component has a role of "status" to assist SR apps in reading out status changes.
+read by AT and set the `aria-live` attribute to either 'polite' or 'assertive'.
+Even though the component has a role of "status" to assist SR apps in reading out status changes, its default
+'aria-live' value is set to 'off'.
 
 [See W3C guidelines](https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA22)
 for more information.
--- a/components/presence/presence.mdx
+++ b/components/presence/presence.mdx
@@ -10,6 +10,7 @@ import Presence from './presence.vue';
 ## Base Style
 
 Presence can be initialized by passing `presence` prop that can take one of 4 values: 'away', 'active', 'offline', and 'busy'.
+By default, it's set to 'active'.
 
 <Canvas>
   <Story id="components-presence--default" />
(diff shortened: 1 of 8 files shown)


components/presence/presence.vue
```diff
@@ -0,0 +1,75 @@
...
+      default: PRESENCE_STATES.ACTIVE,
+      validator: (role) => {
+        return PRESENCE_STATES_LIST.includes(role);
+      },
+    },
+
+    /**
+     * A theme prop to customize the color of the outer border of presence
+     */
+    theme: {
```
modialpad: Need advice on this - I ended up adding a `theme` prop to customize the color of that border, with theme names matching the keys of `BACKGROUND_COLOR_FOR_THEME`. Is this the best approach to make the border themable?

components/presence/presence.vue
```diff
@@ -0,0 +1,75 @@
+<template>
+  <div
+    v-if="presence"
```
modialpad: I made `presence` take a default, we probably don't need `v-if` in this case, right?

components/presence/presence.mdx
```diff
@@ -0,0 +1,83 @@
...
+
+# Presence
+
+<Subtitle>
+    This component displays a presence circle indicating the current status.
+</Subtitle>
+
+## Base Style
+
+Presence can be initialized by passing `presence` prop that can take one of 4 values: 'away', 'active', 'offline', and 'busy'.
```
braddialpad, line 12: Does it default to one?

components/presence/presence.mdx
```diff
@@ -0,0 +1,83 @@
...
+  sr-text="User {{ user }} is active"
+/>
+```
+
+Abbreviations / symbols should be read out in full for voiceover / screen readers.
+
+### References
+
+- [Keyboard Symbols](http://xahlee.info/comp/unicode_computing_symbols.html)
+- [aria-keyshortcuts](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-keyshortcuts)
```
braddialpad: Is `aria-keyshortcuts` relevant here? how would it be used with presence?

components/presence/presence.vue
braddialpad: no, since it will always be true.

components/presence/presence_default.story.vue
```diff
@@ -0,0 +1,14 @@
+<template>
+  <dt-presence
+    presence="active"
```
braddialpad: the default story needs to be modifiable by storybook controls which means you have to reference the built in properties and have to include every prop/slot

```
:presence="presence"
:theme="theme"
:srText="srText"
```

components/presence/presence.vue
```diff
@@ -0,0 +1,75 @@
...
+    :class="['d-presence', BACKGROUND_COLOR_FOR_THEME[theme]]"
+    data-qa="dt-presence"
+  >
+    <span
+      v-if="srText"
+      data-qa="dt-presence-sr-text"
+      class="sr-only"
+    >{{ srText }} </span>
```
braddialpad, line 12: Should this optionally be aria-live="polite" in case we want to live announce a change in status? Not sure the needs of product, maybe it's not necessary.

components/presence/presence.stories.js
```diff
@@ -0,0 +1,67 @@
...
+  },
+};
+
+// Story Collection
+export default {
+  title: 'Components/Presence',
+  component: Presence,
+  argTypes: argTypesData,
+  excludeStories: /.*Data$/,
+  parameters: {
```
braddialpad, line 33: add within parameters
```
controls: {
  sort: 'requiredFirst',
},
```

braddialpad (approved):

braddialpad (commented):
accidentally hit approve, meant to hit comment 😆

components/presence/presence.mdx
modialpad: It does default to 'active'

components/presence/presence.vue
modialpad: Good point, I missed this but my thinking originally to add the `role=status` attribute to the root element, which will implicitly add `aria-live: polite` to its child elements: https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA22

components/presence/presence.mdx
modialpad: You're right, we don't need this. I updated this section to read the following:

```md
The Presence component is purely visual by default, and will not be read out to a screen reader.

If you'd like for a SR to read out any changes to Presence, you should pass the `srText` prop so it is
read by AT. The component has a role of "status" to assist SR apps in reading out status changes.

[See W3C guidelines](https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA22)
for more information.
```

components/presence/presence_default.story.vue
modialpad: Added

components/presence/presence.stories.js
modialpad: Added

components/presence/presence.mdx
braddialpad: Alright cool, maybe just mention that here.

components/presence/presence.mdx
braddialpad: great

components/presence/presence.vue
braddialpad: Sounds good but we should probably also include a way to disable this. I'm thinking if you have 100 contacts in your contacts list with a presence indicator and every time their presence updates it notifies you it would be really annoying.

braddialpad (changes requested):
Just add a way to disable the aria-live

components/presence/presence.vue
modialpad: @braddialpad do you think we still need this `theme` variable? Looks like I can remove it

braddialpad: I think you can remove it, just confirm that `border-color: var(--color-border-base);` in dialtone is correct

components/presence/presence.vue
modialpad: I added a way to set `aria-live` directly on the element by parent. 

```vue
  <div
    class="d-presence"
    data-qa="dt-presence"
    role="status"
    :aria-live="$attrs.ariaLive || 'polite'"

```

braddialpad: Hmm the more I think about this the more I think it should just be off by default. In which case we actually don't need to set this property at all since it can be set via attrs. Lets remove it and just mention that you can optionally do it in the accessibility section of the docs. Sorry for the confusion.

braddialpad (approved):
Just fix the one aria-live thing I mentioned and you're good to merge.

components/presence/presence.vue
modialpad: Set it to default to 'off', updated tests & docs.

PR #570, opened 2022-10-28
fix: update event handling

josedialpad:
# update event handling

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Vue3 version of https://github.com/dialpad/dialtone-vue/pull/562

## :pencil: Checklist

- [ ] I have added tests

### DIFF
--- a/components/button/button.stories.js
+++ b/components/button/button.stories.js
@@ -118,12 +118,6 @@ export const argTypesData = {
     },
   },
 
-  click: {
-    description: 'Native button click event',
-    table: {
-      type: { summary: 'event' },
-    },
-  },
   focusin: {
     description: 'Native button focusin event',
     table: {
--- a/components/modal/modal.vue
+++ b/components/modal/modal.vue
@@ -10,10 +10,7 @@
     ]"
     data-qa="dt-modal"
     :aria-hidden="open"
-    @click.self="close"
-    @keydown.esc="close"
-    @keydown.tab="trapFocus"
-    @after-enter.self="setFocusAfterTransition"
+    v-on="modalListeners"
   >
     <div
       v-if="show && ($slots.banner || bannerTitle)"
@@ -115,6 +112,7 @@ import Modal from '@/common/mixins/modal.js';
 import { MODAL_KIND_MODIFIERS, MODAL_SIZE_MODIFIERS } from './modal_constants';
 import { getUniqueString } from '@/common/utils';
 import DtLazyShow from '../lazy_show/lazy_show';
+import { EVENT_KEYNAMES } from '@/common/constants';
 import SrOnlyCloseButtonMixin from '@/common/mixins/sr_only_close_button';
 import SrOnlyCloseButton from '@/common/sr_only_close_button';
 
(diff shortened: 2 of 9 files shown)


components/popover/popover.vue
```diff
@@ -57,9 +58,8 @@
         }"
         :css="$attrs.css"
         :tabindex="contentTabindex"
-        @keydown.capture="onKeydown"
-        @after-leave="onLeaveTransitionComplete"
-        @after-enter="onEnterTransitionComplete"
+        v-bind="$attrs"
```
juliodialpad: Doing v-bind="$attrs" multiple times within the same component will duplicate classes in both elements, which might break styling or firing multiple events.

components/button/button.stories.js
```diff
@@ -118,12 +118,6 @@ export const argTypesData = {
     },
   },
 
-  click: {
-    description: 'Native button click event',
-    table: {
-      type: { summary: 'event' },
-    },
-  },
```
juliodialpad, line 126: Why is this being removed here and not in Vue 2 version?

juliodialpad (changes requested):
Looking good so far, just two comments.

components/popover/popover.vue
josedialpad: Good catch, updated removing this and emiting the `keydown` event as it's treated as a custom event by defining in the emits https://github.com/dialpad/dialtone-vue/pull/570/commits/85363931daf5f1042ceaa1f8ca05401efc25d8bf

components/button/button.stories.js
josedialpad: In the Vue 2 version we have some native events defined like this just for documentation. In Vue 3 as it's defined in the emits we need to re-emit them without any additional action (in this case). So for that it's removed just in the Vue 3 implementation.

components/modal/modal.vue
```diff
@@ -288,6 +302,32 @@ export default {
...
+      return {
+        click: event => {
+          (event.target === event.currentTarget) && this.close();
+          this.$emit('click', event);
+        },
+
+        keydown: event => {
+          switch (event.code) {
+            case EVENT_KEYNAMES.esc:
+            case EVENT_KEYNAMES.escape:
```
braddialpad, line 315: Would it ever be "esc" rather than "escape" when using code? https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_code_values

braddialpad: Looks like very old versions of IE used "esc". Guess it doesn't hurt to keep it.

braddialpad (approved):

components/modal/modal.vue
josedialpad: Yeah, I've used the same mapping [Vue is using](https://github.com/vuejs/vue/blob/d6bdff890322bc87792094a1690bcd16373cf82d/src/compiler/codegen/events.ts#L21-L36) for the alias of these events and some of them are for IE11 support.

PR #571, opened 2022-10-28
feat(toggle): add size prop

josedialpad:
# add size prop

## :hammer_and_wrench: Type Of Change

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Add prop for `size` with the values: `md` (default) and `sm`

~⚠️ Depends on the dialtone changes on this branch: [feature/toggle-small-variant](https://github.com/dialpad/dialtone/tree/feature/toggle-small-variant)~
Updated PR with the changes released in [Dialtone 7.4.0](https://github.com/dialpad/dialtone/releases/tag/v7.4.0)

## :bulb: Context

Jira ticket: https://dialpad.atlassian.net/browse/DT-737

Changes included in this PR:
- [x] add prop for size: toggles .d-toggle--small class.
- [ ] add prop to hide icons: suppresses .d-toggle__inner. https://github.com/dialpad/dialtone-vue/pull/573
- [ ] add prop for indeterminate state. https://github.com/dialpad/dialtone-vue/pull/579

## :pencil: Checklist

- [x] I have reviewed my changes
- [x] I have added tests
- [ ] I have added all relevant documentation
- [ ] I have validated components with a screen reader
- [ ] I have validated components keyboard navigation
- [ ] I have considered the performance impact of my change
- [ ] I have checked that my change did not significantly increase bundle size
- [ ] I am exporting any new components or constants in the index.js in the component directory
- [ ] I am exporting any new components or constants in the index.js in the root

PR #573, opened 2022-10-31
feat(toggle): add showIcon prop

josedialpad:
# Add showIcon prop

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Add prop to hide the icon in the toggle

~⚠️ Depends on the dialtone changes on this branch: [feature/toggle-small-variant](https://github.com/dialpad/dialtone/tree/feature/toggle-small-variant)~
Updated PR with the changes released in [Dialtone 7.4.0](https://github.com/dialpad/dialtone/releases/tag/v7.4.0)

## :bulb: Context

Jira ticket: https://dialpad.atlassian.net/browse/DT-737

Changes included in this PR:

- [ ] add prop for size: toggles .d-toggle--small class. (added in https://github.com/dialpad/dialtone-vue/pull/571)
- [x] add prop to hide icons: suppresses .d-toggle__inner.
- [ ] add prop for indeterminate state. https://github.com/dialpad/dialtone-vue/pull/579

## :pencil: Checklist

- [x] I have added tests

### DIFF
--- a/components/toggle/toggle.vue
+++ b/components/toggle/toggle.vue
@@ -22,7 +22,7 @@
       v-on="inputListeners"
     >
       <span
-        v-if="shouldShowIcon"
+        v-if="showIcon"
         class="d-toggle__inner"
       />
     </button>
@@ -149,10 +149,6 @@ export default {
 
         },
       ];
-    },
-
-    shouldShowIcon () {
-      return this.showIcon;
     },
   },
 
--- a/components/toggle/toggle.vue
+++ b/components/toggle/toggle.vue
@@ -21,7 +21,10 @@
       v-bind="$attrs"
       v-on="inputListeners"
     >
-      <span class="d-toggle__inner" />
+      <span
+        v-if="shouldShowIcon"
+        class="d-toggle__inner"
+      />
     </button>
   </div>
 </template>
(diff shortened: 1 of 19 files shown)


components/toggle/toggle.vue
```diff
@@ -138,6 +150,10 @@ export default {
         },
       ];
     },
+
+    shouldShowIcon () {
+      return this.showIcon;
+    },
```
braddialpad: Can't we just use the showIcon prop?

josedialpad: Yeah for this PR should be enough. But the next feature is to add the indeterminate state which shouldn't show the icon and so this computed prop logic will be modified to something like:

```js
  shouldShowIcon () {
    return !this.indeterminate && this.showIcon;
  },
```

josedialpad: update: you're right, no need to use a computed prop since the icons are hidden in the indeterminate state [with the CSS classes](https://github.com/dialpad/dialtone/blob/staging/lib/build/less/components/toggle.less#L132-L134). Removed https://github.com/dialpad/dialtone-vue/pull/573/commits/497e0487456bf7820caebf5f4e6324b361db2990

braddialpad (approved):

PR #574, opened 2022-10-31
feat: icon component

juliodialpad:
# Feat: Icon component

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Added Icon component, tests and documentation

## :bulb: Context

We are adding Icon component to handle new Dialtone 7 icon set.

## :pencil: Checklist

- [x] I have added tests

image

### DIFF
--- a/components/icon/icon.vue
+++ b/components/icon/icon.vue
@@ -37,6 +37,12 @@ export default {
     },
   },
 
+  data () {
+    return {
+      icon: null,
+    };
+  },
+
   computed: {
     iconSize () {
       return ICON_SIZE_MODIFIERS[this.size];
@@ -45,9 +51,22 @@ export default {
     iconName () {
       return kebabCaseToPascalCase(this.name);
     },
+  },
 
-    icon () {
-      return require(`@dialpad/dialtone/lib/dist/vue/v7/${this.iconName}`).default;
+  watch: {
+    $props: {
+      immediate: true,
+      deep: true,
+      handler () {
+        this.getIcon();
+      },
+    },
+  },
+
+  methods: {
+    async getIcon () {
+      const icon = await import(`@dialpad/dialtone/lib/dist/vue/v7/${this.iconName}`);
+      this.icon = icon.default;
     },
   },
 };
--- a/components/icon/icon.vue
+++ b/components/icon/icon.vue
@@ -65,7 +65,7 @@ export default {
 
   methods: {
     async getIcon () {
-      const icon = await import(`@dialpad/dialtone/lib/dist/vue/v7/${this.iconName}.vue`);
+      const icon = await require(`@dialpad/dialtone/lib/dist/vue/v7/${this.iconName}.vue`);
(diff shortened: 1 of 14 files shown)


braddialpad:
Bundle size regression check failed. I know there's a ton of new icons, but they should have caused an increase in dialtone rather than dialtone vue. Do you know what is causing the increase here?

components/icon/icon.vue
```diff
@@ -0,0 +1,54 @@
+<template>
+  <component
+    :is="icon"
+    data-qa="dt-icon"
+    :class="iconSize"
+  />
```
braddialpad: Just realized we should probably have the [title](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/title) as a prop. Since this needs to be the first child within the SVG it makes it a bit awkward to set in this component. 

Probably need to use js to add the title element on mounted and then a watch to update the dom if the prop changes.
Other option is to add a title prop into the actual icon vue components on the dialtone side

components/icon/icon.vue
```diff
@@ -0,0 +1,54 @@
...
+      return ICON_SIZE_MODIFIERS[this.size];
+    },
+
+    iconName () {
+      return kebabCaseToPascalCase(this.name);
+    },
+
+    icon () {
+      return require(`@dialpad/dialtone/lib/dist/vue/v7/${this.iconName}`).default;
+    },
```
juliodialpad: @braddialpad looks like with this I'm importing all the Icons from Dialtone and that increased bundle size.

components/icon/icon.vue
juliodialpad: I'd prefer to add it to dialtone, it's easier to do it that way.

components/icon/icon.vue
juliodialpad: Changed require for imports, that should improve the bundle size

braddialpad: Ugh looks like it might be even worse now for some reason.. 182% increase. I don't really have a good solution for this at the moment we'll have to look again later

juliodialpad: I've been looking at this all day... 

WebPack is trying to bundle the icons coming from `dialtone` when I `require` and chunks every Icon into its single file when `importing`.

Tried a lot of solutions and none of them are working correctly:

- Adding `@dialpad/dialtone/lib/dist/vue/v7` to `config.externals`, it does nothing different.
- Tell webpack to ignore those imported modules with `import(/* webpackIgnore: true */ ...);` I get a "Failed to resolve module specifier" error.

Maybe exporting the icons in a different way from `dialtone` can work?

braddialpad: Yeah I have dealt with these tree shaking issues before and they are never easy to figure out. Unfortunately we can't just release a 182% increase in bundle size.

braddialpad (approved):

braddialpad (approved):
Looks great thanks!

PR #579, opened 2022-11-01
feat(toggle): add indeterminate state

josedialpad:
# add indeterminate state

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Add indeterminate state

## :bulb: Context

Jira ticket: https://dialpad.atlassian.net/browse/DT-737

Changes included in this PR:

- [ ] add prop for size: toggles .d-toggle--small class. (added in https://github.com/dialpad/dialtone-vue/pull/571)
- [ ] add prop to hide icons: suppresses .d-toggle__inner. (added in https://github.com/dialpad/dialtone-vue/pull/573)
- [x] add prop for indeterminate state.

## :pencil: Checklist

- [x] I have added tests

### DIFF
--- a/components/toggle/toggle_constants.js
+++ b/components/toggle/toggle_constants.js
@@ -1,0 +1,8 @@
+export const TOGGLE_SIZE_MODIFIERS = {
+  sm: 'd-toggle--small',
+  md: '',
+};
+
+export default {
+  TOGGLE_SIZE_MODIFIERS,
+};
--- a/components/toggle/toggle_constants.js
+++ b/components/toggle/toggle_constants.js
@@ -3,6 +3,8 @@ export const TOGGLE_SIZE_MODIFIERS = {
   md: '',
 };
 
+export const TOGGLE_CHECKED_VALUES = [false, true, 'mixed'];
+
 export default {
   TOGGLE_SIZE_MODIFIERS,
 };
--- a/.esplint.rec.json
+++ b/.esplint.rec.json
@@ -8,6 +8,10 @@
     "common/mixins/modal.js": {
       "complexity": 1
     },
+    "components/dropdown/dropdown.vue": {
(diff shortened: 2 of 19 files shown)


components/toggle/toggle_constants.js
```diff
@@ -3,6 +3,8 @@ export const TOGGLE_SIZE_MODIFIERS = {
   md: '',
 };
 
+export const TOGGLE_CHECKED_VALUES = [false, true, 'mixed'];
```
braddialpad, line 6: Is it possible to set it to "mixed" in storybook? Do we need to?

josedialpad: Yeah, since the `checked` prop is used to set the initial state of the toggle, it should be added https://github.com/dialpad/dialtone-vue/pull/579/commits/2d4231724d703edaae69d747b8b4532ca3703dd1

josedialpad: I simplified the approach here. Removed the `indeterminate` prop as this state can be managed with the `checked` value (`checked = "mixed"` means indeterminate toggle).

francisrupert: `aria-checked="mixed"`. Love it. I suppose I should have thought of this and might have styled it via `.d-toggle[aria-checked="mixed"] {}` rather than `.d-toggle--indeterminate {} `. But even so, `d-toggle--indeterminate` minimizes CSS specificity.

braddialpad: > I simplified the approach here. Removed the `indeterminate` prop as this state can be managed with the `checked` value (`checked = "mixed"` means indeterminate toggle).

lol was going to ask this exact question, I see you beat me to it.

braddialpad (approved):
Very nice, thanks!

PR #581, opened 2022-11-04
feat: list item group component

braddialpad:
# New Component - List Item Group

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Request for this from @MikaDialpad and @dbecherdialpad for contacts search.

This is very similar to the "Dropdown List" component Jose did for dropdown. Just made it a standalone component and individually documented so it could be used outside of dropdown. Changed the header to use `role="presentation"` rather than the nested `li` method as it is simpler and still accessible.

Already updated the example on dropdown to use this new component. Will remove the old "Dropdown List" component in a future PR, want to make sure it's removal is not going to break anything.

I banged this out pretty quick today so please let me know if you notice any oversights.

## :bulb: Context

To make grouped list items useable anywhere and properly documented.

## :pencil: Checklist

- [x] I have added tests

## :crystal_ball: Next Steps

- Add a link to this vue only component to Dialtone site.
- Publish the similar "List Item Separator" component in the same way.
- Remove the existing "dropdown list" component.

Screenshot 2022-11-04 at 1 47 33 PM

## :link: Sources

Example of the `role=presentation` method for group headers
https://w3c.github.io/aria-practices/examples/listbox/listbox-grouped.html

### DIFF
--- a/components/list_item_group/list_item_group.vue
+++ b/components/list_item_group/list_item_group.vue
@@ -1,0 +1,63 @@
+<template>
+  <ul
+    :id="id"
+    :class="['d-ps-relative', 'd-px0', listClass]"
+    data-qa="dt-dropdown-list-wrapper"
+    :aria-labelledby="`${id}-heading`"
+  >
+    <li
+      v-if="heading"
+      :id="`${id}-heading`"
+      role="presentation"
+      data-qa="dt-dropdown-list-heading"
+      :class="['dt-dropdown-list--header', headingClass]"
+    >
+      {{ heading }}
+    </li>
+    <!-- @slot Slot for the list component -->
+    <slot />
+  </ul>
+</template>
+
+<script>
+import {} from './list_item_group_constants.js';
+import { getUniqueString } from '@/common/utils';
+
+export default {
+  name: 'DtListItemGroup',
+
+  props: {
+    /**
+     * Id of the List Item Group
+     */
+    id: {
+      type: String,
+      default () { return getUniqueString(); },
+    },
+
+    /**
+     * List's heading.
+     */
+    heading: {
+      type: String,
+      default: '',
+    },
+
+    /**
+     * Additional class to style the heading
+     */
+    headingClass: {
+      type: [String, Array, Object],
+      default: '',
+    },
+
+    /**
+     * Additional class for the wrapper list element.
+     */
+    listClass: {
+      type: [String, Array, Object],
+      default: '',
+    },
+  },
+};
+</script>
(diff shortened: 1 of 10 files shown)


components/list_item_group/list_item_group.vue
```diff
@@ -0,0 +1,63 @@
...
+    :aria-labelledby="`${id}-heading`"
+  >
+    <li
+      v-if="heading"
+      :id="`${id}-heading`"
+      role="presentation"
+      data-qa="dt-dropdown-list-heading"
+      :class="['dt-dropdown-list--header', headingClass]"
+    >
+      {{ heading }}
```
MikaDialpad: I'd probably wrap this with a slot also in case we need to pass something more than just a string

components/list_item_group/list_item_group.vue
```diff
@@ -0,0 +1,63 @@
+<template>
+  <ul
+    :id="id"
+    :class="['d-ps-relative', 'd-px0', listClass]"
+    data-qa="dt-dropdown-list-wrapper"
+    :aria-labelledby="`${id}-heading`"
```
MikaDialpad, line 7: This needs a role="group", right?

components/list_item_group/list_item_group.vue
```diff
@@ -0,0 +1,63 @@
...
+  >
+    <li
+      v-if="heading"
+      :id="`${id}-heading`"
+      role="presentation"
+      data-qa="dt-dropdown-list-heading"
+      :class="['dt-dropdown-list--header', headingClass]"
```
MikaDialpad, line 14: Will `dt-dropdown-list--header` provide some default styling? The search [figma for DP 2.0](https://www.figma.com/file/ymr1mdTU5RMU8UTy7IIYKZ/Search-Improvements-7%2F30?node-id=4340%3A229431) has some styling for its heading, but I think it's up to the DT team to decide how the default should look like

components/list_item_group/list_item_group.vue
braddialpad: Yeah good idea.

components/list_item_group/list_item_group.vue
braddialpad: Yeah I think we will just leave it unstyled for now. I'll check with design and see if they have some ideas for this though.

components/list_item_group/list_item_group.vue
braddialpad: Whoops, yep

braddialpad:
Requested fixes made.

MikaDialpad (approved):
LGTM! Thanks Brad for getting to this so quickly!

PR #586, opened 2022-11-08
WIP: DtPresence with DtAvatar 

modialpad:
# Add Presence within DtAvatar

This is a draft PR to discuss the solution approach for integrating `DtPresence` into `DtAvatar`. 

Basically, just `absolute`ly positions `dt-presence` within avatar to be at the bottom right.

After the discussion, I'll close this one and open a new one proper. 

image

### DIFF
--- a/components/avatar/avatar.vue
+++ b/components/avatar/avatar.vue
@@ -7,22 +7,28 @@
       AVATAR_SIZE_MODIFIERS[size],
       AVATAR_COLOR_MODIFIERS[color],
       avatarClass,
+      'd-ps-relative',
     ]"
     data-qa="dt-avatar"
   >
     <!-- @slot Slot for avatar content -->
     <slot />
+    <dt-presence
+      v-if="presence"
+      :presence="presence"
+      class="d-ps-absolute d-b0 d-r0"
+    />
   </div>
 </template>
 
 <script>
-import Vue from 'vue';
-import {
-  AVATAR_SIZE_MODIFIERS,
-  AVATAR_COLOR_MODIFIERS,
-  AVATAR_KIND_MODIFIERS,
-} from './avatar_constants.js';
 import { getUniqueString } from '@/common/utils';
+import Vue from 'vue';
+import { DtPresence } from '../presence';
+import {
+  AVATAR_COLOR_MODIFIERS,
+  AVATAR_KIND_MODIFIERS, AVATAR_SIZE_MODIFIERS,
+} from './avatar_constants.js';
 
 /**
  * An avatar is a visual representation of a user or object.
@@ -30,6 +36,7 @@ import { getUniqueString } from '@/common/utils';
  */
 export default {
   name: 'DtAvatar',
+  components: { DtPresence },
 
   inheritAttrs: false,
 
@@ -72,6 +79,16 @@ export default {
     avatarClass: {
       type: [String, Array, Object],
       default: '',
+    },
+
+    /**
+     * Determines whether to show the presence indicator for
+     * Avatar - accepts PRESENCE_STATES values: 'busy', 'away', 'offline',
+     * or 'active'. By default, it's null and nothing is shown.
+     */
+    presence: {
+      type: String,
+      default: null,
     },
   },
 
--- a/components/avatar/avatar_default.story.vue
+++ b/components/avatar/avatar_default.story.vue
@@ -3,6 +3,7 @@
     :id="id"
     :size="size"
     :color="color"
+    :presence="presence"
   >
     <html-fragment
       :html="defaultSlot"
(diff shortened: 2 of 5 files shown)


components/avatar/avatar.vue
```diff
@@ -7,29 +7,36 @@
       AVATAR_SIZE_MODIFIERS[size],
       AVATAR_COLOR_MODIFIERS[color],
       avatarClass,
+      'd-ps-relative',
```
modialpad, line 10: Had to add `relative` class here because I'm absolutely positioning presence - do we need to update the `dt-avatar` DT class to include the `position: relative` attribute? That would be the way to go

braddialpad: Yep I agree with this.

components/avatar/avatar_presence.story.vue
```diff
@@ -0,0 +1,52 @@
+<template>
```
braddialpad, line 1: Either just show one instance of the avatar here or remove the controls from showing up. Doesn't make much sense to change the properties when there's multiple preset variants on one page.

components/avatar/avatar.vue
```diff
@@ -7,29 +7,36 @@
...
     ]"
     data-qa="dt-avatar"
   >
     <!-- @slot Slot for avatar content -->
     <slot />
+    <dt-presence
+      v-if="presence"
+      :presence="presence"
+      class="d-ps-absolute d-b0 d-r0"
+    />
```
braddialpad, line 20: Do we need different sizes of presence for the different sizes of avatar? It looks a bit off at large and small sizes.

components/avatar/avatar_default.story.vue
```diff
@@ -3,6 +3,7 @@
     :id="id"
     :size="size"
     :color="color"
+    :presence="presence"
```
braddialpad, line 6: need to add this to the other stories as well (icon, initials)

components/avatar/avatar.vue
modialpad: Good point - I haven't considered Avatar when integrating this. 

@francisrupert I guess that's kind of a design question as well - do you think the presence circle should adjust its size depending on the Avatar's size?

Here's where you can see this happening: https://vue.dialpad.design/deploy-previews/pr-586/?path=/docs/components-avatar--presence

components/avatar/avatar_presence.story.vue
modialpad: I think it makes sense to keep the controls for this, so I guess you're saying we should remove this variants file?

braddialpad: Either put each instance in a different story, or have all variants displayed without the controls

components/avatar/avatar.vue
francisrupert: Looks fine to me, though I admit the positioning and scale can suddenly feel off when toggling between the sizes in storybook. It won't come off like that when in context of overall layout. 

I've not seen or heard anything about size variants of presence, and doubt there will be. I'll loop it back over with some key designers, but I don't think any update would occur. Meanwhile, I'm curious how many *large* instances of Avatar w/ presence there is or even ever will be.

image

francisrupert: The _positioning_ of Presence should indeed be very slightly different for each size. Presence size remains the same.

Use the alignment from [this Figma file](https://www.figma.com/file/Vw71A2boup3Gb3PqUQTSiu/Avatar-%2B-Presence?node-id=1%3A2&viewport=1966%2C1826%2C3.85&t=e8mYJk5515vSJQbQ-11). Let me know if you've any questions.

image

modialpad: @francisrupert great thanks! It looks like we're moving the presence slightly toward the center as the Avatar gets larger, right?

@braddialpad Here's what I'm planning to do for this:

- I'll open a PR against `dialtone` to add the `position: relative` css attribute to `d-avatar`.
- I'll close this PR and open a new one updating the stories and using `presence` as suggested by Francis above. From what it looks like, we'll have to update the absolute position values of `top` and `right` depending on the avatar size.

francisrupert: @modialpad,

Righto. Up in and in as avatar size increases. Positioning from `bottom` and `right` will work best; probably best to do so contextually in the CSS rather than use positioning CSS Utilities.

Note the actual position value in CSS will vary from how Figma treats it. That is, Figma doesn't consider the gray outline part of the box model while your current HTML/CSS does. So the fun is to zoom in and pixel-push as appropriate to align to Figma visual.

modialpad:
Closing, will open another one w/ new updates discussed here

PR #594, opened 2022-11-10
WIP: Update DtCollapsible to support showing content on collapsed

modialpad:
Collapsible Variants GIF

### DIFF
--- a/components/collapsible/collapsible.vue
+++ b/components/collapsible/collapsible.vue
@@ -57,7 +57,7 @@
       :aria-hidden="`${!isOpen}`"
       :aria-labelledby="labelledBy"
       :aria-label="ariaLabel"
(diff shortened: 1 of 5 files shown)


components/collapsible/collapsible.vue
```diff
@@ -57,7 +57,7 @@
       :aria-hidden="`${!isOpen}`"
```
braddialpad, line 57: We'll need to add some logic to this, it should only be aria-hidden if contentOnCollapsed is null and isOpen is false

braddialpad (commented):
Overall I think this is working pretty well so far. Continue with the rest of it and request me again when it's ready.

modialpad:
Closing, will open a new PR soon

PR #599, opened 2022-11-15
fix(popover): fix popover on the initial load in vue3 version

josedialpad:
# fix popover on the initial load in vue3 version

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Since we have the [`appear` prop set in the lazy show component](https://github.com/dialpad/dialtone-vue/blob/vue3/components/popover/popover.vue#L54) in the popover's content, the hooks for the transition were being executed in the first load. It's only needed to be set in the combobox with popover to get these transition hooks fired and enable the keyboard navigation.

## :pencil: Checklist

- [ ] I have added tests

### DIFF
--- a/components/popover/popover.vue
+++ b/components/popover/popover.vue
@@ -51,7 +51,7 @@
         :aria-modal="`${!modal}`"
         :transition="transition"
         :show="isOpen"
-        appear
+        :appear="contentAppear"
         :class="['d-popover__dialog', { 'd-popover__dialog--modal': modal }, dialogClass]"
         :style="{
           'max-height': maxHeight,
@@ -265,6 +265,14 @@ export default {
       type: String,
       default: null,
       validator: contentWidth => POPOVER_CONTENT_WIDTHS.includes(contentWidth),
+    },
+
+    /**
+     * Whether to apply transition on initial render in the content lazy show component.
+     */
+    contentAppear: {
+      type: Boolean,
+      default: false,
     },
 
     /**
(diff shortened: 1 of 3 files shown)


braddialpad:
Awesome, is it confirmed this fixes the issue in web-clients search?

josedialpad:
> Awesome, is it confirmed this fixes the issue in web-clients search?

yeah, tested this morning with [this commit](https://github.com/josedialpad/dt-vue-test/commit/4a6b94487c0c91eaaf1b7f1db32c56959483d9f9) released in [@josedialpad/dt-vue-test v3.23.4](https://www.npmjs.com/package/@josedialpad/dt-vue-test/v/3.23.4).

components/popover/popover.vue
```diff
@@ -267,6 +267,14 @@ export default {
       validator: contentWidth => POPOVER_CONTENT_WIDTHS.includes(contentWidth),
     },
 
+    /**
+     * Whether to apply transition on initial render in the content lazy show component.
+     */
+    contentAppear: {
+      type: Boolean,
+      default: false,
+    },
```
braddialpad, line 276: hmm wondering if we can know this automatically if the component has the open prop set to true when created

braddialpad (approved):

components/popover/popover.vue
josedialpad: tried that but it doesn't execute the transition hooks anyways

PR #601, opened 2022-11-16
feat(combobox): add prop to pass classes to the empty list item

josedialpad:
# add prop to pass classes to the empty list item

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Add a prop to apply classes in the empty list element. In the product [they are using the `dt-default-list-item` class](https://github.com/dialpad/firespotter/pull/24236#discussion_r1024216301) to apply certain styles, with this PR they can use utility classes in the empty list element.

## :pencil: Checklist

- [x] I have added tests


components/combobox/combobox_empty-list.vue
```diff
@@ -4,9 +4,11 @@
...
     <dt-list-item
-      class="d-d-flex d-jc-center"
+      class="dt-empty-list-item d-fs-200 d-lh4 d-py4 d-px12 d-d-flex d-jc-center d-ai-center"
       role="option"
       navigation-type="none"
+      type="custom"
+      :class="itemClass"
     >
       <span>{{ message }}</span>
     </dt-list-item>
```
braddialpad: Would it be better to just wrap this list item in a slot an allow them to override it however they wish?

components/combobox/combobox_empty-list.vue
```diff
@@ -4,9 +4,11 @@
     data-qa="dt-combobox-empty-list"
   >
     <dt-list-item
-      class="d-d-flex d-jc-center"
+      class="dt-empty-list-item d-fs-200 d-lh4 d-py4 d-px12 d-d-flex d-jc-center d-ai-center"
```
braddialpad: with so many default utility classes in here i'd worry that some of them would not be overridden by `itemclass` due to `!important`

josedialpad: yeah I mean if you use other utility classes will override due to the order, but using custom classes you will need to use `!important` in the already defined default classes. So moved the default styles to a class. https://github.com/dialpad/dialtone-vue/pull/601/commits/e65f25683e0256e9fa1a7ad52e723e0042312057

components/combobox/combobox_empty-list.vue
josedialpad: Yeah to be consistent with other components too, so added a slot to use here https://github.com/dialpad/dialtone-vue/pull/601/commits/ee7edea7b224a308e56286b0dd0b1cc3c0b5677b

braddialpad (approved):
Looks good!

PR #603, opened 2022-11-17
feat: integrate `DtPresence` within `DtAvatar`

modialpad:
# Enable showing presence in `DtAvatar`

Allow a consumer to set a `presence` prop to show a presence indicator for a particular avatar. 

The position of the presence indicator needs to correspond with the size of the Avatar as outlined here: https://www.figma.com/file/Vw71A2boup3Gb3PqUQTSiu/Avatar-%2B-Presence?node-id=1%3A2&viewport=1966%2C1826%2C3.85&t=v4vpriXLRXUR2F4u-0

- [x] Feature

## :book: Description

Following the work for adding Presence, the Avatar component will need to be updated to render this new `<Presence/>` component internally whenever a consumer passes a new presence prop.

## :pencil: Checklist

- [x] I have added tests

image

### DIFF
--- a/components/avatar/avatar.vue
+++ b/components/avatar/avatar.vue
@@ -17,6 +17,7 @@
       :presence="presence"
       :style="presenceStyles"
       class="d-ps-absolute"
+      v-bind="presenceProps"
       data-qa="dt-presence"
     />
   </div>
@@ -90,6 +91,14 @@ export default {
     presence: {
       type: String,
       default: null,
+    },
+
+    /**
+     * A set of props to be passed into the presence component.
+     */
+    presenceProps: {
+      type: Object,
+      default: () => ({}),
     },
   },
 
--- a/components/avatar/avatar.vue
+++ b/components/avatar/avatar.vue
@@ -7,22 +7,28 @@
       AVATAR_SIZE_MODIFIERS[size],
       AVATAR_COLOR_MODIFIERS[color],
       avatarClass,
+      'd-ps-relative',
     ]"
     data-qa="dt-avatar"
   >
     <!-- @slot Slot for avatar content -->
     <slot />
+    <dt-presence
+      v-if="presence"
+      :presence="presence"
+      class="d-ps-absolute d-b0 d-r0"
+    />
   </div>
 </template>
 
 <script>
-import Vue from 'vue';
-import {
-  AVATAR_SIZE_MODIFIERS,
-  AVATAR_COLOR_MODIFIERS,
-  AVATAR_KIND_MODIFIERS,
-} from './avatar_constants.js';
 import { getUniqueString } from '@/common/utils';
+import Vue from 'vue';
+import { DtPresence } from '../presence';
+import {
+  AVATAR_COLOR_MODIFIERS,
+  AVATAR_KIND_MODIFIERS, AVATAR_SIZE_MODIFIERS,
+} from './avatar_constants.js';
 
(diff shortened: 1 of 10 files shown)


package-lock.json
```diff
@@ -22276,415 +22570,443 @@
...
+      "funding": {
+        "type": "opencollective",
+        "url": "https://opencollective.com/postcss/"
       }
     },
-    "node_modules/postcss-normalize-unicode": {
-      "version": "5.1.0",
-      "resolved": "https://registry.npmjs.org/postcss-normalize-unicode/-/postcss-normalize-unicode-5.1.0.tgz",
-      "integrity": "sha512-J6M3MizAAZ2dOdSjy2caayJLQT8E8K9XjLce8AUQMwOrCvjCHv24aLC/Lps1R1ylOfol5VIDMaM/Lo9NGlk1SQ==",
+    "node_modules/postcss-color-hex-alpha/node_modules/source-map": {
```
modialpad, line 22900: Not sure if all these updates are normal, I bumped DT's version and ran `npm install`. Does npm still upgrade minor versions automatically?

braddialpad: 🤷 npm has a mind of it's own.

components/avatar/avatar.vue
```diff
@@ -85,6 +112,35 @@ export default {
...
+            bottom: '-2px',
+            right: '-2px',
+          };
+        default:
+          return {
+            bottom: '0',
+            right: '0',
+          };
+      }
+    },
```
braddialpad: What we probably should have done here is include these styles as classes within avatar in the dialtone repo, and then apply these classes based on the avatar size. This would align with our practices for most of our other components, and the classes would be documented in dialtone instead of hidden in the vue component

I'm okay with merging this as is though and then further iterating to make this change.

braddialpad (approved):
Looks good, just one future change.

components/avatar/avatar.vue
modialpad: Np, I think I'm fine with holding this PR until making that change in DT. I've seen that pattern before so I know what you mean - I created a draft [here](https://github.com/dialpad/dialtone/pull/739) based on that. Could you take a look and let me know? I'm not sure if we have a naming standard for targeting DT components within other DT components.

https://github.com/dialpad/dialtone/pull/739

When that's merged we'll just use a constant to map to these classes for presence here.

francisrupert: I'll do a quick PR in `dialtone` to add it to https://dialpad.design/components/avatar.html

francisrupert: @modialpad Sorry I got ahead of it and didn't notice your draft PR. What I did at https://github.com/dialpad/dialtone/pull/740 should do it.

francisrupert: https://github.com/dialpad/dialtone/pull/740 has been merged to staging. I'll defer to Brad on timing of publishing so you can update this PR to use it.

modialpad: Awesome thanks!

modialpad: @braddialpad I updated this w/ @francisrupert's latest changes - let me know if it LG and I'll just merge.

One thing I noticed is that for the presence class modifiers we don't have an explicit modifier for `sm` case. It's actually the default, so it still works as expected, but in the tests I just had to assume that the styles are correct for the default case.

PR #609, opened 2022-11-18
fix(collapsible): fix keyboard events

iropolo:
# Fix collapsible keyaboard events

- [x] Fix

## :book: Description

The purpose of the change is fix a bug explained in [DT-763](https://dialpad.atlassian.net/browse/DT-763)

## :bulb: Context

We moved directly the `@click` event into the button and remove parents events.
With this we avoid side effects between these.

We also update the test file reflecting this change.

## :pencil: Checklist

- [x] I have added tests

image

## :link: Sources

[DT-763](https://dialpad.atlassian.net/browse/DT-763)

### DIFF
--- a/components/collapsible/collapsible.test.js
+++ b/components/collapsible/collapsible.test.js
@@ -5,8 +5,10 @@ import sinon from 'sinon';
 import axe from 'axe-core';
 import configA11y from '../../storybook/scripts/storybook-a11y-test.config';
 
-const anchorText = 'anchor text';
 const content = '<div data-qa="content-element"> Test Text </div>';
+const baseProps = {
+  anchorText: 'anchor text',
+};
 
 describe('Dialtone vue Collapsible Component Tests', function () {
   // Wrappers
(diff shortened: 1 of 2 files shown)


components/collapsible/collapsible.test.js
```diff
@@ -1,40 +1,40 @@
 import { assert } from 'chai';
 import { config, mount } from '@vue/test-utils';
-import DtCollpasible from './collapsible.vue';
```
braddialpad, line 3: lol

components/collapsible/collapsible.vue
```diff
@@ -105,7 +103,7 @@ export default {
      */
     anchorText: {
       type: String,
-      default: null,
+      default: 'Button',
```
iropolo: This was required for `axe-core accessibility`

_Personal Opinion_ I understand this as mandatory, we should have anchor with text as a default value.

Before it was added by the slot, but we removed the slot override to show the default button instead.

Here we have two options, give the text value by props, or just adding it as a default value.

braddialpad: I think the correct way to fix this is with a custom validator, that validates for this:

> anchorText and anchor slot content cannot both be falsy

You can see an example in combobox.vue, `validateEmptyListProps()`

braddialpad (commented):
Congrats on your first PR!

components/collapsible/collapsible.vue
iropolo: Completely agree :)

braddialpad (approved):
Thanks!

PR #613, opened 2022-11-22
build: fix visual tests

juliodialpad:
# Fix visual tests

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Removed `ref` from the checkout action, it was failing on pushing to staging.

## :bulb: Context

Visual tests failed when merged #608 due to `ref` variable pointing to `refs/pull/` and that is only correct on pull_request actions not on push.

## :pencil: Checklist

- [ ] I have added tests

### DIFF
--- a/.github/workflows/visual_tests.yml
+++ b/.github/workflows/visual_tests.yml
@@ -25,8 +25,6 @@ jobs:
     steps:
       - name: Check out branch
         uses: actions/checkout@v3
-        with:
-          ref: "refs/pull/${{ github.event.pull_request.number }}/merge"
 
       - name: Use Node.js
         uses: actions/setup-node@v1


.github/workflows/visual_tests.yml
```diff
@@ -25,8 +25,6 @@ jobs:
     steps:
       - name: Check out branch
         uses: actions/checkout@v3
-        with:
-          ref: "refs/pull/${{ github.event.pull_request.number }}/merge"
```
braddialpad, line 29: don't we need to use this when it's a PR rather than directly on staging though?

juliodialpad: According to https://github.com/actions/checkout docs:

> The branch, tag or SHA to checkout. When checking out the repository that
> triggered a workflow, this defaults to the reference or SHA for that event.
> Otherwise, uses the default branch.
> ref: ''

I understand that by default it checkouts to the SHA from the pull request, right?

braddialpad (approved):

PR #615, opened 2022-11-22
fix(collapsible): fix keyboard events - Vue3

iropolo:
Vue2 PR: https://github.com/dialpad/dialtone-vue/pull/609

### DIFF
--- a/components/collapsible/collapsible.test.js
+++ b/components/collapsible/collapsible.test.js
@@ -14,11 +14,13 @@ describe('Dialtone vue Collapsible Component Tests', function () {
   let contentElement;
   let contentWrapperElement;
   let anchorElement;
-  let scopedSlots = {};
+  let slots = { content };
 
   // Environment
-  const slots = { content };
-  const propsData = {
+  const attrs = {
+    css: false, // Important attr to let test-utils fire the (after-enter and after-leave) events correctly
+  };
+  const props = {
     anchorText,
   };
 
@@ -26,20 +28,25 @@ describe('Dialtone vue Collapsible Component Tests', function () {
     contentElement = undefined;
     contentWrapperElement = undefined;
     anchorElement = undefined;
-    scopedSlots = {};
+    slots = { content };
   };
 
   const _setChildWrappers = () => {
     anchorElement = wrapper.find('[data-qa="dt-button"]');
     contentElement = wrapper.find('[data-qa="content-element"]');
-    contentWrapperElement = wrapper.findComponent({ ref: 'contentWrapper' });
+    contentWrapperElement = wrapper.getComponent('.d-dt-collapsible__content');
   };
 
   const _mountWrapper = () => {
     wrapper = mount(DtCollapsible, {
-      propsData,
-      scopedSlots,
+      props,
       slots,
+      attrs,
+      global: {
+        stubs: {
+          transition: false,
+        },
+      },
       attachTo: document.body,
     });
     _setChildWrappers();
@@ -82,8 +89,8 @@ describe('Dialtone vue Collapsible Component Tests', function () {
 
   describe('When scoped slot is provided', function () {
     it('should render the scoped slot', function () {
-      const anchor = '<button data-qa="anchor-element" v-bind="props.attrs">click me</button>';
-      scopedSlots = { anchor };
+      const anchor = '<button data-qa="anchor-element">click me</button>';
+      slots = { anchor };
       _mountWrapper();
       const anchorSlotContainer = wrapper.find('[data-qa="anchor-element"]');
       assert.exists(anchorSlotContainer, 'anchor slot exists');
@@ -149,10 +156,10 @@ describe('Dialtone vue Collapsible Component Tests', function () {
   describe('If anchor text and anchor slot content are falsy', function () {
     it('should output error message', async function () {
       const consoleErrorSpy = sinon.spy(console, 'error');
-      propsData.anchorText = undefined;
+      props.anchorText = undefined;
       _mountWrapper();
       assert.isTrue(consoleErrorSpy.calledWith('anchor text and anchor slot content cannot both be falsy'));
-      propsData.anchorText = anchorText;
+      props.anchorText = anchorText;
     });
   });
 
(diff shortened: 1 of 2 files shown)


components/collapsible/collapsible.test.js
```diff
@@ -1,52 +1,45 @@
...
     contentElement = wrapper.find('[data-qa="content-element"]');
-    contentWrapperElement = wrapper.getComponent('.d-dt-collapsible__content');
+    contentWrapperElement = wrapper.findComponent({ ref: 'contentWrapper' });
   };
 
   const _mountWrapper = () => {
-    wrapper = mount(DtCollpasible, {
-      props,
+    wrapper = mount(DtCollapsible, {
+      propsData,
```
braddialpad: propsData is just called "props" in vue 3

components/collapsible/collapsible.test.js
```diff
@@ -1,52 +1,45 @@
...
+  let scopedSlots = {};
 
   // Environment
-  const props = {};
-  const attrs = {
-    css: false, // Important attr to let test-utils fire the (after-enter and after-leave) events correctly
+  const slots = { content };
+  const propsData = {
+    anchorText,
   };
```
braddialpad, line 25: Is this not needed anymore?

components/collapsible/collapsible.test.js
```diff
@@ -1,52 +1,45 @@
...
-    contentWrapperElement = wrapper.getComponent('.d-dt-collapsible__content');
+    contentWrapperElement = wrapper.findComponent({ ref: 'contentWrapper' });
   };
 
   const _mountWrapper = () => {
-    wrapper = mount(DtCollpasible, {
-      props,
+    wrapper = mount(DtCollapsible, {
+      propsData,
+      scopedSlots,
```
braddialpad: scoped slots no longer exists in vue 3, you can just use "slots"

braddialpad (changes requested):
Some differences from vue 2 / vue 3 that you probably had no idea about. 

These guides are good to check out if weird stuff is happening when you transfer from vue 2 to 3. Tests are often the hardest.

Vue 2 to 3 breaking changes: https://v3-migration.vuejs.org/breaking-changes/
Vue test utils 1 to 2 breaking changes: https://test-utils.vuejs.org/migration/#propsdata-is-now-props

braddialpad:
When we do vue 3 PR's we usually don't fill out any PR description, we just add a link to the vue2 PR.

components/collapsible/collapsible.vue
```diff
@@ -11,9 +11,6 @@
         'd-dt-collapsibe__anchor',
```
iropolo, line 11: This should be d-dt-collapsible__anchor ? 🤔

components/collapsible/collapsible.vue
```diff
@@ -229,14 +227,7 @@ export default {
     labelledBy () {
       // aria-labelledby should be set only if aria-labelledby is passed as a prop, or if
       // there is no aria-label and the labelledby should point to the anchor
-      return this.ariaLabelledby || (!this.ariaLabel && getUniqueString('DtCollapsible__anchor'));
-    },
-
-    contentStyle () {
```
iropolo, line 235: I deleted `contentStyle ()` because we are not using it.
Also we are not using it in the current staging-vue3 branch [check](https://github.com/dialpad/dialtone-vue/blob/staging-vue3/components/collapsible/collapsible.vue)

It was there for some reason?

components/collapsible/collapsible.vue
braddialpad: yep that is definitely a typo

components/collapsible/collapsible.vue
braddialpad: Doesn't seem needed.. you can remove

components/collapsible/collapsible.test.js
```diff
@@ -144,6 +153,16 @@ describe('Dialtone vue Collapsible Component Tests', function () {
     });
   });
 
+  describe('If anchor text and anchor slot content are falsy', function () {
+    it('should output error message', async function () {
+      const consoleErrorSpy = sinon.spy(console, 'error');
```
braddialpad: You're setting a spy on the global `console.error` and not restoring it after. This is what is making other tests fail.

components/collapsible/collapsible.test.js
```diff
@@ -144,6 +153,16 @@ describe('Dialtone vue Collapsible Component Tests', function () {
...
+      props.anchorText = undefined;
+      _mountWrapper();
+      assert.isTrue(consoleErrorSpy.calledWith('anchor text and anchor slot content cannot both be falsy'));
+      props.anchorText = anchorText;
```
braddialpad: not necessary, or rather this should be set in the mountWrapper

components/collapsible/collapsible.test.js
```diff
@@ -144,6 +153,16 @@ describe('Dialtone vue Collapsible Component Tests', function () {
     });
   });
 
+  describe('If anchor text and anchor slot content are falsy', function () {
+    it('should output error message', async function () {
+      const consoleErrorSpy = sinon.spy(console, 'error');
+      props.anchorText = undefined;
+      _mountWrapper();
```
braddialpad: Should be doing setup stuff such as this within a beforeEach in the describe. Ideally the `it()` should just be an assertion and that's it.

components/collapsible/collapsible.test.js
```diff
@@ -88,6 +87,16 @@ describe('Dialtone vue Collapsible Component Tests', function () {
     });
   });
 
+  describe('When scoped slot is provided', function () {
+    it('should render the scoped slot', function () {
+      const anchor = '<button data-qa="anchor-element">click me</button>';
+      slots = { anchor };
+      _mountWrapper();
+      const anchorSlotContainer = wrapper.find('[data-qa="anchor-element"]');
```
braddialpad: We put all of our `.find` in `_setChildWrappers`. You can just recall it if needed.

braddialpad: set state in beforeEach as mentioned earlier

braddialpad (changes requested):

braddialpad (approved):
Looks good thanks!

PR #619, opened 2022-11-24
fix(root-layout): slot issues

juliodialpad:
# Fix Root Layout slot issues

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Added a v-if condition to force the removal of the slot if the slot has no content.

## :bulb: Context

Found root layout issues on Vue3, the slots where not being removed if the slot content was empty.

## :pencil: Checklist

- [ ] I have added tests

Uploading Screen Recording 2022-11-24 at 12.58.02 p.m..mov…

### DIFF
--- a/components/root_layout/root_layout.vue
+++ b/components/root_layout/root_layout.vue
@@ -17,11 +17,17 @@
       :sidebar-width="sidebarWidth"
       :sidebar-position="sidebarPosition"
     >
-      <template #sidebar>
+      <template
+        v-if="$slots.sidebar"
+        #sidebar
+      >
         <!-- @slot Slot for the sidebar -->
         <slot name="sidebar" />
       </template>
-      <template #content>
+      <template
+        v-if="$slots.default"
+        #content
+      >
         <!-- @slot Slot for main content -->
         <slot name="default" />
       </template>


components/root_layout/root_layout.vue
```diff
@@ -17,11 +17,17 @@
       :sidebar-width="sidebarWidth"
       :sidebar-position="sidebarPosition"
     >
-      <template #sidebar>
+      <template
+        v-if="$slots.sidebar"
```
josedialpad, line 21: This doesn't seem to be an issue in the Vue2 implementation and works as expected. Since there's an internal change in how the slots are passed between Vue2 and Vue3, I think makes sense to leave the fix just for the vue3 implementation. What do you think?

juliodialpad: I think its better to keep both implementations as close as possible, I'd prefer to maintain this implementation here too.

josedialpad: > I think its better to keep both implementations as close as possible

Yeah I totally agree with that but why add a fix for an issue we don't have in vue2?

josedialpad: I mean, not a big deal adding this specific fix since there won't be any problem but wondering on applying the same criteria in more largar fixes.

josedialpad (approved):

components/root_layout/root_layout.vue
juliodialpad: Yeah, you're right, we need to consider that too. I'll merge this but lets talk about this on our next grooming.

PR #620, opened 2022-11-24
fix(root-layout): slot issues - vue 3

juliodialpad:
# Vue 3 version of #619

### DIFF
--- a/components/root_layout/root_layout.vue
+++ b/components/root_layout/root_layout.vue
@@ -17,11 +17,17 @@
       :sidebar-width="sidebarWidth"
       :sidebar-position="sidebarPosition"
     >
-      <template #sidebar>
+      <template
+        v-if="$slots.sidebar"
+        #sidebar
+      >
         <!-- @slot Slot for the sidebar -->
         <slot name="sidebar" />
       </template>
-      <template #content>
+      <template
+        v-if="$slots.default"
+        #content
+      >
         <!-- @slot Slot for main content -->
         <slot name="default" />
       </template>
--- a/components/root_layout/root_layout_default.story.vue
+++ b/components/root_layout/root_layout_default.story.vue
@@ -1,6 +1,6 @@
 <template>
   <dt-root-layout
-    :body-class="bodyClass"
+    :body-class="$attrs.bodyClass"
     :header-class="$attrs.headerClass"
     :header-sticky="$attrs.headerSticky"
     :header-height="$attrs.headerHeight"


components/root_layout/root_layout.vue
```diff
@@ -17,11 +17,17 @@
       :sidebar-width="sidebarWidth"
       :sidebar-position="sidebarPosition"
     >
-      <template #sidebar>
+      <template
+        v-if="$slots.sidebar"
```
josedialpad, line 21: Wondering why it works in the parent component and not in the `root_layout_body` child component that already has the validation for rendering when the `sidebar` slot is truthy https://github.com/dialpad/dialtone-vue/blob/vue3/components/root_layout/root_layout_body.vue#L7

juliodialpad: Yeah, that's the weird behavior, something in vue3 changed and the slot is passed even if it is empty

josedialpad: There were some breaking changes regarding slots, now they're functions https://v3-migration.vuejs.org/breaking-changes/slots-unification.html

2022-11-24 at 16 58 30@2x

`this.$slots.sidebar()` is truthy since the [html comment](https://github.com/dialpad/dialtone-vue/blob/vue3/components/root_layout/root_layout.vue#L21) is passed as a node. So I guess we need to figure out a better approach to validate here, since we have other components that can have the same issue (for example: the [notice content](https://github.com/dialpad/dialtone-vue/blob/vue3/components/notice/notice_content.vue#L6-L16) will render the title node even when no `title` prop nor `titleOverride` is provided)

juliodialpad: Why is `v-if` not a good approach? we can do the same in all components

josedialpad: It works but I think this doesn't fix the root of the problem. Adding an html comment inside the template of the slot, will cause the issue to appear again. Example:

```
<template>
  <dt-root-layout>
    <template #sidebar>
      <!-- @slot Slot for the sidebar -->
      <div v-if="shouldRenderSidebar">...</div>
  </dt-root-layout>
  <other-stuff>
  ...
</template>
```

josedialpad: Alright, I've just created [a ticket](https://dialpad.atlassian.net/browse/DT-836) to figure out the best approach to migrate in these cases. I think we can move on with this fix for now.

josedialpad (approved):

components/root_layout/root_layout.vue
juliodialpad: > It works but I think this doesn't fix the root of the problem. Adding an html comment inside the template of the slot, will cause the issue to appear again. Example:
> 
> ```
> <template>
>   <dt-root-layout>
>     <template #sidebar>
>       
>       <div v-if="shouldRenderSidebar">...</div>
>   </dt-root-layout>
>   <other-stuff>
>   ...
> </template>
> ```

Oh now I see, sounds good!  Thank you.

PR #621, opened 2022-11-24
feat(combobox): add clickOnSelect option

MikaDialpad:
# feat(combobox): add clickOnSelect option

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

This is my suggestion to solve [this issue](https://github.com/dialpad/web-clients/pull/224#discussion_r1015998282) that came up in a PR that has a more complex use case for the combobox.

## :bulb: Context

The issue I aim to solve is when the combobox contains list items from multiple sources and getting just the index of the item on keyboard selection makes it complicated to match with the actual item.

```
<DtCombobox @select="selectRow">
  <template #input="{ inputProps }">
    ...
  </template>
  <template #list="{ listProps }">
    <DtListItemGroup :heading="First group of items">
      <DtListItem
        v-for="item in items"
        @click="selectRow(item)"
      />
    </DtListItemGroup>
    <DtListItemGroup :heading="Second group of items">
      <DtListItem
        v-for="item in items"
        @click="selectRow(item)"
      />
    </DtListItemGroup>
  </template>
</DtCombobox>
```
In this case `selectRow` has to handle both indexes and item data objects being passed to it and it becomes pretty awkward. We can't just pass indexes anymore in the list item's click handler because we're looping through two lists and they have their own indexes. I think the simplest solution here is to handle the selection in the click handlers of the list items and have the combobox programmatically click on the item when a selection comes from the keyboard navigation.

## :pencil: Checklist

### DIFF
--- a/components/combobox/combobox.vue
+++ b/components/combobox/combobox.vue
@@ -164,6 +164,16 @@ export default {
       type: [String, Object, Array],
       default: '',
     },
+
+    /**
+     * Programmatically click on the active list item element when a selection
+     * comes from keyboard navigation, i.e. pressing the "Enter" key.
+     * @values true, false
+     */
+    clickOnSelect: {
+      type: Boolean,
+      default: false,
+    },
   },
 
   emits: [
@@ -317,6 +327,10 @@ export default {
 
       if (this.highlightIndex >= 0) {
         this.$emit('select', this.highlightIndex);
+
+        if (this.clickOnSelect) {
+          this.activeItemEl?.click();
+        }
       }
     },
 
--- a/components/combobox/combobox.stories.js
+++ b/components/combobox/combobox.stories.js
@@ -11,6 +11,7 @@ export const argsData = {
   onHighlight: action('highlight'),
   onSelect: action('select'),
   onOpened: action('opened'),
+  onItemClick: action('click'),
 };
 
 export const argTypesData = {
@@ -117,6 +118,11 @@ export const argTypesData = {
     },
   },
   onOpened: {
+    table: {
+      disable: true,
+    },
+  },
+  onItemClick: {
     table: {
       disable: true,
     },
--- a/components/combobox/combobox.stories.js
+++ b/components/combobox/combobox.stories.js
@@ -96,6 +96,9 @@ export const argTypesData = {
   loading: {
     defaultValue: false,
   },
+  clickOnSelect: {
+    defaultValue: false,
+  },
 
   // Action Event Handlers
   onEscape: {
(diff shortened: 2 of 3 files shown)


components/combobox/combobox.vue
```diff
@@ -317,6 +327,10 @@ export default {
 
       if (this.highlightIndex >= 0) {
         this.$emit('select', this.highlightIndex);
+
+        if (this.clickOnSelect) {
+          this.activeItemEl?.click();
```
josedialpad, line 332: will this emit the `select` event two times? the second will be for this click

MikaDialpad: My intention was that when using this option you wouldn't handle the selection from the combobox's `@select` event anymore, but just by attaching a `@click` handler on the list item. It still emits the event in case it's needed for some reason, but it's not needed anymore as long as the click handlers are there.

MikaDialpad: If you think this is not clear enough and people might accidentally start using this option in conjunction with the `@select` event, which would admittedly result in double selections, we could consider changing this so that instead of a "select" event it emits a "clicked" or "item-clicked" event instead when this option is being used.

I think it's still important to emit something here even when using this option so that we can distinguish between keyboard and mouse selections in the parent component for amplitude tracking reasons for example.

josedialpad: Ok, yeah I can't think in which case you'd use both, the `select` and `click` events.
I was confused at first and I thought it would emit the select twice but it's because we're using the same event handler for the click in the [SB's default story](https://github.com/dialpad/dialtone-vue/blob/staging/components/combobox/combobox_default.story.vue#L37). So updating this handler to log the `click` event in SB I think will remove the confusion.
Example:

```js
// combobox_default.story.vue
        <dt-list-item
          v-for="(item, i) in displayItems"
          :key="item.id"
          role="option"
          navigation-type="arrow-keys"
          @click="onItemClick(i)"
        >
```

```js
// combobox.stories.js
export const argsData = {
  ...
  onItemClick: action('click'),
};

export const argTypesData = {
  ...
  onItemClick: {
    table: {
      disable: true,
    },
  },

```

josedialpad (approved):
w/ a comment to update the `click` event handler in the `list-item` in the default story to avoid confusion.

components/combobox/combobox.vue
MikaDialpad: Ah gotcha, yeah makes total sense. I'll add a different action name for the click like you suggested.

PR #623, opened 2022-11-25
feat(avatar): add dialtone 7 visual refresh

josedialpad:
# Add dialtone 7 visual refresh

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

- Added `xs` and `xl` sizes

## :pencil: Checklist

- [ ] I have added tests


components/avatar/avatar.vue
```diff
@@ -150,6 +177,30 @@ export default {
...
+      // get 3 unique colors, 2 from colorsWith100 and one from colorsWith200
+      while (colors.size < 3) {
+        if (colors.size === 2) {
+          colors.add(getRandomElement(colorsWith200));
+        } else {
+          colors.add(getRandomElement(colorsWith100));
+        }
+      }
+
+      const shuffledColors = Array.from(colors).sort(() => 0.5 - Math.random());
```
josedialpad, line 224: @francisrupert can you validate this logic is okay? We obtain 2 color stops, two `100` and one `200` in no particular order, which means 100 and 200 color stops can be assigned to any `--avatar-gradient-stop-[1..3]` CSS custom property. That certainly will give us a broader spectrum of variants, right?

francisrupert: Yesh, just perfect. Sweet shuffles. At first I thought it might be a challenge to prevent reusing a color (e.g. don't allow `purple-100` to be used twice), but it appears its simplicity inhibits it: "randomly choose 2 from colorsWith100".

I obviously defer to your judgement on poking holes in it, and of course @braddialpad.

I'm actually curious on the math of this, I'm probably doing this wrong... `(4^2)*((4^2)/2)=128`... doesn't feel right, there's gotta be more than 128 combinations.

components/avatar/avatar.vue
```diff
@@ -150,6 +177,30 @@ export default {
       return element?.tagName?.toUpperCase() === 'IMG';
     },
 
+    randomizeGradientAngle () {
+      const angles = [45, -45, 90, -90, 135, -135, 180, -180];
```
braddialpad: Store in consts file?

components/avatar/avatar.vue
braddialpad: Nice use of `Set()` here to prevent duplicates. I think the logic here should work totally fine. This could possibly be made more dynamic but not sure it's necessary unless we are planning on changing number of colors and number of stops all the time

components/avatar/avatar.vue
```diff
@@ -150,6 +177,30 @@ export default {
...
+    },
+
+    randomizeGradientColorStops () {
+      const colors = new Set();
+      const colorsWith100 = ['purple-100', 'magenta-100', 'gold-100', 'blue-100'];
+      const colorsWith200 = ['purple-200', 'magenta-200', 'gold-200', 'blue-200'];
+
+      // get 3 unique colors, 2 from colorsWith100 and one from colorsWith200
+      while (colors.size < 3) {
+        if (colors.size === 2) {
```
braddialpad: Maybe get rid of these magic numbers, store them in descriptive constants, somethin like MAX_GRADIENT_COLORS, MAX_GRADIENT_COLORS_100

components/avatar/avatar.vue
```diff
@@ -125,12 +144,20 @@ export default {
...
+        const childEl = this.$el;
+
         if (this.kind === 'image') {
-          firstChild.classList.add('d-avatar__image');
+          childEl.firstChild.classList.add('d-avatar__image');
+        }
+        if (this.kind === 'initials') {
+          if (this.size === 'sm') {
+            childEl.innerText = childEl.innerText[0];
+          }
```
braddialpad: when size is 'xs' no text should be shown

components/avatar/avatar.vue
braddialpad: I tried refactoring this to get rid of the sort and it just ended up being uglier than what is currently here.

components/avatar/avatar.vue
josedialpad: Updated

components/avatar/avatar.vue
josedialpad: Yeah good call, updated

components/avatar/avatar.vue
josedialpad: Updated

components/avatar/avatar.vue
```diff
@@ -125,33 +147,94 @@ export default {
...
+        const childEl = this.$el;
+
         if (this.kind === 'image') {
-          firstChild.classList.add('d-avatar__image');
+          childEl.firstChild.classList.add('d-avatar__image');
+          // TO DO: remove attached event listener
+          childEl.firstChild.addEventListener('error', () => {
+            this.kind = 'initials';
+            this.imageLoadedSuccessfully = false;
+            childEl.firstChild.classList.remove('d-avatar__image');
```
josedialpad: @braddialpad it's pending to fix the removal of the class when it falls back to the initials when the image couldn't be loaded. It's removing the first time with the listener but when the change is applied in the DOM, the `init` method is executed again due to the call in the `updated` hook so the class is applied again. If we add the class just when the image is successfully loaded, that implies that the size will be incorrect until it's fully loaded.

components/avatar/avatar.vue
```diff
@@ -125,33 +147,98 @@ export default {
...
+          childEl.firstChild.addEventListener('error', () => {
+            this.kind = 'initials';
+            this.imageLoadedSuccessfully = false;
+            childEl.firstChild?.classList?.remove('d-avatar__image');
+          });
+
+          childEl.firstChild.addEventListener('load', () => {
+            this.imageLoadedSuccessfully = true;
+            childEl.classList.add('d-avatar__image--loaded');
+          });
```
francisrupert: I initially was thinking you would change the `--avatar-color-background` and the `--avatar-gradient-stop-N` to `transparent`, but I like this better.

Nitpick: BEM structure is a bit off for `d-avatar__image--loaded`, since `--loaded` is not directly modifying the `img.d-avatar__image`. 

I'd go with `d-avatar--image-loaded`.

francisrupert:
Still see [Color prop in doc view](https://vue.dialpad.design/deploy-previews/pr-623/iframe.html?viewMode=docs&id=components-avatar--default&args=#with-color).

image

josedialpad:
> Still see [Color prop in doc view](https://vue.dialpad.design/deploy-previews/pr-623/iframe.html?viewMode=docs&id=components-avatar--default&args=#with-color).
> 
> image

thanks, removed

components/avatar/avatar.vue
josedialpad: Yes it's better with `d-avatar--image-loaded`, updated

francisrupert (commented):
Looking good from my standpoint. Really delivers. I defer to @braddialpad for official pr approval.

components/avatar/avatar.vue
```diff
@@ -103,36 +94,108 @@ export default {
...
+            this.formatInitials(this.initials);
+          }
+        }
+
+        if (this.kind === 'initials') {
+          this.formatInitials(firstChild.text || firstChild.textContent);
         }
-        this.validateImageAttrsPresence();
+      }
+    },
```
braddialpad: I think we should actually try to reduce the complexity of this by adding more functions instead of ignoring the linter error. My other comments should help with this as well.

components/avatar/avatar.vue
```diff
@@ -103,36 +94,108 @@ export default {
...
 
   methods: {
-    init () {
-      const firstChild = this.$el.firstChild;
-      if (firstChild) {
-        this.setKind(firstChild);
-        if (this.kind === 'image') {
-          firstChild.classList.add('d-avatar__image');
+    async init () {
+      const firstChild = this.$slots.default[0]?.elm || this.$slots.default[0];
```
braddialpad: What is `.elm` here?

components/avatar/avatar.vue
```diff
@@ -103,36 +94,108 @@ export default {
...
+      const iconChild = this.$slots.default[0]?.componentOptions?.tag === 'dt-icon';
```
braddialpad: isIcon?

components/avatar/avatar.vue
braddialpad: hasDefaultSlot?

components/avatar/avatar.vue
```diff
@@ -103,36 +94,108 @@ export default {
...
+
+      if (firstChild || iconChild) {
```
braddialpad: we can probably just return early instead of nesting the whole thing in an `if`

`if (!firstChild && !iconChild) return;`

or we can maybe just check firstChild, we'd never have an iconChild if there was no firstChild right?

components/avatar/avatar.vue
```diff
@@ -103,36 +94,108 @@ export default {
...
+          });
+
+          if (this.initials) {
+            this.formatInitials(this.initials);
+          }
+        }
+
+        if (this.kind === 'initials') {
+          this.formatInitials(firstChild.text || firstChild.textContent);
         }
```
braddialpad: I think you only need to call `this.formatInitials` once at the end. Use a variable to store what you're passing into it based on the above logic

components/avatar/avatar.vue
```diff
@@ -143,21 +206,53 @@ export default {
     },
 
     isSvgType (element) {
-      return element?.tagName?.toUpperCase() === 'SVG';
+      return element?.tagName?.toUpperCase() === 'SVG' || element?.tag?.toUpperCase() === 'SVG';
```
braddialpad: maybe something like...

```js
element = element.$el || element
return element?.tagName?.toUpperCase() === 'SVG'
```

components/avatar/avatar.vue
```diff
@@ -143,21 +206,53 @@ export default {
...
     },
 
     isImageType (element) {
-      return element?.tagName?.toUpperCase() === 'IMG';
+      return element?.tagName?.toUpperCase() === 'IMG' || element?.tag?.toUpperCase() === 'IMG';
```
braddialpad: same deal here

braddialpad (changes requested):
Just a few optimizations on the image load stuff

components/avatar/avatar.vue
josedialpad: It's for getting the html in the slot since the default slot isn't rendered until the kind is defined. Since it's an internal prop from the vnode, will try to refactor just using the exposed `$el` prop

braddialpad: alright cool, I just had never seen it before was curious what it was.

components/avatar/avatar.vue
josedialpad: Yeah I agree, I did a refactor more similar to the previous approach and seems to be less complex

components/avatar/avatar.vue
josedialpad: The challenge found here was the logic for showing initials/fallback to initials when the image doesn't load and having all the variants supported by the default slot. If it was supported by a prop and thus reactive, things might be easier.

components/avatar/avatar.vue
josedialpad: Updated the approach using `$el` as in the previous approach so this line isn't needed anymore

components/avatar/avatar.vue
josedialpad: Updated the approach using `$el` as in the previous approach so this line isn't needed anymore

components/avatar/avatar.vue
josedialpad: Updated the approach using `$el` as in the previous approach so this line isn't needed anymore

components/avatar/avatar.vue
josedialpad: `formatInitials` is called twice inside the `init` method now. One in case the image fails to load, which means it'll fallback to initials and the other is just when the avatar is initial type.

components/avatar/avatar.vue
josedialpad: rollbacked this change, with the refactor just using `element?.tagName?.toUpperCase() === 'SVG'` is enough

components/avatar/avatar.vue
josedialpad: rollbacked this change, with the refactor just using `element?.tagName?.toUpperCase() === 'IMG'` is enough

josedialpad:
@braddialpad I updated the PR and did a refactor following a similar approach of the previous implementation rendering the default slot when mounting the component so it uses `this.$el` for determining the kind, and applying the logic for initials, image, etc.

braddialpad (approved):
Looks much cleaner, thanks!

Future update idea, add a prop to pass in a seed so we could get the same unique gradient every time for a user.

francisrupert:
@josedialpad Sweet! Go go go! :)

francisrupert (approved):

PR #626, opened 2022-11-29
feat(combobox): add label

iropolo:
# Add label to combobox component
 
We need to be able to add a Label to `combobox` `combobox_multi_select` `combobox_with_popover`
We also need to have a `labelVisible` prop to show or hide it.

Jira: [DT-807](https://dialpad.atlassian.net/browse/DT-807)

- [x] Feature

## :book: Description

**Note**: this is breaking change since we change the label prop on input component to requried.

## :pencil: Checklist

- [x] I have added tests

image


components/combobox/combobox.test.js
```diff
@@ -89,6 +92,30 @@ describe('DtCombobox Tests', function () {
...
 
+    describe('When label is provided', function () {
+      beforeEach(async function () {
+        scopedSlots = { input: '<input v-bind="props.inputProps" />' };
+        _mountWrapper();
+        _setChildWrappers();
+      });
+
+      it('should provide proper props to input element', function () {
+        console.log(wrapper.find('input').attributes());
```
braddialpad: console.log

components/combobox/combobox.stories.js
```diff
@@ -3,9 +3,12 @@ import { createTemplateFromVueFile } from '@/common/storybook_utils';
 import DtCombobox from './combobox';
 import DtComboboxMdx from './combobox.mdx';
 import DtComboboxDefaultTemplate from './combobox_default.story.vue';
+import { LABEL_SIZES } from './combobox_constants';
 
 // Default Prop Values
 export const argsData = {
+  label: 'Label Text',
+  labelVisible: true,
```
braddialpad, line 11: We need to add labelVisible to the input component as well, as specified in the story.

components/combobox/combobox.vue
```diff
@@ -79,13 +80,40 @@ export default {
...
+
+    /**
+     * Size of the input, one of `xs`, `sm`, `md`, `lg`, `xl`
+     * @values null, xs, sm, md, lg, xl
+     */
+    labelSize: {
+      type: String,
+      default: null,
+      validator: (t) => Object.values(LABEL_SIZES).includes(t),
+    },
```
braddialpad: This should not be label size but rather just "size". From the story:

> Prop to change size of the input which would then change the size of the label (e.g. d-label-base, d-label-small, etc)

Should work fine if you just pass it down to the input component.

recipes/comboboxes/combobox_with_popover/combobox_with_popover.test.js
```diff
@@ -117,6 +120,30 @@ describe('DtRecipeComboboxWithPopover Tests', function () {
...
+        assert.equal(wrapper.find('input').attributes('description'), basePropsData.description);
+      });
+
+      describe('If `labelVisible` prop is `false`', function () {
+        beforeEach(async function () {
+          await wrapper.setProps({ labelVisible: false });
+        });
+        it('label should be empty', function () {
+          assert.equal(wrapper.find('input').attributes('label'), '');
+        });
```
braddialpad: Maybe also check that the aria-label is still set even if labelVisible is false since that could be something that's easily missed.

braddialpad (changes requested):
Looks pretty good, just make sure to work alongside the requirements in the story and make sure they are all covered.

recipes/comboboxes/combobox_with_popover/combobox_with_popover_default.story.vue
```diff
@@ -1,9 +1,13 @@
 <template>
   <dt-recipe-combobox-with-popover
     ref="comboboxWithPopover"
+    :label="label"
+    :label-visible="labelVisible"
+    :label-size="labelSize"
+    :description="description"
+    :size="labelSize"
```
braddialpad: labelSize passing to two different props?

components/combobox/combobox.test.js
iropolo: could we add 
`'no-console': ['warn', { allow: ['error', 'info'] }]`

 on our eslint file?

components/combobox/combobox.vue
iropolo: On multiselect we already have size prop (for chip)
How we can do it there? with labelSize or we should update size to chipSize?
I suggest update it to chipSize and use size only for the input (same of combobox and combobox_with_popover)

components/combobox/combobox.stories.js
iropolo: Label on input component should be required also?

recipes/comboboxes/combobox_with_popover/combobox_with_popover.test.js
iropolo: I move this test to input component and added aria-label test here

components/combobox/combobox.stories.js
braddialpad: Yes

components/combobox/combobox.test.js
```diff
@@ -89,6 +92,29 @@ describe('DtCombobox Tests', function () {
...
+        scopedSlots = { input: '<input v-bind="props.inputProps" />' };
+        _mountWrapper();
+        _setChildWrappers();
+      });
+
+      it('should provide proper props to input element', function () {
+        assert.equal(wrapper.find('input').attributes('label'), basePropsData.label);
+        assert.equal(wrapper.find('input').attributes('size'), basePropsData.size);
+        assert.equal(wrapper.find('input').attributes('description'), basePropsData.description);
+      });
```
braddialpad: Should ideally have one assert per 'it'. so in this case one for each prop. I know it's a little tedious but it is more maintainable.

recipes/comboboxes/combobox_multi_select/combobox_multi_select.test.js
```diff
@@ -83,6 +84,21 @@ describe('DtRecipeComboboxMultiSelect Tests', function () {
   describe('Presentation Tests', function () {
     it('should render the component', function () { assert.exists(wrapper, 'wrapper exists'); });
     it('should render the input', function () { assert.isTrue(input.exists()); });
+    it('should render the input label', function () {
+      assert.isTrue(wrapper.find('[data-qa="dt-input-label"]').exists());
+    });
+    it('should not render label if `labelVisible` is false', async function () {
+      await wrapper.setProps({ labelVisible: false });
+      assert.isFalse(wrapper.find('[data-qa="dt-input-label"]').exists());
```
braddialpad: Change state in "describes", assert in "it".

in this case:

```
describe('when labelVisible prop is false', async function () {
  // beforeEach here to set labelVisible: false
  it('should not render label', async function () {
      assert.isFalse(wrapper.find('[data-qa="dt-input-label"]').exists());
  });
}});
```

As a rule if you have to write "if something" in your it statement you're probably need to rework it.

recipes/comboboxes/combobox_multi_select/combobox_multi_select.test.js
```diff
@@ -83,6 +84,21 @@ describe('DtRecipeComboboxMultiSelect Tests', function () {
...
+    });
+    it('should still set aria-label even if label visible is false', async function () {
+      await wrapper.setProps({ labelVisible: false });
+      assert.equal(wrapper.find('input').attributes('aria-label'), basePropsData.label);
```
braddialpad: all `wrapper.find` statements should be set in the `_setChildWrappers` function. You can just recall this function if needed.

recipes/comboboxes/combobox_multi_select/combobox_multi_select.vue
```diff
@@ -124,19 +126,38 @@ export default {
...
+    },
+
+    /**
+     * Size of the input, one of `xs`, `sm`, `md`, `lg`, `xl`
+     * @values null, xs, sm, md, lg, xl
+     */
+    labelSize: {
       type: String,
       default: null,
+      validator: (t) => Object.values(LABEL_SIZES).includes(t),
```
braddialpad: Do we need this at all? this component already has a "size" prop that sets the size of the input

recipes/comboboxes/combobox_with_popover/combobox_with_popover.test.js
```diff
@@ -117,6 +120,21 @@ describe('DtRecipeComboboxWithPopover Tests', function () {
...
+        scopedSlots = { input: '<input v-bind="props.inputProps" />' };
+        _mountWrapper();
+        _setChildWrappers();
+      });
+
+      it('should provide proper props to input element', function () {
+        assert.equal(wrapper.find('input').attributes('label'), basePropsData.label);
+        assert.equal(wrapper.find('input').attributes('size'), basePropsData.size);
+        assert.equal(wrapper.find('input').attributes('description'), basePropsData.description);
+        assert.equal(wrapper.find('input').attributes('aria-label'), basePropsData.label);
```
braddialpad: same issue with multiple asserts

braddialpad:
> could we add 
`'no-console': ['warn', { allow: ['error', 'info'] }]`

 on our eslint file? 

Yeah.. I thought was already there actually..

braddialpad:
> On multiselect we already have size prop (for chip)
How we can do it there? with labelSize or we should update size to chipSize?
I suggest update it to chipSize and use size only for the input (same of combobox and combobox_with_popover) 

Size sets the size of the input and the chip so it can just pass through to input

components/combobox/combobox.vue
```diff
@@ -237,7 +270,7 @@ export default {
         // The list has to be positioned relatively so that the auto-scroll can
         // calculate the correct offset for the list items.
         class: 'd-ps-relative',
-        'aria-label': this.listAriaLabel,
+        'aria-label': this.label,
```
braddialpad, line 273: You can remove this the listbox doesn't need an aria label.

braddialpad (approved):
Nice looks good, thanks for the fixes. Just need to remove aria-label from the listbox and it's good to go.

components/combobox/combobox.vue
iropolo: I can see an error on a11y test now related to `role="listbox"` I believe its because this update.
Its seems its necessary to have this `aria-label` or... Im missing sth?

braddialpad: My mistake, this does actually need an aria-label (thanks axe-core for catching this!)

braddialpad (approved):

PR #629, opened 2022-12-01
feat(collapsible): update `DtCollapsible` to show content when collapsed

modialpad:
# Update `DtCollapsible` to show content when collapsed

This use-case is needed in several parts of our new app. We need to be able to show content when collapsible is collapsed.

Jira: https://dialpad.atlassian.net/browse/DT-843

- [x] Feature

## :book: Description

**Note**: this is breaking change since we change the slot name from `content` to `contentOnExpanded` to indicate clearly that this is content that is shown when expanded, as opposed to `contentOnCollapsed` which will be shown only when the content is collapsible is collapsed.

## :bulb: Context

Jira: https://dialpad.atlassian.net/browse/DT-843

## :pencil: Checklist

- [x] I have added tests

Collapsible Update GIF

### DIFF
--- a/components/collapsible/collapsible.stories.js
+++ b/components/collapsible/collapsible.stories.js
@@ -16,7 +16,15 @@ const argTypesData = {
       },
     },
   },
-  content: {
+  contentOnCollapsed: {
+    control: 'text',
+    table: {
+      type: {
+        summary: 'VNode',
+      },
+    },
+  },
+  contentOnExpanded: {
     control: 'text',
     table: {
       type: {
--- a/components/collapsible/collapsible.stories.js
+++ b/components/collapsible/collapsible.stories.js
@@ -90,6 +90,12 @@ const DefaultTemplate = (args, { argTypes }) => createTemplateFromVueFile(
   DtCollapsibleDefaultStory,
 );
 
+// Stories
+export const Default = DefaultTemplate.bind({});
+Default.args = {
+  maxWidth: '512px',
+};
+
 const VariantTemplate = (args, { argTypes }) => createTemplateFromVueFile(
   args,
   argTypes,
@@ -97,9 +103,3 @@ const VariantTemplate = (args, { argTypes }) => createTemplateFromVueFile(
 );
 
 export const Variants = VariantTemplate.bind({});
-
-// Stories
-export const Default = DefaultTemplate.bind({});
-Default.args = {
-  maxWidth: '512px',
-};
--- a/components/collapsible/collapsible.stories.js
+++ b/components/collapsible/collapsible.stories.js
@@ -1,9 +1,10 @@
+import { createTemplateFromVueFile } from '@/common/storybook_utils';
 import {
   DtCollapsible,
 } from './';
-import { createTemplateFromVueFile } from '@/common/storybook_utils';
 import CollapsibleMdx from './collapsible.mdx';
 import DtCollapsibleDefaultStory from './collapsible_default.story.vue';
+import DtCollapsibleVariantsStory from './collapsible_variants.story.vue';
 
 const argTypesData = {
   // Slots
@@ -80,6 +81,14 @@ const DefaultTemplate = (args, { argTypes }) => createTemplateFromVueFile(
   argTypes,
   DtCollapsibleDefaultStory,
 );
+
+const VariantTemplate = (args, { argTypes }) => createTemplateFromVueFile(
+  args,
+  argTypes,
+  DtCollapsibleVariantsStory,
+);
+
+export const Variants = VariantTemplate.bind({});
 
 // Stories
 export const Default = DefaultTemplate.bind({});
--- a/components/collapsible/collapsible.test.js
+++ b/components/collapsible/collapsible.test.js
@@ -23,16 +23,8 @@ describe('Dialtone vue Collapsible Component Tests', function () {
   let anchorSlotElement;
 
   // Environment
-  let slots = { contentOnExpanded, contentOnCollapsed };
-  let propsData = baseProps;
-
-  const _clearChildWrappers = () => {
-    contentWrapperElement = undefined;
-    contentOnExpandedElement = undefined;
-    contentOnCollapsedElement = undefined;
-    anchorElement = undefined;
-    anchorSlotElement = undefined;
-  };
+  const defaultSlots = { contentOnExpanded, contentOnCollapsed };
+  const defaultPropsData = baseProps;
 
   const findAllElements = () => {
     anchorElement = wrapper.find('[data-qa="dt-button"]');
@@ -42,7 +34,7 @@ describe('Dialtone vue Collapsible Component Tests', function () {
     contentWrapperElement = wrapper.findComponent({ ref: 'contentWrapper' });
   };
 
-  const mountWrapper = () => {
+  const mountWrapper = ({ propsData = defaultPropsData, slots = defaultSlots } = {}) => {
     wrapper = mount(DtCollapsible, {
       propsData,
       slots,
@@ -68,12 +60,6 @@ describe('Dialtone vue Collapsible Component Tests', function () {
     config.renderStubDefaultSlot = false;
   });
 
-  afterEach(async function () {
-    propsData = baseProps;
-    slots = { contentOnExpanded, contentOnCollapsed };
-    _clearChildWrappers();
-  });
-
   describe('Test default rendering', function () {
     it('should render the component', function () {
       assert.isTrue(wrapper.exists(), 'wrapper exists');
@@ -94,8 +80,9 @@ describe('Dialtone vue Collapsible Component Tests', function () {
 
   describe('When scoped slot is provided for anchor', function () {
     beforeEach(function () {
-      slots = { anchor };
-      mountWrapper();
+      mountWrapper({
+        slots: { anchor },
+      });
     });
 
     it('should render the anchor slot', function () {
@@ -105,8 +92,9 @@ describe('Dialtone vue Collapsible Component Tests', function () {
 
   describe('When scoped slot is not provided for any content', function () {
     beforeEach(function () {
-      slots = { };
-      mountWrapper();
+      mountWrapper({
+        slots: {},
+      });
     });
 
     it('should NOT render any content if no slot is passed', function () {
@@ -130,19 +118,15 @@ describe('Dialtone vue Collapsible Component Tests', function () {
         assert.isFalse(contentOnExpandedElement.exists());
         assert.isTrue(contentOnCollapsedElement.isVisible());
       });
-    });
-
-    describe('when user clicks anchor when it is collapsed', function () {
-      beforeEach(async function () {
-        // need to trigger twice since the initial state of the
-        // wrapper is expanded. We want it to be collapsed initially.
-        await anchorElement.trigger('click');
-        await anchorElement.trigger('click');
-      });
-      it('should be expanded', function () {
-        findAllElements();
-        assert.isTrue(contentOnExpandedElement.isVisible());
-        assert.isFalse(contentOnCollapsedElement.exists());
+      describe('when user clicks anchor when it is collapsed', function () {
+        beforeEach(async function () {
+          await anchorElement.trigger('click');
+        });
+        it('should be expanded', function () {
+          findAllElements();
+          assert.isTrue(contentOnExpandedElement.isVisible());
+          assert.isFalse(contentOnCollapsedElement.exists());
+        });
       });
     });
   });
@@ -182,11 +166,12 @@ describe('Dialtone vue Collapsible Component Tests', function () {
 
     describe('Test open prop set to false', function () {
       beforeEach(async function () {
-        propsData = {
-          ...baseProps,
-          open: false,
-        };
-        mountWrapper();
+        mountWrapper({
+          propsData: {
+            ...baseProps,
+            open: false,
+          },
+        });
       });
 
       it('content starts collapsed', function () {
@@ -224,8 +209,9 @@ describe('Dialtone vue Collapsible Component Tests', function () {
 
     beforeEach(async function () {
       consoleErrorSpy = sinon.spy(console, 'error');
-      propsData = { ...baseProps, anchorText: undefined };
-      mountWrapper();
+      mountWrapper({
+        propsData: { ...baseProps, anchorText: undefined },
+      });
     });
 
     afterEach(function () {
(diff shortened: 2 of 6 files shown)


components/collapsible/collapsible.vue
```diff
@@ -8,6 +8,7 @@
     <div
       :id="!ariaLabelledBy && labelledBy"
       ref="anchor"
+      :aria-label="ariaLabel"
       :class="[
         'd-dt-collapsibe__anchor',
```
braddialpad: Might want to try merging master into this branch and make sure there are no conflicts. There was a fix we made to collapsible keyboard functionality the other day https://github.com/dialpad/dialtone-vue/commit/39a3f7608646d7095b3a98079c69664e978b2e3b. Also includes fixing this typo haha.

components/collapsible/collapsible_lazy_show.vue
```diff
@@ -106,6 +107,11 @@ export default {
      */
     afterEnter (element) {
       element.style.height = null;
+      // Note: since the mode out transition is "out-in"
+      // the 'enter' event will be the one triggered last in this
+      // transition. It will mark the end of the transition, hence
+      // the trigger of the event below
```
braddialpad: Well that is damn confusing. Thanks for the comment explaining it though.

components/collapsible/collapsible.test.js
```diff
@@ -67,45 +70,72 @@ describe('Dialtone vue Collapsible Component Tests', function () {
...
+
+  describe('When scoped slot is not provided for any content', function () {
+    beforeEach(function () {
+      slots = { };
+      _mountWrapper();
+    });
+
+    it('should NOT render any content if no slot is passed', function () {
+      contentOnCollapsedElement = wrapper.find('[data-qa="content-on-collapsed-element"]');
+      contentOnExpandedElement = wrapper.find('[data-qa="content-on-expanded-element"]');
```
braddialpad: We already have this declared in `_setChildWrappers()`. If you need to re-run the find just call `_setChildWrappers()`

components/collapsible/collapsible.test.js
```diff
@@ -67,45 +70,72 @@ describe('Dialtone vue Collapsible Component Tests', function () {
...
+      assert.isFalse(contentOnExpandedElement.exists());
+      assert.isFalse(contentOnCollapsedElement.exists());
     });
   });
 
   describe('Test open prop undefined', function () {
```
braddialpad, line 106: Shouldn't this just be under "Test default rendering" since we are not changing the state?

components/collapsible/collapsible.test.js
```diff
@@ -115,18 +145,24 @@ describe('Dialtone vue Collapsible Component Tests', function () {
         await wrapper.setProps({ open: true });
       });
 
-      it('content is expanded', function () {
-        assert.isTrue(contentElement.isVisible());
+      it('content is expanded when open = true', function () {
```
braddialpad: Don't need to describe state here since you've already done it in the parent describe. Just "content is expanded".

components/collapsible/collapsible.test.js
```diff
@@ -115,18 +145,24 @@ describe('Dialtone vue Collapsible Component Tests', function () {
...
         await anchorElement.trigger('click');
-        assert.isTrue(contentElement.isVisible());
+        contentOnExpandedElement = wrapper.find('[data-qa="content-on-expanded-element"]');
+        contentOnCollapsedElement = wrapper.find('[data-qa="content-on-collapsed-element"]');
+        assert.isTrue(contentOnExpandedElement.isVisible());
+        assert.isFalse(contentOnCollapsedElement.exists());
       });
 
       it('updating open prop does collapse content', async function () {
         await wrapper.setProps({ open: false });
```
braddialpad: Setting state should be done in describes, ideally 'it' is just an assertion if possible. More on this here: https://vue.dialpad.design/?path=/story/docs-component-driven-development-unit-tests--page

This goes for most of the tests in this PR. It is far more maintainable and scalable to write tests in this way.

Example:

```
describe('When open prop is set to true', function () {
  describe('When anchor is clicked', function () {
    // beforeEach here to perform click
    it('does not collapse content', function () {
      // assertion
    });
  });
});
```

components/collapsible/collapsible.test.js
```diff
@@ -171,6 +214,44 @@ describe('Dialtone vue Collapsible Component Tests', function () {
   });
 
   describe('Accessibility Tests', function () {
+    describe('Anchor aria-label', function () {
+      it('should should correctly set the aria-label attribute on anchor', async function () {
```
braddialpad: two 'shoulds' typo

components/collapsible/collapsible.vue
```diff
@@ -8,6 +8,7 @@
     <div
       :id="!ariaLabelledBy && labelledBy"
       ref="anchor"
+      :aria-label="ariaLabel"
```
braddialpad: Shouldn't this be on the button itself?

braddialpad (changes requested):
Thanks! Looks pretty good. Mostly testing issues.

components/collapsible/collapsible.vue
modialpad: I wasn't sure about this from an accessibility POV - I know the button already has a label inside of it, so does it make sense to add an `aria-label` to it that's different from the visible text already inside it? 

I used VoiceOver to test this and indeed the tool only announces the aria-label in this case, so I'm inclined to think that's fine.

components/collapsible/collapsible.test.js
modialpad: Removed

components/collapsible/collapsible.test.js
modialpad: Done

components/collapsible/collapsible.test.js
modialpad: Updated, I renamed this method to `findAllElements`.

components/collapsible/collapsible.vue
modialpad: I rebased this on staging, do you see any conflicts?

components/collapsible/collapsible.test.js
modialpad: Sure, but one issue I noticed with this approach is when multiple user actions are needed for a specific test. For example, for one test, I wanted to test the user clicking the anchor multiple times and perform assertions based on that. I usually did that sequentially in an `it` block, like so:

```js
it('test', async function () {
         await anchorElement.trigger('click');
        // test stuff
        await anchorElement.trigger('click'); // click again
       // test more stuff
})
```

In a describe, it would look a bit awkward:

```js
describe('when user clicks anchor when it is collapsed', function () {
      beforeEach(async function () {
        // need to trigger twice since the initial state of the
        // wrapper is expanded. We want it to be collapsed initially.
        await anchorElement.trigger('click');
        await anchorElement.trigger('click');
      });
      it('should be expanded', function () {
        findAllElements();
        assert.isTrue(contentOnExpandedElement.isVisible());
        assert.isFalse(contentOnCollapsedElement.exists());
      });
    });

```

components/collapsible/collapsible.test.js
modialpad: I'm actually setting the default (in `afterEach`) to set both `contentOnCollapsed` and `contentOnExpanded`, that's why I had to create a separate test where both are undefined.

components/collapsible/collapsible.vue
braddialpad: Yes you're right that the visible text should be read by default. You would only set aria label if the text you wish to be read is different than the text that's actually in the button.

components/collapsible/collapsible.test.js
braddialpad: You do have a way of setting the initial expanded state via the `open` prop. I'd recommend just using that to render the desired initial state and then clicking to test the trigger. Alternatively you could have a nested describe for each click and have one test to check whether it's collapsed, and one test to check whether it's expanded.

components/collapsible/collapsible.test.js
braddialpad: Yeah I'm a bit confused why we are doing this in `afterEach` instead of `beforeEach`. Generally we set state like this in `beforeEach` before the `mountWrapper();`

components/collapsible/collapsible.vue
braddialpad: nope seems fine

components/collapsible/collapsible.test.js
modialpad: I opted for the latter, the thing with this test is that we're trying to make sure that the behaviour works correctly when the `open` prop is undefined, so it wouldn't work if we explicitly set the `open` prop here.

I just nested the `describe`s as you mentioned and it works.

components/collapsible/collapsible.test.js
modialpad: Agreed, I don't think we need the `afterEach` here. I removed it and updated the `mountWrapper` to contain prop/slot defaults.

modialpad:
@braddialpad BTW this is a breaking change because we're changing the slot API of `DtCollapsible`, is there a process for dealing with this before releasing a new version?

braddialpad:
> @braddialpad BTW this is a breaking change because we're changing the slot API of `DtCollapsible`, is there a process for dealing with this before releasing a new version?

Normally yes, but there are zero usages of DtCollapsible in firespotter currently so go ahead and break it!

components/collapsible/collapsible.stories.js
```diff
@@ -1,9 +1,10 @@
+import { createTemplateFromVueFile } from '@/common/storybook_utils';
 import {
   DtCollapsible,
 } from './';
-import { createTemplateFromVueFile } from '@/common/storybook_utils';
 import CollapsibleMdx from './collapsible.mdx';
 import DtCollapsibleDefaultStory from './collapsible_default.story.vue';
+import DtCollapsibleVariantsStory from './collapsible_variants.story.vue';
 
 const argTypesData = {
```
braddialpad, line 9: remove the old "content" slot from the argTypesData, it's making this weird uncategorized field show up in storybook

braddialpad: Then add the new slots `contentOnCollapsed` `contentOnExpanded` and make their control 'text' and summary 'VNode' as the other slots below.

components/collapsible/collapsible.vue
```diff
@@ -64,13 +64,20 @@
...
       v-on="$listeners"
-      @after-leave="onLeaveTransitionComplete"
-      @after-enter="onEnterTransitionComplete"
+      @transitionfinished="onTransitionFinished"
     >
-      <!-- @slot Slot for the collapsible element that is expanded by the anchor -->
-      <slot
-        name="content"
-      />
+      <!-- @slot Slot for content that is shown when collapsible is expanded -->
```
braddialpad: Put these @slot comments above the `<slot>` rather than the `<template>` or else they don't show up in storybook

braddialpad (approved):
Couple final storybook cleanup things but good to go once you do that. Pre-approving

braddialpad:
That transition looks really nice!

components/collapsible/collapsible.vue
modialpad: Done

components/collapsible/collapsible.stories.js
modialpad: Done

modialpad:
> That transition looks really nice!

@braddialpad Just a note on the transition. In cases where you have only `contentOnExpanded`, I see some delay (~500ms >) between the time when you click to expand and when the content actually shows up. You can see this in the `default` story. This is because we've set the transition to be `out-in` and we always create the top-level component element for cases even when `contentOnCollapsed` is not defined, so the animation duration stills runs even when you don't have collapsed content.

The way to fix this is by conditionally rendering these wrapper elements depending on whether the underlying slot (`contentOnExpanded` or `contentOnCollapsed`) exists, like so:

```vue
// collapsible_lazy_show.vue
...
    <component
      :is="elementType"
      v-if="(isExpanded && $slots.contentOnExpanded)" // <--- new check here
      key="onOpen"
      v-on="$listeners"
    >
      <!-- @slot slot for Content when collapsible is expanded -->
      <slot name="contentOnExpanded" />
    </component>
    <component
      :is="elementType"
      v-else-if="(!isExpanded && $slots.contentOnCollapsed)"    // <--- new check here
      key="onClose"
      v-on="$listeners"
    >
      <!-- @slot slot for Content when collapsible is collapsed -->
      <slot name="contentOnCollapsed" />
    </component>

```

I have some other changes to the animation transition handlers we were settings (I don't think we need those), and fixed the events emitted when `open` is updated. Here's a PR that contains the diffs I'm proposing: https://github.com/dialpad/dialtone-vue/pull/631

Could you take a look and let me know if it's good to incorporate these changes here?

braddialpad:
> Could you take a look and let me know if it's good to incorporate these changes here?

Cool thanks, not sure I would have noticed if you hadn't mentioned it haha. I left some comments on the PR

components/collapsible/collapsible_default.story.vue
```diff
@@ -23,7 +23,7 @@
         v-html="anchor"
       />
     </template>
-    <template slot="content">
+    <template #contentOnExpanded>
       <div
         v-if="content"
         v-html="content"
```
braddialpad: Just realized these "content" need to be changed to "contentOnExpanded" or else changes in storybook are not reflected.

braddialpad: Also you'll need to add another slot template for contentOnCollapsed like so

```
<template #contentOnCollapsed>
  <div
    v-if="contentOnCollapsed"
    v-html="contentOnCollapsed"
  />
</template>
```

components/collapsible/collapsible.stories.js
```diff
@@ -81,6 +90,14 @@ const DefaultTemplate = (args, { argTypes }) => createTemplateFromVueFile(
   DtCollapsibleDefaultStory,
 );
 
+const VariantTemplate = (args, { argTypes }) => createTemplateFromVueFile(
+  args,
+  argTypes,
+  DtCollapsibleVariantsStory,
+);
+
+export const Variants = VariantTemplate.bind({});
```
braddialpad, line 105: Put the variants below the default as this is the order it displays in storybook

braddialpad (commented):
Couple more things I noticed while messing around with this

components/collapsible/collapsible_default.story.vue
modialpad: Done

components/collapsible/collapsible.stories.js
modialpad: Done in that other PR

PR #631, opened 2022-12-06
WIP: Fix Collapsible animation

modialpad:
# PR Title

- [ ] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :pencil: Checklist

- [ ] I have added tests

### DIFF
--- a/components/collapsible/collapsible.vue
+++ b/components/collapsible/collapsible.vue
@@ -64,7 +64,6 @@
       tabindex="-1"
       appear
       v-on="$listeners"
-      @transitionfinished="onTransitionFinished"
     >
       <template #contentOnExpanded>
         <!-- @slot Slot for content that is shown when collapsible is expanded -->
@@ -253,21 +252,17 @@ export default {
 
   methods: {
 
-    onTransitionFinished () {
-      this.$emit('opened', this.isOpen);
-      if (this.open !== null) {
-        this.$emit('update:open', this.isOpen);
-      }
-    },
-
     defaultToggleOpen () {
       if (this.open === null) {
         this.toggleOpen();
+      } else {
+        this.$emit('update:open', !this.isOpen);
       }
     },
 
     toggleOpen () {
       this.isOpen = !this.isOpen;
+      this.$emit('opened', this.isOpen);
     },
 
     validateProperAnchor () {
--- a/components/collapsible/collapsible.vue
+++ b/components/collapsible/collapsible.vue
@@ -253,16 +253,12 @@ export default {
   methods: {
 
     defaultToggleOpen () {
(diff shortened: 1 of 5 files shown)


components/collapsible/collapsible_variants.story.vue
```diff
@@ -1,10 +1,13 @@
 <template>
   <div>
+    <p>Collapsible is open: {{ isOpenLocal }}</p>
     <dt-collapsible
       anchor-text="Favorites"
       aria-label="Favorites, click to expand or collapse"
       content-element-type="ol"
       class="d-p0 d-w100p"
+      :open="isOpenLocal"
+      @update:open="newValue => isOpenLocal = newValue"
```
braddialpad: We should be able to just use `open.sync="isOpenLocal"`

components/collapsible/collapsible_lazy_show.vue
```diff
@@ -78,7 +82,15 @@ export default {
...
-  emits: ['transitionfinished'],
+  computed: {
+    hasContentOnCollapsed () {
+      return !!this.$slots.contentOnCollapsed;
+    },
+
+    hasContentOnExpanded () {
+      return !!this.$slots.contentOnExpanded;
+    },
+  },
```
braddialpad: I think you need to put these directly in the template since this.$slots is not reactive, either that or use a method.

components/collapsible/collapsible.vue
```diff
@@ -253,21 +252,17 @@ export default {
...
       if (this.open === null) {
         this.toggleOpen();
+      } else {
+        this.$emit('update:open', !this.isOpen);
       }
     },
 
     toggleOpen () {
       this.isOpen = !this.isOpen;
+      this.$emit('opened', this.isOpen);
```
braddialpad: I think there's a misunderstanding in `opened` vs `update:open` here (it is a bit confusing). 

- `update:open` is purely for `open.sync` and is hidden from the public interface. It should only be triggered when open is not null. Current behaviour is correct.
- `opened` should be triggered every time the collapsible is opened/closed regardless of `this.open` value. Currently it only emits when `this.open === null`

modialpad: I updated it to trigger the event before that specific logic.

components/collapsible/collapsible_lazy_show.vue
modialpad: Added in the template

components/collapsible/collapsible_variants.story.vue
modialpad: Done

braddialpad (approved):

PR #632, opened 2022-12-06
refactor: remove extra attrs from combobox wrapper

MikaDialpad:
# refactor: remove extra attrs from combobox wrapper

- [ ] Fix
- [ ] Feature
- [x] Refactoring
- [ ] Documentation

## :book: Description

The vue2 version of this component was refactored a while ago and these attrs were moved onto the input. For the vue3 version these were not removed from the wrapper, so it was having duplicated attributes. This change unifies the vue2 and vue3 versions and after this change the code is exactly the same for both versions.

### DIFF
--- a/components/combobox/combobox.vue
+++ b/components/combobox/combobox.vue
@@ -1,11 +1,5 @@
-<!-- eslint-disable vuejs-accessibility/interactive-supports-focus -->
 <template>
   <div
-    role="combobox"
-    :aria-expanded="showList.toString()"
-    :aria-controls="listId"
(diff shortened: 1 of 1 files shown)


MikaDialpad:
Here's how VO currently reads this. Notice the menu box is duplicated
image

And here's how it's after the fix. I assume it doesn't "highlight" the entire combobox anymore since there's no role on it
image

braddialpad (approved):

braddialpad:
Yeah that is definitely duplicating what's in the input. Not what it should be doing.

PR #634, opened 2022-12-07
feat(collapsible): collapsible updates to vue3

modialpad:
# Update Collapsible with recent changes to Vue3 branch

More info here: https://github.com/dialpad/dialtone-vue/pull/629

- [x] Feature

## :book: Description

Note: this is breaking change since we change the slot name from content to contentOnExpanded to indicate clearly that this is content that is shown when expanded, as opposed to contentOnCollapsed which will be shown only when the content is collapsible is collapsed.

## :pencil: Checklist

- [x] I have added tests

### DIFF
--- a/components/collapsible/collapsible.mdx
+++ b/components/collapsible/collapsible.mdx
@@ -9,11 +9,13 @@ import { Canvas, Story, Subtitle, ArgsTable, PRIMARY_STORY } from '@storybook/ad
 
 ## Basic Style
 
-The collapsible has two components. The anchor and the content.
+The collapsible has three components. The anchor, contentOnExpanded, and contentOnCollapsed.
 
 The **anchor** will default to a basic button with some text if the slot is not filled.
 
-The **content** must be provided and is the element that can be hidden or shown when the anchor is clicked.
+The **contentOnExpanded** is the element that is shown when the anchor is expanded.
+
+The **contentOnCollapsed** is the element that is shown when the anchor is collapsed.
 
 <Canvas>
   <Story id="components-collapsible--default" />
@@ -34,7 +36,7 @@ import { DtCollapsible } from '@dialpad/dialtone-vue';
   :anchor-text="Click Me!"
   @opened="onOpen"
 >
-  <template #content>
+  <template #contentOnExpanded>
     <p>
       This will be shown in the expanded area.
     </p>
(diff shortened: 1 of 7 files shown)


components/collapsible/collapsible.test.js
```diff
@@ -5,54 +5,53 @@ import sinon from 'sinon';
...
+  const mountWrapper = ({ propsData = defaultPropsData, slots = defaultSlots } = {}) => {
     wrapper = mount(DtCollapsible, {
-      props,
+      props: propsData,
       slots,
       attrs,
       global: {
         stubs: {
-          transition: false,
+          transition: true,
```
modialpad, line 49: @braddialpad I had to update the transition stub to `true` here, otherwise the transition element won't update its children. It's already stubbed by default in Vue 2, so I'm not sure why we set it to `false` here.

braddialpad: There's a lot of weird stuff with transitions and testing in vue 3 and vue-test-utils keeps changing how it works, so guessing that was leftover from something to do with that. Fine to change if it fixes the problem.

components/collapsible/collapsible.mdx
```diff
@@ -58,10 +60,40 @@ import { DtCollapsible, DtButton } from '@dialpad/dialtone-vue';
...
     <p>
       This will be shown in the expanded area.
     </p>
   </template>
 </dt-collapsible>
 ```
+
+### Basic Usage with `contentOnCollapsed`
+
+You can also choose to show content when under the anchor when it is collapsed. To do so,
```
braddialpad: same text change here as the other PR

components/collapsible/collapsible.vue
```diff
@@ -51,10 +48,10 @@
     <dt-collapsible-lazy-show
       :id="id"
       ref="contentWrapper"
-      :aria-hidden="`${!isOpen}`"
+      data-qa="contentWrapper"
+      :aria-hidden="`${!isOpen && !$slots.contentOnCollapsed}`"
       :aria-labelledby="labelledBy"
-      :aria-label="ariaLabel"
-      :show="isOpen"
+      :is-expanded="isOpen"
```
braddialpad, line 54: hmm doesn't appear to perform the transition from open to closed when using the "open" prop rather than just clicking directly. This doesn't seem to be a problem in Vue 2

braddialpad (commented):
This seems to still have the "delay" issue you fixed on vue 2. Was that change not integrated into this PR yet?

dbecherdialpad (changes requested):
@modialpad sorry, but we're going to have to re-think this implementation:

1. the way you have it, the slots will get entirely re-rendered every time you expand/collapse the section. so for example, if you're in 20 channels and you have 10 unread channels, when you collapse the section it will unmount all 20 and re-create the same 10, instead of doing an efficient diff.
2. you've negated the value of what `DtCollapsibleLazyShow` was doing. the point of that component is that it would only render the content once the first time it was open, then use `v-show` to hide the content. your implementation undoes that optimization which has subtle implications (for example, the state of the component in the content will be lost every time you collapse)
3. this shouldn't have been a breaking change, as it will require a migration on the firespotter side

Honestly this would have been much simpler to just make an option that the `content` slot is *always* rendered and the parent component handled conditional rendering, rather than trying to do 2 slots. Or maybe we need to revisit my feedback that DtCollapsible isn't the right component for what you're trying to do :)

components/collapsible/collapsible.vue
modialpad: That's weird, I investigated this a bit and my best guess is that it has to do with our version of Storybook for Vue 3. I added an example in our `variants` component where the collapsible `open` prop updates based on button clicks and it works without issues.

components/collapsible/collapsible.mdx
modialpad: Updated

modialpad:
> This seems to still have the "delay" issue you fixed on vue 2. Was that change not integrated into this PR yet?

It turns out Vue 3 treats slots differently now, I had to had an explicit check to determine whether there's any children or not, similar to what we're doing in [`list_section.vue`](https://github.com/dialpad/dialtone-vue/blob/8357e52a8cd535df3289a536eea33efbd5474aca/components/list_section/list_section.vue#L132-L136).

PR #637, opened 2022-12-08
feat(badge): component rework

braddialpad:
# feat(badge): component rework

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

https://dialpad.atlassian.net/browse/DT-841

Update of badge component to support the new visual refresh of badge in Dialtone (and some new functionality). See story for specific requirements.

## :bulb: Context

Better looking badges with functional styles and easier to use.

## :pencil: Checklist

- [x] I have added tests

## :crystal_ball: Next Steps

Might need a small migration in product

## :link: Sources

https://dialpad.design/components/badge.html


francisrupert (commented):
Geez this is smoove. 

### Ai Count

I see you've at least not listed Ai variant as a Count in the docs view...

image

...though it remains possible to create a `kind=count; type=ai` through https://vue.dialpad.design/deploy-previews/pr-637/?path=/story/components-badge--default&args=kind:count;type:ai;default:1

Would it be possible to inhibit that? Too much to ask if Storybook has the possibility to disable the Label `<option>` and vice versa?

image

Or perhaps a build error|warning if someone attempts...

```
<dt-badge kind="count" type="ai">badge</dt-badge>
```

### Icon selection

The icons are lit in here. Once I select one, I stumbled to figure out how to remove the icon before realizing it's the blank `<option>`. Would it be possible for that to be labeled? Something like "(none)". And if that needs to occur in other Vue components like [Button](https://user-images.githubusercontent.com/1165933/206349447-41e12079-ae4a-48e4-b705-e2cb00f32a22.png).

### Text prop

#### Capitalization

image

Teenzy weenzy nitpick. Can we sentence-case them; capitalize first character?

#### Kind=count

Upon selection of `kind="count"` would it be possible in Storybook to change the `text` value to a number and vice versa? 

Nit-picky. Not the most important thing, just looking for ways to discourage choosing the `count` kind for a non-number `text` string. For all I know that's against the grain in Storybook, and technically the `text` prop for `kind="count"` can be any alphanumeric mix (e.g. "-12%" or "5 widgets".

### Examples

In the [Docs view](https://vue.dialpad.design/deploy-previews/pr-637/?path=/docs/components-badge--default), can we replicate the examples from the [same dialpad.design section](https://dialpad.design/components/badge.html#examples)? I've already heard from several Product Designers that this helps them understand the new Kinds and Types.

image

components/badge/badge.vue
```diff
@@ -2,19 +2,42 @@
...
+    </span>
+    <span class="d-badge__label">
+      <!-- @slot Slot for badge content, defaults to text prop -->
+      <slot>
+        {{ text }}
+      </slot>
+    </span>
+    <span
+      v-if="$slots.iconRight"
+      class="d-badge__icon-right"
```
modialpad: (Just commenting here since I'll be using this in the Leftbar)

nit: are we using slots here because we anticipate different kinds of icons in badges? We can simplify this to contain only the markup for the types we've specified without any slots, unless we do have requirements for other kinds of icons (other than Ai).

francisrupert: @modialpad Both `kind`s, `label` and `count`, must support any of our icons. Dialtone docs list [some examples](https://dialpad.design/deploy-previews/pr-764/components/badge.html#examples).

braddialpad: @modialpad You're saying just have a leftIcon/rightIcon prop where you pass in the icon name rather than a slot? This is probably a good idea since this component is very fixed size. This would of course prevent any customization of the icon.. which might be what we want.

francisrupert: Ah, that makes sense when you put it that way.

braddialpad: Made the change to use prop instead of slot

common/storybook_utils.js
```diff
@@ -24,7 +24,7 @@ export const createTemplateFromVueFile = (args, argTypes, templateComponent) =>
  * @returns {string[]} icon component names
  */
 export function getIconNames () {
-  return Object.keys(dialtoneIcons).map(name => pascalCaseToKebabCase(name));
```
juliodialpad, line 27: Is that empty option necessary? I removed it before because I found that empty option at the beginning annoying.

components/badge/badge.stories.js
```diff
@@ -15,13 +15,33 @@ export const argTypesData = {
       },
     },
   },
+  iconLeft: {
+    control: {
+      type: 'select',
+      options: getIconNames(),
```
juliodialpad: We might want to create a const with icon names here and reuse it, because we are needlessly calling the function twice.

components/emoji/index.js
```diff
@@ -1,2 +1 @@
 export { default as DtEmoji } from './emoji.vue';
-export { EMOJI_SIZES } from './emoji_constants';
```
juliodialpad, line 2: Thanks for removing this, didn't notice I missed that there.

juliodialpad (changes requested):
Nice work @braddialpad! Just a couple comments.

common/storybook_utils.js
braddialpad: the problem is you have no way to set it to "no icon" without it

braddialpad:
@francisrupert We're getting a contrast violation on `kind=bulletin`

> Element has insufficient color contrast of 4.42 (foreground color: #f9f9f9, background color: #7c52ff, font size: 9.0pt (12px), font weight: normal). Expected contrast ratio of 4.5:1

components/badge/badge.vue
modialpad: > @modialpad You're saying just have a leftIcon/rightIcon prop where you pass in the icon name rather than a slot? This is probably a good idea since this component is very fixed size. This would of course prevent any customization of the icon.. which might be what we want.

Funny enough I actually meant it the way Francis interpreted, basically hardcoding the icons inside if we know all the use-cases, but I'm glad it ended up for the best haha. I agree, using `DtIcon` is a better approach for this.

francisrupert:
> contrast violation...

Yeah, supposed to be white text. It uses `var(--fc-primary-inverted)` which used to be white and I forgot it updated to map to `--black-100` a good while back. 

I can piggypack that change onto [PR-762](https://github.com/dialpad/dialtone/pull/762) which I was about to merge.

francisrupert:
> In the [Docs view](https://vue.dialpad.design/deploy-previews/pr-637/?path=/docs/components-badge--default), can we replicate the examples from...

I see your update, but I probably wasn't specific enough; sorry. I was thinking an "Examples" section _beneath_ the Variants section you had going.

I definitely like the scannable list of all variants and would like to retain that and with the addition of the examples below.

image

components/badge/badge.stories.js
braddialpad: Good call thanks, done

common/storybook_utils.js
juliodialpad: Right, I know what you mean... Is it possible that you can add that option to only this story options instead of adding it to all of them?

I mean it's not a big problem, but I'll prefer to keep it the way it is now on the other components that uses this.

braddialpad:
> ...though it remains possible to create a `kind=count; type=ai`

We do output a console error on this. This is about the best we can do because when you're writing code you can pass in anything you type.

> Would it be possible to inhibit that? Too much to ask if Storybook has the possibility to disable the Label `<option>` and vice versa?

Storybook does have something called conditional controls where you can hide a control based on the value of some other control. However I tested this and if you set the kind to count beforehand and then change the type to ai, the previously selected kind=count remains so not sure if this really helps us at all.



> The icons are lit in here. Once I select one, I stumbled to figure out how to remove the icon before realizing it's the blank `<option>`. Would it be possible for that to be labeled? Something like "(none)". And if that needs to occur in other Vue components like [Button](https://user-images.githubusercontent.com/1165933/206349447-41e12079-ae4a-48e4-b705-e2cb00f32a22.png).

Actually yes, took me a while but I figured out how to do this using the `mapping` and `labels` options. Made the empty string now "(empty)"



> Upon selection of `kind="count"` would it be possible in Storybook to change the `text` value to a number and vice versa?

Unfortunately no, you can't dynamically change values. What I could do though is create a separate story for count, I think that might work

braddialpad:
> I see your update, but I probably wasn't specific enough; sorry. I was thinking an "Examples" section _beneath_ the Variants section you had going.
> 
> I definitely like the scannable list of all variants and would like to retain that and with the addition of the examples below.

Got it, should be fixed now.

francisrupert:
@braddialpad Cool cool cool to your https://github.com/dialpad/dialtone-vue/pull/637#issuecomment-1343417024. Trust your judgment and discretion. I don't think I have anything further, and I have no specific review for the Vue bits. 

I suppose we wait for `staging` to be released for the a111y color adjustment.

braddialpad:
> I suppose we wait for `staging` to be released for the a111y color adjustment.

Yep will do this now

common/storybook_utils.js
braddialpad: I managed to get this to show (empty) instead of a weird empty string and the (empty) will be selected by default Are there still cases in storybook where we do NOT want this functionality?

francisrupert (approved):
Great updates. Smooth slots and props interface.

juliodialpad (approved):
Really nice job here @braddialpad, thanks!

PR #639, opened 2022-12-09
fix: revert recent collapsible changes

modialpad:
# Revert Collapsible Changes

After some discussion w/ @dbecherdialpad and @braddialpad about this feature, we need to re-visit the design of allowing collapsible to show content on collapse.


braddialpad (approved):

braddialpad:
Hmm I'm not sure if the revert keyword will trigger a new library release.. 1 sec

braddialpad:
Yeah I think it will not, let's just change it to `fix:`

PR #642, opened 2022-12-13
fix(combobox-multi-select): firefox fix

iropolo:
# Firefox fix for combobox_multi_select component

- [x] Fix

## :book: Description

[DT-759](https://dialpad.atlassian.net/browse/DT-759)
We added parentClass prop to give access to handle clasess in parent element of `dt-chip` component

Now we can manipulate accordingly the `span` element.

This also fix the close icon and events (we can add and delete elements on combobox)

image


components/chip/chip.vue
```diff
@@ -145,6 +145,14 @@ export default {
...
       default: '',
     },
+
+    /**
+     * Additional class name for the span element.
+     */
+    parentClass: {
+      type: [String, Array, Object],
+      default: '',
+    },
```
braddialpad: You shouldn't need to do this, if you pass something in as just `class="whateverclass"` it will automatically be merged with the classes on the parent element.

https://v2.vuejs.org/v2/guide/class-and-style.html#With-Components

braddialpad (changes requested):

components/chip/chip.vue
iropolo: This fix its for the Vue3 version, I replicate here to be consistent.

Check where Im using this `parentClass` prop, its on the <span element above the <component, this <component are reciving `v-bind="$attrs"` because of this I cannot access to <span element as you mention.


image



https://vuejs.org/guide/essentials/class-and-style.html#binding-html-classes

braddialpad: Hmm you are definitely right that vue 3 binds class with `v-bind="$attrs"` while vue 2 does not.

I think we originally made a bit of a mistake in setting `v-bind="$attrs"` on the inner label here. It doesn't really make sense that it would set it on that element by default. I'd prefer if we had no `v-bind="$attrs"` at all set and then had a prop "labelClass" that would set classes on that element.

This is somewhat of a breaking change but there's not much usage of chip so I will check if it broke anything with the current product implementations after release

iropolo: I agree on this `I'd prefer if we had no v-bind="$attrs"` but idk how much you guys want to have similitude between v2 and v3, thats why I didn't clean or touch too much

`then had a prop "labelClass"` Okay, I'll update `parentClass` to `labelClass `

braddialpad: After we make this change it should be the same in vue 2 / vue 3 again. But in general if we have to change vue 3 branch because of a breaking change between vue 2 and vue 3, it is fine to make that change.

josedialpad:
The [Lint PR title action](https://github.com/dialpad/dialtone-vue/actions/runs/3750557422/jobs/6370424558) is failing because it's not following all the rules in the [commit convention](https://github.com/dialpad/dialtone-vue/blob/staging/.github/COMMIT_CONVENTION.md) due to capitalizing the subject part of the title and not using `kebab-case` for the scope. It should be: `fix(combobox-multi-select): firefox fix`.

Once you fix the PR title here, please apply the same changes to the [Vue3 PR](https://github.com/dialpad/dialtone-vue/pull/643).

components/chip/chip.vue
```diff
@@ -1,7 +1,6 @@
 <template>
-  <span class="d-chip">
+  <span :class="'d-chip'">
```
braddialpad: no reason to do this if it's just a string, leave it as an unbound attribute: `class="d-chip"`

components/chip/chip.vue
```diff
@@ -145,6 +142,14 @@ export default {
...
       default: '',
     },
+
+    /**
+     * Additional class name for the span element.
+     */
+    labelClass: {
+      type: [String, Array],
+      default: '',
+    },
```
braddialpad: add type `Object` as well, like you have for contentClass.

components/chip/chip.vue
```diff
@@ -203,10 +208,7 @@ export default {
 
   methods: {
     chipClasses () {
-      return [
-        this.$attrs['grouped-chip'] ? 'd-chip' : 'd-chip__label',
-        CHIP_SIZE_MODIFIERS[this.size],
-      ];
+      return [...this.labelClass, CHIP_SIZE_MODIFIERS[this.size]];
```
braddialpad: Something broke with the styling here:

Chip
Screenshot 2022-12-21 at 9 22 50 AM

Grouped Chip
Screenshot 2022-12-21 at 9 23 00 AM

braddialpad (changes requested):

braddialpad (approved):
Thanks!

PR #648, opened 2022-12-15
fix(root-layout): full height support

braddialpad:
# fix(root-layout): full height support

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Been getting some requests for this so decided to fix it right now since it was pretty low effort. 

Root layout component was not taking up the full viewport by default and consumers were having to pass in a bunch of utility classes to fix issues that this component should already handle. Added some flexbox classes to handle the sizing of the main content pane and render at full viewport height.

## :bulb: Context

This component should have taken up the full viewport from the start, since it's only meant to be used at the root. Decided to implement it directly instead of making it a prop.

## :pencil: Checklist

## :crystal_ball: Next Steps

release for implementation in web-clients

### DIFF
--- a/components/root_layout/root_layout.vue
+++ b/components/root_layout/root_layout.vue
@@ -1,5 +1,5 @@
 <template>
-  <div class="root-layout">
+  <div class="root-layout d-d-flex d-fd-column d-h100vh">
     <header
       v-if="$slots.header"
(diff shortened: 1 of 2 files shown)


braddialpad:
Adding Mika/David/Alex since they have all asked for this at some point

dbecherdialpad (approved):

AlexLamDialpad (approved):

PR #650, opened 2022-12-16
fix(chip): fix chip with avatar story

josedialpad:
# fix chip with avatar story

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

The chip with avatar story wasn't working since the content of the avatar should be passed using the default slot so changing to use the `html-fragment` component to render the HTML as slot instead of `v-html`.

## :pencil: Checklist

- [ ] I have added tests

### DIFF
--- a/components/combobox/combobox_default.story.vue
+++ b/components/combobox/combobox_default.story.vue
@@ -41,10 +41,7 @@
           @click="onListItemSelect(i)"
         >
           <template slot="left">
-            <dt-avatar
-              kind="initials"
-              color="purple-400"
-            >
+            <dt-avatar>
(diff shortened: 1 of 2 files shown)


components/combobox/combobox_default.story.vue
```diff
@@ -41,10 +41,7 @@
           @click="onListItemSelect(i)"
         >
           <template slot="left">
-            <dt-avatar
-              kind="initials"
-              color="purple-400"
```
josedialpad, line 46: Removed as these props are deprecated

braddialpad (approved):

PR #658, opened 2022-12-20
fix(combobox): combobox components issues

iropolo:
# Combobox components issues

- [x] Fix

## :book: Description

Bug fixs [DT-860](https://dialpad.atlassian.net/browse/DT-860)

image


components/chip/chip.vue
```diff
@@ -145,6 +145,14 @@ export default {
...
       default: '',
     },
+
+    /**
+     * Additional class name for the span element.
+     */
+    parentClass: {
+      type: [String, Array, Object],
+      default: '',
+    },
```
braddialpad: Didn't we decide to go with labelClass in this PR? https://github.com/dialpad/dialtone-vue/pull/642 should be the same here.

iropolo: It is not the same ticket.
Once the previous fix is approved, I will merge it here and it will be updated as we wish.

josedialpad:
The [Lint PR title action](https://github.com/dialpad/dialtone-vue/actions/runs/3742501735/jobs/6353521153) is failing because it's not following all the rules in the [commit convention](https://github.com/dialpad/dialtone-vue/blob/staging/.github/COMMIT_CONVENTION.md) due to capitalizing the subject part of the title. It should be: `fix(combobox): combobox components issues`.

Also since the scope is already defining the component affected by these changes, I'd suggest using something like: `fix(combobox): fix issues` and maybe detail the issues solved if it's not very extensive.

Once you fix the PR title here, please apply the same changes to the [Vue3 PR](https://github.com/dialpad/dialtone-vue/pull/657).

recipes/comboboxes/combobox_with_popover/combobox_with_popover.vue
```diff
@@ -94,6 +94,7 @@
           <div
             v-if="$slots.footer"
             ref="footer"
+            tabindex="0"
```
braddialpad: We shouldn't be focusing a non interactive element. [focusout](https://developer.mozilla.org/en-US/docs/Web/API/Element/focusout_event) should still trigger when the input is focused as it captures anything inside of it as well this is the difference between [focusout](https://developer.mozilla.org/en-US/docs/Web/API/Element/focusout_event) and [blur](https://developer.mozilla.org/en-US/docs/Web/CSS/filter-function/blur)

braddialpad (changes requested):

recipes/comboboxes/combobox_with_popover/combobox_with_popover.vue
braddialpad: I would suspect the problem is in this logic somewhere

```js
const isComboboxStillFocused = comboboxRefs.some((ref) => {
    return this.$refs[ref]?.contains(e.relatedTarget);
  }) || this.visuallyHiddenClose;
```

iropolo: Final result after discussion

```
const isComboboxStillFocused = e.relatedTarget === null || comboboxRefs.some((ref) => {
        return this.$refs[ref]?.contains(e.relatedTarget);
      }) || this.visuallyHiddenClose;
```

braddialpad (approved):

PR #661, opened 2022-12-20
feat(avatar): seeeded random colors

braddialpad:
# feat(avatar): seeeded random colors

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

My quick and fun project for the holidays 🎄. Added seeded randomization to the avatar. Just pass in a seed as a prop and it will always generate the same randomized colors for that seed.

## :bulb: Context

Allows using the same randomly generated colors for a specific user.

## :pencil: Checklist

## :link: Sources

https://www.npmjs.com/package/seedrandom


components/avatar/avatar.vue
```diff
@@ -206,22 +216,27 @@ export default {
...
 
     randomizeGradientColorStops () {
       const colors = new Set();
 
+      let count = 0;
       // get 3 unique colors, 2 from colorsWith100 and one from colorsWith200
       while (colors.size < MAX_GRADIENT_COLORS) {
+        // add count to the seed since we are looking for 3 unique colors. If the seed makes it always
+        // return the same color we'll get an infinite loop.
+        const seedForColor = this.seed === undefined ? undefined : this.seed + count.toString();
```
josedialpad, line 230: nit:

```suggestion
        const seedForColor = this.seed ? this.seed + count.toString() : undefined;
```

components/avatar/avatar_default.story.vue
```diff
@@ -4,6 +4,7 @@
     :size="size"
     :presence="presence"
     :initials="initials"
+    :seed="seed"
```
josedialpad, line 7: nit: add this prop to the icon story too, I think this can be useful to find a good combination of background/icon or for testing purposes.

josedialpad (approved):
Great addition!

josedialpad:
BTW not sure why the preview for this PR isn't working

components/avatar/avatar.vue
braddialpad: not quite the same because in this case empty string would become undefined

braddialpad:
> BTW not sure why the preview for this PR isn't working

Yeah super weird.. it is now that it rebuilt.

components/avatar/avatar_default.story.vue
braddialpad: good call, added

PR #665, opened 2022-12-22
feat(popover): add appendTo prop.

juliodialpad:
# Add appendTo prop to popover.

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Added appendTo prop to popover to enable users to append the popover to any HTML element or 'parent'.

## :bulb: Context

We have a dev with a case where they have a nested shadow dom and they need it to be within that body rather than the root body.

## :pencil: Checklist

- [ ] I have added tests


components/popover/popover.vue
```diff
@@ -451,6 +451,19 @@ export default {
...
+    /**
+     * Sets the element to which the popover is going to append to.
+     * @values 'parent', document.body
+     */
+    appendTo: {
+      type: [HTMLElement, String],
+      default: () => document.body,
+      validator: v => {
+        if (v instanceof HTMLElement) return true;
+        return !(typeof v === 'string' && v !== 'parent');
```
josedialpad: The same conditional but avoiding negation is simpler

```suggestion
        return typeof v === 'string' && v === 'parent';
```

josedialpad (approved):
w/ one comment

juliodialpad:
@josedialpad Forgot to push some changes yesterday, just noticed until finished vue3 version, I had refactored the validation to extract constants and improve readability. Please give another look.

josedialpad (approved):

PR #667, opened 2023-01-05
fix(dt-toggle): removed internal state

vishaldialpad:
# PR Title
Jira: https://dialpad.atlassian.net/browse/DT-870
Github issue: https://github.com/dialpad/dialtone-vue/issues/668

- [X] Fix
- [X] Refactoring

## :pencil: Checklist

- [X] I have reviewed my changes
- [X] Unit tests are passing locally
- [X] Linters are passing locally
- [X] Library builds locally
- [X] Documentation builds locally

## :link: Sources

fix for for dt-toggle where `internalChecked:true` while `checked:false`
https://drive.google.com/file/d/1VNvl4ckIF0bkxdqJHKz3ZMAWtlrgNtlV/view?usp=sharing

### DIFF
--- a/components/toggle/toggle.vue
+++ b/components/toggle/toggle.vue
@@ -14,7 +14,7 @@
       :id="id"
       :role="toggleRole"
       type="button"
-      :aria-checked="internalChecked.toString()"
+      :aria-checked="checked.toString()"
       :disabled="disabled"
       :aria-disabled="disabled.toString()"
       :class="toggleClasses"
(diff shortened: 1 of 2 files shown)


braddialpad (changes requested):
Couple of issues here:

1. Setting the checked prop in storybook no longer changes the value of the toggle (I think you need a watch on checked)
2. Getting an error on load

```
[Vue warn]: Invalid prop: custom validator check failed for prop "checked".
```
3. This is a breaking change as clicking the toggle when setting the checked prop (rather than using v-model) will no longer change the state of the toggle. Making this change would require a migration if any of our products are using the checked prop in this way. Also we want it to work this way because this is how a native checkbox works and we want it to be a close to native as possible. I think what we perhaps need to do is ignore the internal state ONLY when using v-model.

PR #678, opened 2023-01-06
docs(dropdown): keyboard nav fix

iropolo:
# Dropdown Keyboard nav fix

- [x] Fix

## :book: Description
[DT-794](https://dialpad.atlassian.net/browse/DT-794)

keyValidation blocks proper behaviour of keyboard navigation.

image

[DT-794]: https://dialpad.atlassian.net/browse/DT-794?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


components/dropdown/dropdown.vue
```diff
@@ -432,7 +432,7 @@ export default {
     },
 
     onKeyValidation (e, eventHandler) {
-      if (this.open !== null && !this.openOnContext) { return; }
+      if (this.open !== null && !this.openOnContext && !this.isOpen) { return; }
```
braddialpad: I spoke to Jose about his original intention with this code since it was confusing me. It's basically just to prevent the menu from opening with the arrow keys when it's a context menu. We found a way to greatly simplify this logic since it's very confusing with all these opens and isOpens.

We can just use the `open-with-arrow-keys` prop on popover. Something like this:

```
<dt-popover
 :open-with-arrow-keys="shouldOpenWithArrowKeys"
 ...
...

<script>
...
    shouldOpenWithArrowKeys () {
      return !this.openOnContext;
    },
</script>
```

Then we can probably just get rid of onKeyValidation entirely. Don't think we'll need it anymore.

braddialpad (changes requested):
For when you get back from vacation, just let me know if you need any further info on this.

components/dropdown/dropdown.vue
```diff
@@ -431,10 +431,8 @@ export default {
       return this.onNavigationKey(e.key);
     },
 
-    onKeyValidation (e, eventHandler) {
-      if (this.open !== null && !this.openOnContext) { return; }
-
-      this[eventHandler](e);
+    shouldOpenWithArrowKeys () {
+      return !this.openOnContext;
```
braddialpad: hmm think you just need to make this a computed rather than a method.

braddialpad (approved):
Seems to be working now, nice!

braddialpad:
Remember to transfer the recent changes to vue 3 PR as well

PR #683, opened 2023-01-06
fix(popover): blocking wheel event

juliodialpad:
# Fix blocking wheel event on popover

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Moved @wheel event to popover's *transparent overlay* and *lazy-show content* to avoid filling the window with events when there's too many popovers.

This also fixes the wheel event behavior, when modal, it shouldn't let the user to scroll, with this fix it actually prevents the user from scrolling when popover open.

## :bulb: Context

Found a warning while working with a lot of popovers on a single page (422), the browser throws a warning and the page remains blocked until all the popovers get rendered.

Example error, limited to 4 popovers for readability.
Screenshot 2023-01-06 at 5 31 08 p m

## :pencil: Checklist

- [ ] I have added tests

### DIFF
--- a/components/popover/popover.vue
+++ b/components/popover/popover.vue
@@ -5,6 +5,7 @@
         class="d-modal--transparent"
         :aria-hidden="modal && isOpen ? 'false' : 'true'"
         @click.prevent.stop
+        @wheel="(e) => (isOpen && modal) && e.preventDefault()"
       />
     </portal>
     <component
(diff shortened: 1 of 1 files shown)


components/popover/popover.vue
```diff
@@ -823,6 +824,11 @@ export default {
         this.$refs.content.$el.focus();
       }
     },
+
+    handleWheel (e) {
+      e.preventDefault();
+      console.log(e);
+    },
```
braddialpad: Do you need this?

juliodialpad: No, forgot to remove it, was using it for testing.

braddialpad (approved):

PR #684, opened 2023-01-06
fix(popover): blocking wheel event - vue3

juliodialpad:
# Vue 3 version of #683


juliodialpad:
Looks like actions are bugged here, will close this and try to recreate it again.

PR #688, opened 2023-01-09
fix(popover): refactor wheel event

juliodialpad:
# Refactored wheel event

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Removed wheel event from the popover instances, added a `window.addEventListener` to prevent scrolling only when the popover is modal and open. This should be the right behavior and fix the issues with scrolling.

## :bulb: Context

After merging #683 Found that the event was still having issues, so found a different and better way to solve this scrolling issues with popover. The listener should only be added when the popover is opened.

## :pencil: Checklist

- [ ] I have added tests

## :crystal_ball: Next Steps

Remove the ability to scroll through the page with arrow-keys, the scrolling is blocked now, but arrow keys still can be used to move the content of the page.

### DIFF
--- a/components/popover/popover.vue
+++ b/components/popover/popover.vue
@@ -5,7 +5,6 @@
         class="d-modal--transparent"
         :aria-hidden="modal && isOpen ? 'false' : 'true'"
         @click.prevent.stop
-        @wheel.prevent.stop
       />
     </portal>
     <component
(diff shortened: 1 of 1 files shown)


braddialpad (approved):
Nice, this is a good solution.

PR #692, opened 2023-01-11
refactor(popover): tippy instance

juliodialpad:
# Refactor popover's tippy instance

- [ ] Fix
- [ ] Feature
- [x] Refactoring
- [ ] Documentation

## :book: Description

Moved the tippy instance on popover to be initialized when open and not on mounted to improve performance.

## :bulb: Context

We had a lot of performance issues while implementing a page with 400+ popovers, found out that was due to the fact that popover's tippy instance where initialized on mounted, so we decided to move the initialization on open to prevent a lot of tippy instances unused.

## :pencil: Checklist

- [ ] I have added tests

## :crystal_ball: Next Steps

Refactor tooltip and any other component that uses tippy on its core to do this too.

### DIFF
--- a/recipes/comboboxes/combobox_with_popover/combobox_with_popover.vue
+++ b/recipes/comboboxes/combobox_with_popover/combobox_with_popover.vue
@@ -23,7 +23,6 @@
         :id="externalAnchor"
         ref="input"
         @focusin="onFocusIn"
-        @focusout="onFocusOut"
         @keydown.up="openOnArrowKeyPress($event)"
         @keydown.down="openOnArrowKeyPress($event)"
       >
@@ -60,7 +59,6 @@
           <div
             v-if="$slots.header"
             ref="header"
-            @focusout="onFocusOut"
           >
             <slot name="header" />
           </div>
@@ -71,7 +69,7 @@
             ref="listWrapper"
             :class="[DROPDOWN_PADDING_CLASSES[padding], listClass]"
             @mouseleave="clearHighlightIndex"
-            @focusout="clearHighlightIndex; onFocusOut;"
+            @focusout="clearHighlightIndex"
           >
             <combobox-loading-list
               v-if="loading"
@@ -94,7 +92,6 @@
           <div
             v-if="$slots.footer"
             ref="footer"
-            @focusout="onFocusOut"
           >
             <slot name="footer" />
           </div>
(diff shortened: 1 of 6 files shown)


recipes/comboboxes/combobox_with_popover/combobox_with_popover.vue
```diff
@@ -430,14 +432,12 @@ export default {
...
       // Check if the focus change was to another target within the combobox component
-      const isComboboxStillFocused = e.relatedTarget === null || comboboxRefs.some((ref) => {
-        return this.$refs[ref]?.contains(e.relatedTarget);
-      }) || this.visuallyHiddenClose;
+      const isComboboxStillFocused = this.$refs?.combobox?.$el.contains(e.target);
+      if (isComboboxStillFocused) return;
 
       // If outside the combobox then close
-      if (!isComboboxStillFocused) { this.closeComboboxList(); }
+      this.closeComboboxList();
```
braddialpad: This is certainly cleaner but unfortunately is not going to work. If you click anywhere within the popover with the mouse it will immediately close. This is the problem the old previous code was solving. If the click was within the open popover, or goes to "unfocused" (`e.relatedTarget === null`) by clicking a non interactive element, the popover will remain open.

braddialpad (changes requested):

recipes/comboboxes/combobox_with_popover/combobox_with_popover.vue
juliodialpad: I'm sure I tested that a million times and it was working correctly, let me see if I forgot to push something.

juliodialpad: Don't know what happened, added back the right code, should be fixed.

braddialpad: yep looks good now thanks!

braddialpad (approved):

braddialpad:
The accessibility tests are broken on staging, feel free to merge if you wish.

juliodialpad:
> The accessibility tests are broken on staging, feel free to merge if you wish.

Yeah, looks like updating dialtone version broke Accessibility tests. I'll merge it then, thanks!

PR #696, opened 2023-01-12
docs: update contributing.md with a note about breaking changes

modialpad:
# Update CONTRIBUTING.md with a note about breaking changes

Add a small note about breaking changes in the "Submitting" section that emphasizes the importance of communicating breaking changes in advance before submitting a PR.

- [x] Documentation

## :bulb: Context

Previous work on Collapsible had to be reverted because it broke some functionality that was thought to be unnecessary. Ideally, updates to that component should not have removed functionality that was thought to be useful for other teams using that component. 

giphy 2


.github/CONTRIBUTING.md
```diff
@@ -38,6 +38,8 @@ The second is to check with the Dialtone team (in [#dialtone](https://dialpad.sl
 
 Before submitting a pull request, make sure to communicate what you wish to change to the Dialtone team. The easiest way to do this is via the [#dialtone](https://dialpad.slack.com/messages/dialtone/) Slack channel. It's possible your change is already being worked on, has already been fixed, or maybe we just need to discuss the best solution to the problem. This prevents you from having to re-write your entire change, or even having to scrap it entirely.
 
+The Dialtone library exists to serve multiple teams, so it is important to ensure that your PR does not introduce any unnecessesary breaking changes. Breaking changes will require extra steps that may prolong the submission review process. If breaking changes are absolutely necessary for your use case, then you must communicate them to the Dialtone team in advance to allow for a smooth migration strategy for all dependant projects.
```
braddialpad: "unnecessary"

.github/CONTRIBUTING.md
```diff
@@ -38,6 +38,8 @@ The second is to check with the Dialtone team (in [#dialtone](https://dialpad.sl
...
+
```
braddialpad, line 53: Maybe include an example of something that's a breaking change vs a non-breaking change. Developers that are new to library development may have trouble understanding what a breaking change is.

braddialpad (commented):
Thanks, looking good!

.github/CONTRIBUTING.md
modialpad: Agreed, I'm planning to add the following paragraph:

```md
Examples of breaking changes:

- Changing the component API: renaming or removing props and emits, or changing their expected types
- Removing or altering features: changing a component's existing expected behaviour or output. For example, changing or removing a component's animation style, or updating a component's sorting algorithm for its child items, etc.

Examples of non-breaking changes:

- Augmenting an existing component's feature set: for example, adding a new prop to a component that triggers a different flow of behaviour or output, or adding a new emit event to detect specific changes. 
- Refactoring: updating a component to use a different util or library (without causing any outward changes). 
- Accessibility: updating a component to be more in line with our accessibility guidelines, without changing any external behaviour

Some changes can unintentionally cause breaking changes, like when integrating a new NPM library or using a new util. This is why it is important to evaluate the change carefully and asses its impact on existing usage.
```

Do you think we should add more examples?

.github/CONTRIBUTING.md
modialpad: Updated

.github/CONTRIBUTING.md
braddialpad: I think this covers it fairly well. Couple things I have to comment on.

> Changing the component API

maybe "Changing or removing the component API" 

> Augmenting an existing component's feature set

Maybe instead of "Augmenting an existing component's feature set" we could say "Making new additions to the component API". Trying to make it very clear that change / remove = bad, new = good

> Removing or altering features: changing a component's existing expected behaviour or output. For example, changing or removing a component's animation style, or updating a component's sorting algorithm for its child items, etc.

I wouldn't really say this is a breaking change. If you change a components output but don't change anything to do with the public interface it's not going to break. You do however have to think about how your change will affect any system currently using the component. Example you move the text to align right instead of left on all dropdown items. It doesn't change the public interface, but it will surely affect how everything looks. Is it okay that it does this on every single component currently in use?

> Refactoring: updating a component to use a different util or library (without causing any outward changes).

Refactoring: changing the internal code of the component without changing any external behaviour.

> Accessibility: updating a component to be more in line with our accessibility guidelines, without changing any external behaviour

I would argue that most accessibility changes are changing external behaviour. It's possible to change external behaviour without making a breaking change though.

modialpad: Got it, I guess I thought of 'breaking' in a broader sense, where the functionality also changes but without causing actual runtime failures. I'll update the doc with what you mentioned.

braddialpad (approved):
Thanks!

PR #702, opened 2023-01-16
feat: enable shadowMode

braddialpad:
# feat: enable shadowMode 

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Related to https://dialpad.atlassian.net/browse/DP-63528

Added support for `shadowMode` which allows css classes in the library to be rendered within a shadow dom rather than always at the root.

The way to enable this is a bit cumbersome as vue.config.js doesn't really have a built in option to change this directly for all instance, so you have to go through the vue-cli generated config and manually set it on each rule. After searching a bunch this appears to be the best way available.

The consumer must set the `shadowRoot: element` as an option on the vue instance where element is the element where the styles should be rendered.

Joao has tested this change on the kare agent assist widget and has confirmed it fixes the issue.

## :bulb: Context

The Kare agent assist widget renders in a shadow dom within firespotter, however any styles from dialtone vue were being rendered at the root rather than within the shadow dom. This was overriding dialtone classes already in the project with older versions. The css needs to be isolated within the shadow dom.

## :pencil: Checklist

## :crystal_ball: Next Steps

Joao to update to the new release in agent assist widget

## :link: Sources

https://vue-loader.vuejs.org/options.html#shadowmode
https://developer.mozilla.org/en-US/docs/Web/Web_Components/Using_shadow_DOM

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,10 @@ import { getEmojiData } from '@dialpad/dialtone-vue/emoji'
 const emojiData = getEmojiData();
 ```
 
+## Shadow DOM
+
+If you are using Dialtone Vue within a shadow DOM set the `shadowRoot: YOUR_ELEMENT` option on your vue instance and the styles will be isolated to that element rather than rendered at the root.
+
 ## Contributing
 
 If you would like to contribute to Dialtone Vue the first step is to get the project running locally. Follow the below quickstart to do so.
--- a/vue.config.js
+++ b/vue.config.js
@@ -30,7 +30,8 @@ function enableShadowCss (config) {
     config.module.rule('stylus').oneOf('normal').use('vue-style-loader'),
   ];
   configs.forEach(c => c.tap(options => {
-    options.shadowMode = true;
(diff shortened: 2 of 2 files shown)


josedialpad (approved):
As an alternative, using the built-in `scoped` style should work?

josedialpad:
Seems that it's breaking the unit tests

braddialpad:
> Seems that it's breaking the unit tests

yep.. doesn't seem to like localvue for some reason (tests work fine in vue 3)

> As an alternative, using the built-in `scoped` style should work?

Maybe.. I'm not sure it would work with stuff like this:

```
.tippy-box[data-popper-reference-hidden],
.tippy-box[data-popper-escaped] {
  .d-popover__dialog {
    visibility: hidden;
    pointer-events: none;
  }
}
```

braddialpad:
tests fixed, just disabled the option when running tests.

PR #708, opened 2023-01-19
fix(popover): blocking wheel event

juliodialpad:
# Fix popover blocking wheel event

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Fixed the popover blocking wheel event by bypassing events comming from inside the popover.

This gives another issue that the popover might be scrolled out of window by scrolling inside of it with over a no scrollable element.

## :bulb: Context

https://dialpad.atlassian.net/browse/DT-910

## :pencil: Checklist

- [ ] I have added tests

## :crystal_ball: Next Steps

Capture the events that its target is not a scrollable element within the popover.

### DIFF
--- a/components/popover/popover.vue
+++ b/components/popover/popover.vue
@@ -698,6 +698,11 @@ export default {
     **/
     preventScrolling (e) {
       if (!this.modal) return;
+      const isEventComingWithinPopover = Object.keys(this.$refs).some(refName => {
+        const ref = this.$refs[refName];
+        return ref.$el ? ref.$el.contains(e.target) : ref.contains(e.target);
+      });
+      if (isEventComingWithinPopover) return;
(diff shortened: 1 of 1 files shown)


components/popover/popover.vue
```diff
@@ -698,6 +698,11 @@ export default {
     **/
     preventScrolling (e) {
       if (!this.modal) return;
+      const isEventComingWithinPopover = Object.keys(this.$refs).some(refName => {
```
braddialpad: Do we need to check all refs here? can't we just check within ref="content"?

juliodialpad: I think it might work with content only, let me test it.

braddialpad (approved):

PR #710, opened 2023-01-20
fix(root-layout): detect empty slot

braddialpad:
# fix(root-layout): detect empty slot

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Using `$slots.footer` to detect a slot is empty no longer works correctly in vue 3. In fact it is very cumbersome to simply detect whether a slot is empty. Copied the method posted by andrey-hohlov in [this thread](https://github.com/vuejs/core/issues/4733). There is an RFC to fix this in vue, but has not been worked on yet.

## :bulb: Context

Root layout was still taking up space for the footer even when it was empty, the v-if on the slot wrapper was not working correctly in Vue 3

## :pencil: Checklist

## :crystal_ball: Next Steps

Inform DP 2.0 of this change.

### DIFF
--- a/common/utils.js
+++ b/common/utils.js
@@ -3,7 +3,11 @@ import {
   DEFAULT_VALIDATION_MESSAGE_TYPE,
   VALIDATION_MESSAGE_TYPES,
 } from './constants';
-import { h } from 'vue';
+import {
+  h,
+  Comment,
+  Text,
+} from 'vue';
 const seedrandom = require('seedrandom');
 
 let UNIQUE_ID_COUNTER = 0;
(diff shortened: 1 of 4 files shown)


josedialpad (approved):
> In fact it is very cumbersome to simply detect whether a slot is empty. 

Yeah it's a pain that.

We already had a code for trying to validate the empty slot in the tooltip. I think we can replace here too with the `hasSlotContent` from the utils.

https://github.com/dialpad/dialtone-vue/blob/vue3/components/tooltip/tooltip.vue#L255-L267

braddialpad:
> 

Ahh yeah good find will replace

PR #711, opened 2023-01-20
feat: css as separate file

braddialpad:
# feat: css as separate file

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Instead of using js to inject css, store it in a separate file and allow the consumer to import it manually.

## :bulb: Context

Fixes the issue from the following slack thread: https://dialpad.slack.com/archives/C03EAH70A5P/p1674006782305509

Css classes from Dialtone Vue in Dialpad 2.0 were being completely ignored.

## :pencil: Checklist

## :crystal_ball: Next Steps

- Communicate this to teams using Dialtone Vue 3 since it is a breaking change.
- Move the new addition to the readme file over to the main staging branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,6 +26,12 @@ Dialtone Vue components can be imported directly from the package. Some componen
 import { DtInput, VALIDATION_MESSAGE_TYPES } from '@dialpad/dialtone-vue';
 ```
 
+If you are using the Vue 3 version of Dialtone, you must also import the css:
+
+```js
+import '@dialpad/dialtone-vue/css';
+```
+
 Projects using Dialtone Vue should be aware of the requirements:
 
 - Dialtone classes must be made available globally (to avoid duplication, Dialtone Vue does not do this for you).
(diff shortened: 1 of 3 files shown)


vue.config.js
```diff
@@ -1,46 +1,8 @@
...
-  ];
-  configs.forEach(c => c.tap(options => {
-    options.shadowMode = true;
-    return options;
-  }));
-}
-
 module.exports = defineConfig({
   lintOnSave: false,
-  css: { extract: false },
```
braddialpad, line 40: note that the default for css extract is false on development, true in production, which is what we want in this case.

vue.config.js
```diff
@@ -1,46 +1,8 @@
...
-    config.module.rule('stylus').oneOf('vue-modules').use('vue-style-loader'),
-    config.module.rule('stylus').oneOf('vue').use('vue-style-loader'),
-    config.module.rule('stylus').oneOf('normal-modules').use('vue-style-loader'),
-    config.module.rule('stylus').oneOf('normal').use('vue-style-loader'),
-  ];
-  configs.forEach(c => c.tap(options => {
-    options.shadowMode = true;
-    return options;
-  }));
-}
```
braddialpad, line 36: Not needed anymore since the consumer can control exactly where the styling goes. (will tell joao about this so he can update)

README.md
```diff
@@ -26,6 +26,12 @@ Dialtone Vue components can be imported directly from the package. Some componen
 import { DtInput, VALIDATION_MESSAGE_TYPES } from '@dialpad/dialtone-vue';
 ```
 
+If you are using the Vue 3 version of Dialtone, you must also import the css:
```
josedialpad, line 29: Remove the `## Shadow DOM` section in the readme as it's not relevant anymore?

josedialpad (approved):

README.md
braddialpad: yep, thanks

PR #722, opened 2023-01-26
fix(popover): fix prevent scroll logic

braddialpad:
# fix(popover): fix prevent scroll logic

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

https://dialpad.atlassian.net/browse/DT-929

Updated the prevent scroll logic to do a couple things differently:
- If modal popover is opened on it's own, set the body to `d-of-hidden` to prevent scrolling.
- If modal popover is opened within another popover, set the z-index of the popover it is within to `d-zi-popover` rather than `d-zi-modal-element` which means it is the currently active modal window. This puts it behind the transparent modal overlay and prevents scrolling / clicking etc.

As we know setting `d-of-hidden` on the body removes the scrollbar if there is one making the content "jump". I don't think this should be a problem in our products as the body doesn't have a scrollbar in Dialpad/DPM. It may affect our documentation site and the Dialpad marketing site, but I don't think there are any dt popovers used there.

There is a potential solution to scrollbar issue shown here https://css-tricks.com/prevent-page-scrolling-when-a-modal-is-open/ however I've not had much luck with it so far. Since this is breaking product we need to get this fix in by Friday so maybe it is best to release this and continue to iterate.

## :bulb: Context

Popover prevent scroll logic was broken when a popover was within a popover. It could not be scrolled at all. We needed to find a better way of doing this rather than intercepting the mouse wheel event.

## :pencil: Checklist

## :crystal_ball: Next Steps

- test in product
- find an override fix on the docsite for the scrollbar issue.

## :link: Sources

https://css-tricks.com/prevent-page-scrolling-when-a-modal-is-open/
https://stackoverflow.com/questions/9280258/prevent-body-scrolling-but-allow-overlay-scrolling


braddialpad:
Updated the "Popover with Dropdown" variant example to have scrollbars so you can test this.

josedialpad:
Testing in the section where the bug happened found two issues that don't seem to be in beta:
1. Double click is needed in the status button to show the second popover.
2. The popover isn't removed from the DOM after closing it.

https://user-images.githubusercontent.com/83774467/214964494-9ed9b016-c538-48ab-97ac-6a33ffc33e66.mp4

I can't understand how / if your changes are causing this since I can't reproduce them on beta

braddialpad:
hmm there was an error happening in the component due to modal.js bug from earlier. I just merged in staging try now...

josedialpad:
> hmm there was an error happening in the component due to modal.js bug from earlier. I just merged in staging try now...

that was! works great now

josedialpad (approved):

braddialpad:
Nice, thanks for checking

PR #724, opened 2023-01-27
fix(tooltip): tooltip with circle button corner fix

iropolo:
# fix(tooltip): tooltip with circle button corner fix

- [x] Fix

## :book: Description

Replace wrapper element from `<div>` to `<a>`, with this, we solve the corner issue on circle buttons.
I also removed unused code.

## :bulb: Context

Hover trigger Tooltip on corners of a circle button (because wrapper its a rectangle)
As it was mentioned on [DP-58183](https://dialpad.atlassian.net/browse/DP-58183)

## :pencil: Checklist

[DP-58183]: https://dialpad.atlassian.net/browse/DP-58183?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/components/tooltip/tooltip.vue
+++ b/components/tooltip/tooltip.vue
@@ -1,7 +1,9 @@
 <template>
   <div data-qa="dt-tooltip-container">
-    <div
+    <a
       ref="anchor"
+      tabindex="0"
+      role="button"
       data-qa="dt-tooltip-anchor"
       @focusin="onEnterAnchor"
       @focusout="onLeaveAnchor"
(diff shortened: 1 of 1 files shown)


braddialpad (changes requested):
This is a rather clever solution to this problem however you're going to need to remove the tabindex and set role="presentation". Remember that this is the wrapper for the anchor, not the anchor itself so it should not be interactive. The actual anchor element is passed in through the slot.

braddialpad (approved):

PR #726, opened 2023-01-30
fix(tab): fix tabindex value tied to the first element in the panel

josedialpad:
# fix tabindex value tied to the first element in the panel

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Updated the `tabindex` logic regarding the first element in the panel described in the notes (4) in https://www.w3.org/WAI/ARIA/apg/patterns/tabpanel/.
This PR adds the validation when the first element with content is not focusable, the tabpanel should set `tabindex="0"`.

## :pencil: Checklist

- [x] I have added tests

2023-01-30 at 12 13 00@2x

2023-01-30 at 12 16 15@2x

## :link: Sources

- https://www.w3.org/WAI/ARIA/apg/patterns/tabpanel/
- https://www.w3.org/WAI/ARIA/apg/example-index/tabs/tabs-manual.html

### DIFF
--- a/components/tabs/tab_panel.vue
+++ b/components/tabs/tab_panel.vue
@@ -85,7 +85,9 @@ export default {
     if (!firstFocusableElement) {
       this.isFirstElementFocusable = false;
     } else {
-      // If the first focusable element isn't the first element, then we need to set the panel tabindex to 0
+      // If the first focusable element isn't the first element in the panel,
+      // then we need to set the panel tabindex to 0.
+      // See notes in https://www.w3.org/WAI/ARIA/apg/patterns/tabpanel/
       this.isFirstElementFocusable = this.isFirstElementOfPanel(firstFocusableElement);
     }
   },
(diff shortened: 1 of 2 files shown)


components/tabs/tab_panel.vue
```diff
@@ -80,7 +80,33 @@ export default {
...
+      let current = element;
+      let isFirstElement = true;
+
+      while (current) {
+        if (current.previousElementSibling !== null) {
+          isFirstElement = false;
+          break;
+        }
+        current = current.parentNode !== this.$el ? current.parentNode : null;
+      }
```
braddialpad, line 106: Do we even need to loop this? If there's any previous sibling before the firstFocusableElement then we know right away it's not the first element. I think just this would work right?

```js
isFirstElementOfPanel (element) {
  return element.previousElementSibling !== null
}

**Review by braddialpad — CHANGES_REQUESTED** [review]

File: components/tabs/tab_panel.vue (hunk 1 above)
Comment (josedialpad) [reply components/tabs/tab_panel.vue:106]: Yeah that should work for the siblings of the element's parent but with the loop we also check for other elements that aren't in that node. For example:

```html
<dt-panel>
  <h2>Content</h2>
  <div id="container>
    <button>button</button>
  </div>
<dt-panel>
```

`element.previousElementSibling` will only check inside the `div#container` but the first element in the panel is the `h2`.

braddialpad: Hmm yep I see. This is good then.

braddialpad (approved):

PR #738, opened 2023-02-02
fix(root-layout): add fixed prop

braddialpad:
# fix(root-layout): add fixed prop

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Root Layout was not making the main content scrollable, instead the entire thing would scrolling putting the header/footer out of view. In most cases this is not how it will be used (but it will be in some cases, such as web settings). Added a new prop called "fixed" that locks the header and footer in place and makes the content scrollable.

## :bulb: Context

Issue in DP 2.0 that Alex was having.

## :pencil: Checklist

## :crystal_ball: Next Steps

Update in DP 2.0 and make sure it works as expected.


components/root_layout/root_layout_body.vue
```diff
@@ -15,7 +15,7 @@
     <main
       v-if="$slots.content"
       :class="['root-layout__content', contentClass]"
-      :style="{ 'min-inline-size': contentWrapWidthPercent }"
+      :style="{ 'min-inline-size': contentWrapWidthPercent, 'height': fixed ? mainHeight : null }"
```
iropolo: For future implementations in planning, would like a refactor on the components to avoid empty style/class attributes (This doesn't affect the proper working but it is a more presentable component).
Also to avoid large style attributes or logic in the template.

Template should be cleaner as possible.

components/root_layout/root_layout_default.story.vue
```diff
@@ -25,7 +26,236 @@
       <span v-html="sidebar" />
     </template>
     <template v-if="defaultSlot">
-      <span v-html="defaultSlot" />
+      Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aliquam dignissim eleifend condimentum. Vestibulum
```
josedialpad: I think this can be removed from the template by repeating a sample text passed from the story definition like the pattern [we've followed in the modal](https://github.com/dialpad/dialtone-vue/blob/staging/components/modal/modal.stories.js#L148).

Also make the sample text fallbacks the defaultSlot.
```html
<template>
    <span v-if="defaultSlot" v-html="defaultSlot" />
    <span v-else>{{ sampleText}}</span>
</template>

```

components/root_layout/root_layout_body.vue
```diff
@@ -91,6 +106,10 @@ export default {
         { 'root-layout__body--invert': this.sidebarPosition === ROOT_LAYOUT_SIDEBAR_POSITIONS.RIGHT },
       ];
     },
+
+    mainHeight () {
+      return `calc(100vh - (${this.headerHeight} + ${this.footerHeight}))`;
```
josedialpad: We need to consider the case in a small viewport (`< 512px`) where the sidebar is on top of the content reducing the viewport size so the sidebar height should be added to this calc.

2023-02-02 at 12 10 38@2x

josedialpad (changes requested):
Looking good, left 2 suggestions

components/root_layout/root_layout_body.vue
josedialpad: The `100vh` is very tricky on mobile devices. Safari or Android browsers handle it very inconsistently (e.g. not reducing the viewport when [adding some browser bars](https://chanind.github.io/javascript/2019/09/28/avoid-100vh-on-mobile-web.html)).

A more reliable calculation instead of `100vh` may be to use `window.innerHeight` as suggested [here](https://dev.to/nirazanbasnet/dont-use-100vh-for-mobile-responsive-3o97).

components/root_layout/root_layout_body.vue
braddialpad: Yep good call, really no need for this logic to be in the template, it can just go in the mainHeight computed property.

components/root_layout/root_layout_default.story.vue
braddialpad: well that's much cleaner, thanks.

components/root_layout/root_layout_body.vue
braddialpad: Was a bit difficult to determine whether the sidebar was above or below the footer due to it being pure css, however I managed to do it with a resize observer and offsetTop.

braddialpad:
Thanks for the comments guys, ready for re-review

components/root_layout/root_layout_body.vue
braddialpad: updated to use innerHeight in a resizeObserver

josedialpad (approved):
Looks great,thanks!

PR #739, opened 2023-02-02
fix(collapsible): correct anchor slot check

braddialpad:
# fix(collapsible): correct anchor slot check

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Code was using `this.$slots.$anchor` which is incorrect it should be `this.$slots.anchor`. Also updated to use our new `hasSlotContent` function to get around the slot content issues in vue 3.

This is a vue 3 only issue.

## :link: Sources

https://dialpad.atlassian.net/browse/DP-64811

### DIFF
--- a/components/collapsible/collapsible.vue
+++ b/components/collapsible/collapsible.vue
@@ -79,7 +79,7 @@
 </template>
 
 <script>
-import { getUniqueString } from '@/common/utils';
+import { getUniqueString, hasSlotContent } from '@/common/utils';
 import DtCollapsibleLazyShow from './collapsible_lazy_show';
 import { DtButton } from '../button';
 import { DtLazyShow } from '../lazy_show';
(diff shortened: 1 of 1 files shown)


josedialpad (approved):
Shouldn't all instances of `this.$slots` used in validations be wrapped by the `hasSlotContent` function to detect the content correctly?
I found this similar case in other validations, for example: 
https://github.com/dialpad/dialtone-vue/blob/vue3/components/toggle/toggle.vue#L193 
https://github.com/dialpad/dialtone-vue/blob/vue3/components/combobox/combobox.vue#L400
https://github.com/dialpad/dialtone-vue/blob/vue3/components/modal/modal.vue#L370

braddialpad:
> Shouldn't all instances of `this.$slots` used in validations be wrapped by the `hasSlotContent` function to detect the content correctly? I found this similar case in other validations, for example: https://github.com/dialpad/dialtone-vue/blob/vue3/components/toggle/toggle.vue#L193 https://github.com/dialpad/dialtone-vue/blob/vue3/components/combobox/combobox.vue#L400 https://github.com/dialpad/dialtone-vue/blob/vue3/components/modal/modal.vue#L370

Hmm probably. I will make this change in a separate PR.

PR #741, opened 2023-02-02
chore(deps): update deps

josedialpad:
# update deps

- [ ] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :pencil: Checklist

- [ ] I have added tests


storybook/package.json
```diff
@@ -8,23 +8,23 @@
...
-    "@storybook/vue3": "^6.5.14",
-    "@vue/compiler-sfc": "^3.2.45",
+    "@storybook/addon-viewport": "^6.5.16",
+    "@storybook/addons": "^6.5.16",
+    "@storybook/builder-webpack5": "^6.5.16",
+    "@storybook/manager-webpack5": "^6.5.16",
+    "@storybook/source-loader": "^6.5.16",
+    "@storybook/theming": "^6.5.16",
+    "@storybook/vue3": "^6.5.16",
+    "@vue/compiler-sfc": "^3.2.47",
```
braddialpad: Hmmm I think we don't need this anymore do we? Unless maybe storybook doesn't work without it?

https://www.npmjs.com/package/@vue/compiler-sfc

> Note: as of 3.2.13+, this package is included as a dependency of the main vue package and can be accessed as vue/compiler-sfc. This means you no longer need to explicitly install this package and ensure its version match that of vue's. Just use the main vue/compiler-sfc deep import instead.

josedialpad: Good catch, it's no longer need to explicitly define here

braddialpad (approved):

PR #742, opened 2023-02-03
fix(combobox-multi-select): wrong chips position while on modal

juliodialpad:
# Fix combobox multi select chips position while on modal

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Reverted the setChipsTopPosition logic, the only change that we really needed was the chips size and the input padding.

Tested Default, with Max select validation and inside a modal, everything should be working fine now.

## :bulb: Context

A bug was introduced on #736 when changed the way we were calculating the chips position.

## :pencil: Checklist

- [ ] I have added tests


braddialpad (approved):
Tested this on ubervoice using npm link, seems to work great!

PR #745, opened 2023-02-03
feat(avatar): group variant

josedialpad:
# Avatar group variant

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Vue implementation of avatar group and refactor implemented in https://github.com/dialpad/dialtone/pull/813/files.

## :pencil: Checklist

- [x] I have added tests


components/avatar/avatar.vue
```diff
@@ -114,6 +129,25 @@ export default {
...
+    /**
+     * Determines whether to show a group avatar.
+     * Limit to 2 digits max, though supports “99+”.
+     * if the number is 1 or less it would just show the regular avatar as if group had not been set.
+     */
+    group: {
+      type: [Number, String],
+      default: undefined,
+      validator: (group) => AVATAR_GROUP_VALIDATOR(group),
+    },
```
braddialpad: the control input renders as an "object" in storybook I think due to the multiple types. Probably not even necessary to allow both types, maybe only allow "number"

components/avatar/avatar.vue
```diff
@@ -3,20 +3,34 @@
     :id="id"
     :class="[
       'd-avatar',
-      AVATAR_KIND_MODIFIERS[kind],
       AVATAR_SIZE_MODIFIERS[size],
       avatarClass,
+      gradient ? null : 'd-avatar--no-gradient',
+      showGroup ? 'd-avatar--group' : null,
```
braddialpad: use object syntax rather than a ternary here

```
:class="[
  'd-avatar',
  AVATAR_SIZE_MODIFIERS[size],
  avatarClass,
  {
    'd-avatar--no-gradient': !gradient,
    'd-avatar--group': showGroup,
  },
]"
```

components/avatar/avatar.vue
braddialpad: I think we can also just set it to 99+ when any number greater than 99 is entered.

braddialpad (changes requested):
Lookin nice, just some changes around the group number type/math/validation

components/avatar/avatar.vue
josedialpad: Good call, refactored and moved to a computed prop as that's the convention we've been using https://github.com/dialpad/dialtone-vue/pull/745/commits/3cd94c9122b5ec6454d465245239f14248354624

components/avatar/avatar.vue
josedialpad: yeah followed the approach you've suggested, thanks https://github.com/dialpad/dialtone-vue/pull/745/commits/541fb7968a722db9b7ac493ec648f1f17f588c99

braddialpad (approved):

braddialpad:
Remember to update to the new dialtone (once it is released) before merging

PR #746, opened 2023-02-06
feat(list-item): add selected prop

iropolo:
# PR feat(list-item): add selected prop

- [x] Feature

## :book: Description

[DT-527](https://dialpad.atlassian.net/browse/DT-527)

We added `selected` prop on the list-item component.
This prop enable the possibility to the user to quickly mark a list item as selected.

This add styles and a check mark on the right.

## :pencil: Checklist

- [x] I have added tests

image

image

image

image

[DT-527]: https://dialpad.atlassian.net/browse/DT-527?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iropolo:
Im wondering if this way its useful or it would be better if we manage this on the default_list_item component as a prop.
We would need to add the `selected` attribute on the list_item then the default_list_item would manage the style and check mark icon.

With this PR we are creating a slot called `selected` and we are using this slot if `selected` prop is provided. 
In this case, list_item is in charge of the logic.
But, we also are giving the possibility to the user to use this `selected` slot, I dont know if this should be on this way.

You can test the behavior [here](http://localhost:9010/?path=/story/components-list-item--default ) and switch the `select` prop boolean 


image

image

image

components/list_item/list_item.vue
```diff
@@ -24,6 +25,14 @@
...
         <slot :name="slotName" />
       </template>
+      <template
+        v-if="selected"
+        #selected
+      >
+        <dt-icon
+          name="check"
+        />
+      </template>
```
josedialpad: Slot for the selected icon that defaults to the check icon?

```suggestion
      <template
        v-if="selected"
        #selected
      >
        <slot name="selectedIcon">
          <dt-icon
            name="check"
          />
        </slot>
      </template>
```

components/list_item/list_item.vue
```diff
@@ -224,14 +243,15 @@ export default {
   border-radius: var(--size-300);
 }
 
-.dt-list-item--focusable:focus,
-.dt-list-item--focusable:focus-within,
```
josedialpad: Shouldn't the selected variant re-style the background for these states rather than removing them from the default list item?

components/list_item/default_list_item.test.js
```diff
@@ -102,5 +102,16 @@ describe('DtDefaultListItem tests', function () {
       it('should render the slot wrapper', function () { assert.isTrue(bottomWrapper.exists()); });
       it('should render the provided content', function () { assert.isTrue(wrapper.text().includes('bottom')); });
     });
+
+    describe('When selected content is provided', function () {
+      // Test Setup
+      beforeEach(async function () {
+        slots = { selected: 'selected' };
```
josedialpad, line 109: We should add a test case for the selected prop in the list item checking that the default selected icon appears.

josedialpad (changes requested):

components/list_item/default_list_item.vue
```diff
@@ -38,6 +38,9 @@
       <!-- @slot Slot for the right content -->
       <slot name="right" />
     </section>
+    <template v-if="$slots.selected">
+      <slot name="selected" />
+    </template>
```
josedialpad: No reason to wrap in a template here

```suggestion
    <slot name="selected" />
```

josedialpad:
> Im wondering if this way its useful or it would be better if we manage this on the default_list_item component as a prop.
We would need to add the selected attribute on the list_item then the default_list_item would manage the style and check mark icon.

The styling is done above the list item, so the approach you've taken in this PR looks good to me.

braddialpad:
> Im wondering if this way its useful or it would be better if we manage this on the default_list_item component as a prop. We would need to add the `selected` attribute on the list_item then the default_list_item would manage the style and check mark icon.

Yep I 100% agree with this logic. default_list_item is meant to dictate styling while custom list item should really have no styling at all. custom list item probably shouldn't even do anything when selected is true by default, maybe just highlighting?

components/list_item/default_list_item.vue
iropolo: Oops, yep, that should not be there

components/list_item/list_item.vue
iropolo: Thats sth I though, but I dont know if there is a reason to have it customizable.

The idea is to have a common shared component through all our applications, so thats why I guess we should limit the flexibility here, and dont let the user change the 'checkmark icon' for selected item.

Maybe @braddialpad could check this out :)

francisrupert: I agree we should restrict to `<dt-icon name="check" />` — at least to begin with. At some point on Product Design side we'll inventory and standardize on any other visuals ways to show a checked state of a List Item. But yeah, there isn't currently any fundamental reason for it to be a different icon. If there is, that's what the custom option is for.

I'd also love to see examples of it reflected as an example on at least one [Dropdown](https://vue.dialpad.design/?path=/story/components-dropdown--variants)'s nested List items.

components/list_item/list_item.vue
iropolo: ---EDITED

Deprecated with latest changes :)

components/list_item/list_item.vue
```diff
@@ -224,14 +243,15 @@ export default {
...
 .dt-list-item--highlighted {
   background-color: hsla(var(--black-900-h), var(--black-900-s), var(--black-900-l), 0.05);
 }
 
-.dt-list-item--highlighted:active {
-  background-color: hsla(var(--black-900-h), var(--black-900-s), var(--black-900-l), 0.1);
+.dt-list-item--selected {
+  background-color: var(--purple-100);
```
francisrupert: Remove the background color, the presence of the check icon is sufficient. Plus, there may be multiple checked items and a background color muddies it. For example...

image

components/list_item/list_item.vue
```diff
@@ -224,14 +243,15 @@ export default {
...
+  .d-icon{
+    color: var(--purple-400);
```
francisrupert: Remove color, deferring to inherited color.

components/list_item/list_item.vue
```diff
@@ -24,6 +25,14 @@
         <!-- @slot named slots for custom list items -->
         <slot :name="slotName" />
       </template>
+      <template
+        v-if="selected"
+        #selected
+      >
+        <dt-icon
+          name="check"
```
francisrupert, line 32: Use `size="400"`.

francisrupert (changes requested):
Liking.

components/list_item/list_item.vue
iropolo: Thanks for your attention Francis :)

Here is some examples on the Dropdown implementation

---EDITED

Deprecated with latest changes :)

components/list_item/list_item.vue
francisrupert, line 32: Add `class="d-ml8"`

francisrupert (changes requested):

components/list_item/list_item.vue
iropolo: Since it will not have background color and the icon will have inherited color, I removed `.dt-list-item--selected` completely :)

braddialpad (approved):
Looking good now, thanks!

francisrupert (approved):
🤘

josedialpad (approved):

PR #755, opened 2023-02-09
fix(root-layout): sidebar independently scrollable

braddialpad:
# fix(root-layout): sidebar independently scrollable

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Got a report that while the main content was scrollable on overflow, the sidebar was not. Added overflow y auto to the sidebar and also made the entire main content section (sidebar + content) scrollable when at mobile width, as this makes more sense than what we were doing before.

## :pencil: Checklist

## :crystal_ball: Next Steps

Test in DP 2.0

### DIFF
--- a/components/root_layout/root_layout_body.vue
+++ b/components/root_layout/root_layout_body.vue
@@ -1,14 +1,15 @@
 <template>
   <div
     ref="root-layout-body"
-    :class="['root-layout__body', 'd-fl-grow1', bodyClasses]"
+    :class="['root-layout__body', { 'd-of-y-auto': isInMobileMode }, 'd-fl-grow1', bodyClasses]"
+    :style="{ 'height': isInMobileMode ? mainHeight : null }"
     data-qa="root-layout-body"
   >
     <aside
       v-if="$slots.sidebar"
       ref="root-layout-sidebar"
-      :class="['root-layout__sidebar', sidebarClass]"
-      :style="{ 'flex-basis': sidebarWidth }"
+      :class="['root-layout__sidebar', { 'd-of-y-auto': !isInMobileMode }, sidebarClass]"
+      :style="{ 'flex-basis': sidebarWidth, 'height': !isInMobileMode ? mainHeight : null }"
       data-qa="root-layout-sidebar"
     >
       <!-- @slot Slot for the sidebar -->
@@ -17,8 +18,8 @@
     <main
       v-if="$slots.content"
       ref="root-layout-content"
-      :class="['root-layout__content', contentClass]"
-      :style="{ 'min-inline-size': contentWrapWidthPercent, 'height': mainHeight }"
+      :class="['root-layout__content', { 'd-of-y-auto': !isInMobileMode }, contentClass]"
+      :style="{ 'min-inline-size': contentWrapWidthPercent, 'height': !isInMobileMode ? mainHeight : null }"
       data-qa="root-layout-content"
     >
       <!-- @slot Slot for the main content -->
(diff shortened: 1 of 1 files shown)


components/root_layout/root_layout_body.vue
```diff
@@ -1,14 +1,15 @@
 <template>
   <div
     ref="root-layout-body"
-    :class="['root-layout__body', 'd-fl-grow1', bodyClasses]"
+    :class="['root-layout__body', { 'd-of-y-auto': isInMobileMode }, 'd-fl-grow1', bodyClasses]"
+    :style="{ 'height': isInMobileMode ? mainHeight : null }"
```
modialpad, line 5: Just tested this and looks good, just have a few thoughts on this

Can we use media queries to check for mobile? I think many of the styles here (except the dynamic `height`) can be updated to use media queries that test against the current device. So the for the `root-layout__body` class, we'd have something like this:

```css
.root-layout__body {
  display: flex;
  /* (phones, 400px and down) */
  @media only screen and (max-width: 400px) {
     flex-direction: column;
  }
}
/* all styles below use media queries as a cursor for whether we're on mobile or not */
```

So this would shift our mobile checking from JS to CSS. I'm wary of making the performance argument here for this since I haven't ran tests, nor do I think the kind of JS we do now is complex in any way, but using media queries is IMO more standard and helps keep JS limited to non-style concerns.

I don't think we can achieve this dynamic height required when `fixed` prop is passed through CSS alone. So we might just leave this as it is. Likewise for `sidebarPosition`, although for that what I was thinking originally was to have two different slots for a sidebar on each side of the main content - one named `leftbarsidebar` and another `rightsidebar`, so depending on what the consumer wants they just initialize the slot they want so we don't need to deal too much with inverting anything.

My 2 cents, I don't want to block this since it works but might be worth considering as we evolve this component.

braddialpad: Yeah we definitely should use pure css as much as possible here. David was also talking to me about this yesterday. If we have ways to remove JS logic it would be great. 

Currently we use the `contentWrapWidthPercent` to determine the point at which the layout "stacks" which is done when the main content reaches this specified percentage of width compared to the sidebar. This is done with pure css and is perhaps an overly "fancy" way of doing it compared to standard media queries, the issue is I was unable to determine whether you are in "mobile mode" or not without javascript.

We could change this to a media query at a specified pixel width as you have mentioned, and I think this probably is a good idea, but unfortunately this would not get rid of all our javascript because we still have to set the height of the main section based on the heights of the header and footer. I was not able to find a pure css way of doing this so far but please let me know if you have any ideas.

The reason we didn't include a right sidebar is because it is not static like the leftbar is and is generally dependent on dynamic content in the main section. So the right sidebar should be "within the main" so to speak.

francisrupert: Definitely recommend an intentional approach to pure CSS for this kind of thing. Marshall is currently leading the effort to explore responsive views at this level; e.g. collapsed state for left sidebar, etc.

Getting ahead of that would go a long way to setting that up for success.

components/root_layout/root_layout_body.vue
```diff
@@ -120,20 +121,19 @@ export default {
...
-    // in the main content height calculation. Otherwise it is 0 since it is at equal height with the main content.
-    extraSidebarHeight () {
+    // When the sidebar is above the header due to contentWrapWidthPercent it is considered to be in mobile mode.
+    isInMobileMode () {
       if (this.contentTop > this.sidebarTop) {
-        return this.$refs['root-layout-sidebar'].offsetHeight + 'px';
+        return true;
       }
-      return '0px';
+      return false;
```
josedialpad, line 134: nit:

`return this.contentTop > this.sidebarTop`

components/root_layout/root_layout_body.vue
```diff
@@ -1,14 +1,15 @@
 <template>
   <div
     ref="root-layout-body"
-    :class="['root-layout__body', 'd-fl-grow1', bodyClasses]"
+    :class="['root-layout__body', { 'd-of-y-auto': isInMobileMode }, 'd-fl-grow1', bodyClasses]"
```
josedialpad, line 4: One suggestion to try to clean up a bit the template is to apply a `stacked` modifier class to the body and define the overflow styles based on this class. Example:

```css
.root-layout__body {
  &.root-layout__body--stacked {
    overflow-y: auto;
    
    .root-layout__sidebar, .root-layout__content {
      overflow-y: unset
    }
  }
}

.root-layout__sidebar, .root-layout__content {
  overflow-y: auto;
}
```

josedialpad (approved):
LGTM if we're not approaching the migration to a CSS approach in this PR.

components/root_layout/root_layout_body.vue
josedialpad: I guess the `height` can also be added in these CSS classes if they're defined in the component with https://vuejs.org/api/sfc-css-features.html#v-bind-in-css

braddialpad:
Merging in what's here for now, created a ticket to rework this component with pure CSS.

PR #758, opened 2023-02-11
feat(stack): new stack component

iropolo:
# feat(stack): initial approach

- [x] Feature

## :book: Description

We build a new Dialtone component called 'Stack'

Stack is a composable component that's helpful for creating layouts with equal spacing between elements. It can create horizontal or vertical stacks of elements.

Quite literally can be the basis for nearly any Dialpad product column or row of content/component. Follows model for how Product Design structures parts in Figma (auto-layout, which 98% emulates flexbox).

This component have 3 customizable props

dir: Set this prop to the direction to stack the items. You can override the default direction with 'default' key. All the undefined breakpoints will have 'default' value

as: Set this prop to render a specific HTML element.

gap: Set this prop to have the space between each stack item

## :pencil: Checklist

- [ ] I have added tests


components/stack/stack.vue
```diff
@@ -0,0 +1,286 @@
...
+  },
+
+  computed: {
+    spacing () {
+      return this.gap;
+    },
+
+    stackGap () {
+      return { '--stack-gap': DT_STACK_GAP[this.spacing] };
+    },
```
iropolo, line 63: Will v-bind on vue 3

https://learnvue.co/tutorials/reactive-styles-vue-3

components/stack/stack.vue
```diff
@@ -0,0 +1,286 @@
...
+</script>
+
+<style lang="less">
+.dt-stack {
+  display: flex;
+
+  &--column {
+    flex-direction: column;
+    & > :not(style) ~ :not(style) {
+      margin-top: var(--stack-gap);
```
iropolo: Will use v-bind on vue 3

https://learnvue.co/tutorials/reactive-styles-vue-3

components/stack/stack.vue
```diff
@@ -0,0 +1,286 @@
...
+        flex-direction: row-reverse;
+        & > :not(style) ~ :not(style) {
+          margin-top: 0;
+          margin-inline: 0 var(--stack-gap);
+          margin-bottom: 0;
+        }
+      }
+    }
+  }
+}
```
braddialpad: I think the CSS definitely needs some refining here.

We should probably be using flexbox gap instead of margin, it has been supported in all major browsers since 2021 and will prevent you from having to do this crazy margin swapping stuff.

Here's an example from the story using [gap for a stack component](https://www.leereamsnyder.com/blog/row-stack-space-layout-components). There's also an every-layout article but I think it's a bit dated and was from before gap was supported in browsers.

components/stack/stack_constants.js
```diff
@@ -0,0 +1,19 @@
...
+export const DT_STACK_RESPONSIVE_BREAKPOINTS = ['sm', 'md', 'lg', 'xl'];
+
+export const DT_STACK_GAP = {
+  default: '0',
+  100: '0.1rem',
+  200: '0.2rem',
+  300: '0.4rem',
+  400: '0.8rem',
+  500: '1.6rem',
+  600: '3.2rem',
```
braddialpad: Use the actual CSS vars here ex `var(--space-300)`

components/stack/stack.vue
```diff
@@ -0,0 +1,286 @@
+<template>
+  <component
+    :is="as"
+    :class="[
+      'dt-stack',
+      stackClasses(),
+    ]"
+    :style="stackGap"
```
francisrupert: Rather than doing an inline `style` attribute of `style="--stack-gap:{val}"` have them toggle between...

```
dt-stack--gap-100
dt-stack--gap-200
dt-stack--gap-300
dt-stack--gap-400
dt-stack--gap-500
dt-stack--gap-600
```

Example:

```html
<div class="dt-stack dt-stack--gap-200">
  <div>item</div>
  <div>item</div>
  <div>item</div>
</div>
```

...and `default` (none) doesn't have a `dt-stack--gap` class at all.

components/stack/stack.vue
francisrupert: Correct, that old-timer lobotomized owl (`* + *`) technique is mostly deprecated. Here's a suggested simplification:

```less
.d-stack {
  --stack-gap: 0;
  --stack-direction: column;
  
  display: flex;
  flex-direction: var(--stack-direction);
  gap: var(--stack-gap);

  &--column-reverse {
    --stack-direction: column-reverse;
  }

  &--row {
    --stack-direction: row;
  }

  &--row-reverse {
    --stack-direction: row-reverse;
  }

  &--gap-100 { --stack-gap: var(--space-100); }
  &--gap-200 { --stack-gap: var(--space-200); }
  &--gap-300 { --stack-gap: var(--space-300); }
  &--gap-400 { --stack-gap: var(--space-400); }
  &--gap-500 { --stack-gap: var(--space-500); }
  &--gap-600 { --stack-gap: var(--space-500); }
}
```

(though I didn't do the media queries).

components/stack/stack_default.story.vue
```diff
@@ -0,0 +1,62 @@
...
+          <p class="d-fs-200 d-fw-bold d-fc-black-800">
+            3 = Lorem ipsum dolor
+          </p>
+          <p class="d-fs-100">
+            Lorem ipsum dolor sit amet, consectetur adipiscing elit.
+            Donec fermentum molestie semper. Morbi finibus nulla turpis, nec molestie mi rutrum.
+          </p>
+        </div>
+      </template>
+    </dt-card>
```
francisrupert: Let's simplify the sample stack items. Let's start with this:

```
<div class="d-bgc-magenta-100">Stack item 1</div>
<div class="d-bgc-magenta-100">Stack item 2</div>
<div class="d-bgc-magenta-100">Stack item 3</div>
```

Column
image

Row:
image

components/stack/stack.vue
```diff
@@ -0,0 +1,286 @@
...
+      default: 'column',
+    },
+
+    /**
+     * Set this prop to render stack as a specific HTML element.
+     */
+    as: {
+      type: String,
+      default: 'div',
+    },
```
francisrupert, line 42: If the value is blank, it should revert to the default (`div`). Right now deleting the value removes it outright.

image

components/stack/stack.vue
francisrupert: @braddialpad,

My instinct was the style would ultimately be in `dialtone.css` (`.d-stack`) and then used by `dt-stack`. Whatchathink?

Looking forward, I can see this refactoring existing (and future) components. And _definitely_ throughout Vue components and recipes. Left sidebar is actually a great early destination for it. 

Like I quoted before: _Once you think in stacks, you'll see them everywhere_. Here's a sidebar row you're currently working...

```
<dt-stack gap="300">
  <dt-avatar />
  <dt-stack>
    <div>Ben Parker</div>
    <div>In a meeting</div>
  </dt-stack>
  <dt-badge />
</dt-stack>
```

francisrupert (commented):
Great start! A few design-perspective comments. Keep at it!

components/stack/stack.vue
iropolo: ---
It will use utility class, no more var :)

components/stack/stack.vue
braddialpad: > My instinct was the style would ultimately be in `dialtone.css` (`.d-stack`) and then used by `dt-stack`. Whatchathink?

Yep I think this was actually Ignacio's intention also, he just started out with the Vue component and added the styles there meaning to move them to dialtone later

components/stack/stack.vue
iropolo: Nice catch, I completely forget the `as` fallback

components/stack/stack.vue
francisrupert: Coolness. In that case, name is `.d-stack`.

components/stack/stack_default.story.vue
iropolo: Super! thats what I was looking to replace the owl stuff!!

components/stack/stack.vue
iropolo: Yep, as Brad said it was here just to show the proper behaviour, will move the style to dialtone as `d-stack`

iropolo:
Awesome feedback, lets go with another review round.
Check if the expected behaviour works properly and the style refactor looks ok, I would move those to Dialtone and create a PR there.

Also check the documentation on the .mdx file

components/stack/stack_constants.js
iropolo: With the style refactor we are not using anymore this styles, we just use it for validation

components/stack/stack.vue
```diff
@@ -0,0 +1,204 @@
...
+      return this.dir;
+    },
+
+    stackGap () {
+      return DT_STACK_GAP[this.gap] ? `d-stack--gap-${this.gap}` : null;
+    },
+
+    stackElement () {
+      return this.as ? this.as : 'div';
+    },
```
braddialpad: Don't think this is necessary, we can just put `this.as` directly in the `is` as it defaults to `div`

components/stack/stack_constants.js
```diff
@@ -0,0 +1,18 @@
...
+export const DT_STACK_RESPONSIVE_BREAKPOINTS = ['sm', 'md', 'lg', 'xl'];
+
+export const DT_STACK_GAP = {
+  100: '100',
+  200: '200',
+  300: '300',
+  400: '400',
+  500: '500',
+  600: '600',
+};
```
braddialpad: It can just be an array now since it's all the same, validate using `.includes()`

components/stack/stack.vue
```diff
@@ -0,0 +1,204 @@
...
+export default {
+  name: 'DtStack',
+
+  props: {
+    /**
+     * Set this prop to the direction to stack the items.
+     * You can override the default direction with 'default' key.
+     * All the undefined breakpoints will have 'default' value
+     */
+    dir: {
```
braddialpad: Lets call this "direction" to avoid confusion. This makes me think "directory" (I know dir is written in the story lol)

components/stack/stack.vue
```diff
@@ -0,0 +1,204 @@
...
+      default: 'div',
+    },
+
+    /**
+     * Set this prop to have the space between each stack item
+     */
+    gap: {
+      type: String,
+      default: '0',
+    },
```
braddialpad: Validate to only allow the values specified in the story `0|100|200|300|400|500|600`

components/stack/stack.vue
braddialpad: Add validation `column|column-reverse|row|row-reverse`

components/stack/stack_constants.js
```diff
@@ -0,0 +1,18 @@
...
+  column: 'column',
+  row: 'row',
+  'row-reverse': 'row-reverse',
+  'column-reverse': 'column-reverse',
+};
+
+export const DT_STACK_RESPONSIVE_BREAKPOINTS = ['sm', 'md', 'lg', 'xl'];
+
+export const DT_STACK_GAP = {
+  100: '100',
```
braddialpad: 0 is a valid gap value so it should be included in this array. You'll have to change the handling in `stackGap` computed property a bit

components/stack/stack.vue
```diff
@@ -0,0 +1,204 @@
...
+      DT_STACK_DIR,
+      DT_STACK_GAP,
+      DT_STACK_RESPONSIVE_BREAKPOINTS,
+    };
+  },
+
+  computed: {
+    stackDir () {
+      return this.dir;
+    },
```
braddialpad: You can just use this.dir directly

components/stack/stack.vue
```diff
@@ -0,0 +1,204 @@
...
+        return defaultStyle !== 'column' ? `d-stack--${DT_STACK_DIR[defaultStyle]}` : null;
+      }
+    },
+
+    stackResponsive () {
+      return [
+        ...DT_STACK_RESPONSIVE_BREAKPOINTS.map((breakpoint) => {
+          return this.stackDir[breakpoint] ? `d-stack--${breakpoint}--${this.stackDir[breakpoint]}` : null;
+        })];
+    },
```
braddialpad: Any reason these can't just be computed?

components/stack/stack.vue
```diff
@@ -0,0 +1,204 @@
...
+        --stack-direction: column-reverse;
+      }
+
+      &--row {
+        --stack-direction: row;
+      }
+
+      &--row-reverse {
+        --stack-direction: row-reverse;
+      }
```
braddialpad: Since these nested media query values are always all the same we can just put them in a mixin to avoid repetition

```less
.direction-options() {
    &--column {
      --stack-direction: column;
    }
    &--column-reverse {
      --stack-direction: column-reverse;
    }
    &--row {
      --stack-direction: row;
    }
    &--row-reverse {
      --stack-direction: row-reverse;
    }
}
```

and then

```less
&--xl {
    @media screen and (max-width: 1264px) {
       .direction-options();
   }
}

&--lg {
    @media screen and (max-width: 980px) {
       .direction-options();
   }
}

// and so on

File: components/stack/stack.test.js [hunk 16]
```diff
@@ -0,0 +1,168 @@
...
+  });
+
+  // Teardown
+  afterEach(function () {
+    slots = baseSlotsData;
+  });
+
+  beforeEach(function () { _setWrappers(); });
+
+  describe('Presentation Tests', function () {
```
braddialpad, line 44: Nice tests, good pattern of setting state in describe/beforeEach and then asserting within the "it"

braddialpad (changes requested):
Few more comments, nice work so far. CSS looks much better.

components/stack/stack.vue
iropolo: It was just to quick fix reactivity on the storybook side.
If you remove the `div` value on the `as` prop, it will not render anything.

If you use this fallback, it will not detect nothing and then it will render a `div`.

Im also agree to just use `as` directly on the component.

components/stack/stack_constants.js
iropolo: I avoid to use a method instead use a 'object directly access' due performance, I believe it would be better to use Set and just check if it exists with .has

I would prefer to avoid as much as we can to use those methods on components if its possible.

The validation is inherit with the code behaviour, it is not necessary to have any validator.
If it dont find the key, it will not work.

Anyways since the impact is insignificant I can update it and use .includes for the validation.
Also the benefit to use validator is the error msg in console, it would help the developer :)

components/stack/stack.vue
iropolo: Yup, it probably was a leftover

components/stack/stack_constants.js
iropolo: I didnt put 0, because we dont have any validator here also 0 is the default value. I didnt see it necessary.

But now we will have validator on the props, so your POV its correct :)

components/stack/stack.vue
iropolo: I didnt see necessary to have the cache of the computed property working here.
But maybe it is if the template need to be reactive to some expensive change.

I can use it on Vue3, bcs it use proxy-based observer.

So.. to be consistent on both I can use computed here also :)

components/stack/stack.vue
iropolo: Excelent +1

components/stack/stack.vue
braddialpad: might as well

iropolo:
I believe it would be better if some github actions not run on _Draft PR_, it's not useful to have this `✔️ Deploy Preview ready!` doing comments on the PR 🫠

components/stack/stack.mdx
```diff
@@ -0,0 +1,92 @@
...
+### Available Directions
+
+```js
+['column', 'column-reverse', 'row', 'row-reverse']
+```
+
+### Available Breakpoints
+
+```js
+['sm', 'md', 'lg', 'xl']
```
braddialpad: Mention what px width these breakpoints are at

components/stack/stack.mdx
```diff
@@ -0,0 +1,92 @@
...
+  md: string,
+  lg: string,
+  xl: string,
+}
+```
+
+### Examples
+
+```html
+<dt-stack :dir="{'default': 'row', 'sm': 'column', 'lg': 'column-reverse'}" as="article" gap="100">
```
braddialpad: dir -> direction (few other cases below)

components/stack/stack.mdx
```diff
@@ -0,0 +1,92 @@
...
+  <div>Stack item 1</div>
+  <div>
+    <div>Stack item 2</div>
+      <dt-stack dir="row-reverse" gap="600">
+          <div>Stack item 3</div>
+          <div>Stack item 4</div>
+      </dt-stack>
+  </div>
+</dt-stack>
+```
```
braddialpad: Add short descriptions to these describing them ex: 
"stacks column at small screen size, but column reverse at large screen size",
"stack with gap 300"

components/stack/stack.vue
braddialpad: Hmm weird I'm surprised this wouldn't work in storybook. Going to look into it.

braddialpad (changes requested):

braddialpad:
> Hmm weird I'm surprised this wouldn't work in storybook. Going to look into it.

You sure this is the case? just tested locally with `:is="as"` and got this

components/stack/stack.vue
braddialpad: You sure this is the case? just tested locally with `:is="as"` and got this

iropolo: https://user-images.githubusercontent.com/89984179/219400002-64ef507c-da7a-4961-81ee-ea3381d94e88.mov

braddialpad:
> I believe it would be better if some github actions not run on _Draft PR_, it's not useful to have this `✔️ Deploy Preview ready!` doing comments on the PR 🫠

Yeah maybe we could add a label to disable it.

components/stack/stack.mdx
```diff
@@ -0,0 +1,101 @@
...
+
+### Available Breakpoints
+
+```js
+{ prefix: 'sm', mediaQuery: '(max-width: 480px)' },
+{ prefix: 'md', mediaQuery: '(max-width: 640px)' },
+{ prefix: 'lg', mediaQuery: '(max-width: 980px)' },
+{ prefix: 'xl', mediaQuery: '(max-width: 1264px)' },
+```
```
braddialpad: This is a bit confusing since there's no javascript object like this that actually exists. Probably can just remove it from the code tags and use a markdown table. Something like this:

```
| Prefix | Max Width |
| -- | ------ |
| sm | 480px  |
| md | 640px  |
| lg | 980px  |
| xl | 1264px |
```
Which renders like this

| Prefix | Max Width |
| -- | ------ |
| sm | 480px  |
| md | 640px  |
| lg | 980px  |
| xl | 1264px |

components/stack/stack.vue
braddialpad: Ah I see it's only when it's completely blank. Does it fix it if we add

```
stackElement () {
  return this.as ? this.as : 'div';
},
```

to `stack_default.story.vue` rather than the component itself?

iropolo: Ye sure, I didn't thing on the possibility of manipulate a little bit the default.story.
Im seeing you guys are doing this kind of things to not contaminate the component :)

braddialpad (approved):
Looks great thanks! Remember not to merge until the dialtone side changes are released.

PR #760, opened 2023-02-14
chore: update dialtone-icons and extract into constant

juliodialpad:
# Update dialtone-icons dependency

- [ ] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Updated dialtone-icons to latest.

## :bulb: Context

The newest version of dialtone-icons includes a fix to an error happening due to duplicated `clip-path` id's.

## :pencil: Checklist

- [ ] I have added tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,7 +87,7 @@
     "node": "16"
   },
   "dependencies": {
-    "@dialpad/dialtone-icons": "0.0.18",
+    "@dialpad/dialtone-icons": "latest",
     "@linusborg/vue-simple-portal": "^0.1.5",
     "emoji-regex": "^10.2.1",
     "emoji-toolkit": "^6.6.0",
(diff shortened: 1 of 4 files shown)


braddialpad (approved):
Change looks fine, but seems it has broken a unit test

juliodialpad:
> Change looks fine, but seems it has broken a unit test

Yes, I'm fixing it, dialtone-icons published with `vue 3` instead of `vue 2`

package.json
```diff
@@ -87,7 +87,7 @@
     "node": "16"
   },
   "dependencies": {
-    "@dialpad/dialtone-icons": "0.0.17",
+    "@dialpad/dialtone-icons": "latest",
```
braddialpad, line 90: Hmm I know it is annoying to update this manually all the time, but is it safe to do this? especially since this is a non-dev dependency?

It's possible we could have a bunch of stuff in staging, update dialtone-icons with a breaking change, and then release dialtone-vue unknowingly with this broken change in it. Also how does this work if a project has included dialtone-vue as a dependency? dialtone-icons would just update to the latest version when they npm install in product potentially breaking something?

juliodialpad: I don't think so, this is just to avoid npm errors between `vue3` and `latest` tags.

You still need to do `npm update @dialpad/dialtone-icons` to update this to latest version of that tag. If you do `npm install` it doesn't update the current version you have, but prevents `vue3` tag version updating to `latest` and breaking everything cause vue2/vue3 dependencies.

PR #769, opened 2023-02-16
feat(leftbar): add general row recipe

josedialpad:
# add general row recipe

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Jira ticket: https://dialpad.atlassian.net/browse/DT-958

Add Leftbar general row recipe.

Tests will be added in a separated PR (added here: https://github.com/dialpad/dialtone-vue/pull/771)

## :pencil: Checklist

- [x] I have added tests


recipes/leftbar/general_row/general_row.vue
```diff
@@ -0,0 +1,173 @@
...
+.leftbar-general-row {
+  padding: var(--su6) var(--su8);
+  height: var(--su24);
+  border-radius: var(--su24);
+  display: flex;
+  align-items: center;
+  justify-content: space-between;
+
+  &:hover {
+    background-color: var(--theme-sidebar-row-color-background-hover);
```
josedialpad: @francisrupert do we need to update this variable to `black 200` as it's in the design?

recipes/leftbar/general_row/leftbar_general_row_icon.vue
```diff
@@ -0,0 +1,95 @@
...
+    :class="contactCenterIconClasses"
+  />
+  <div
+    v-else-if="isDialbotType"
+    :class="dialbotClasses"
+  >
+    <icon-dialpad-glyph
+      class="d-svg--size16"
+      color="white"
+    />
```
josedialpad: Used the v6 icon here as it's not direct mapping in the dialtone-icons v7

braddialpad: Yeah closest we have is the dialpad brand icon from https://dialpad.design/design/icons/. I wouldn't worry about this at the moment.

recipes/leftbar/general_row/general_row.vue
```diff
@@ -0,0 +1,173 @@
...
+        data-qa="leftbar-general-row-description"
+      >
+        {{ description }}
+      </dt-emoji-text-wrapper>
+    </div>
+    <div class="d-d-flex d-ai-center d-pl8">
+      <dt-badge
+        v-if="hasUnreadMessages"
+        kind="count"
+        color="purple-400"
```
francisrupert: Don't need `color` prop anymore as `type` accounts for that now.

recipes/leftbar/general_row/general_row.vue
```diff
@@ -0,0 +1,173 @@
+<template>
+  <li
+    :class="leftbarGeneralRowClasses"
+    data-qa="leftbar-general-row"
+  >
```
francisrupert: In DP1.0, it's an `<a>` element, should that be retained? Is it a `<router-link>`?  Or is more accurately a `<button>` or...? I'm averse to it not being an actual actionable element. 

It may ultimately be in *within* an `<li>`, so let's choose that carefully.

How does DP2.0 see it, or did so in their wip? Since it's a rewrite, I imagine they'd be particular about that.

recipes/leftbar/general_row/general_row.vue
```diff
@@ -0,0 +1,173 @@
...
+        />
+      </div>
+      <dt-emoji-text-wrapper
+        :class="descriptionClasses"
+        data-qa="leftbar-general-row-description"
+      >
+        {{ description }}
+      </dt-emoji-text-wrapper>
+    </div>
+    <div class="d-d-flex d-ai-center d-pl8">
```
francisrupert: Let's avoid horizontal padding and margin where possible, favoring `gap`.

recipes/leftbar/general_row/general_row.vue
```diff
@@ -0,0 +1,173 @@
...
+export default {
+  name: 'DtRecipeGeneralRow',
+
+  components: {
+    DtEmojiTextWrapper,
+    DtBadge,
+    DtRecipeLeftbarGeneralRowIcon,
+  },
+
+  props: {
```
francisrupert, line 60: We'll need a `selected` boolean prop. [Figma reference](https://www.figma.com/file/oTY56LgQha8ljb4c8w4GtO/DT-Shared%3A-App-Components?node-id=9240%3A7131&t=fUu3sW7wEstPCHg4-11).

recipes/leftbar/general_row/general_row.vue
```diff
@@ -0,0 +1,173 @@
...
+      },
+    },
+
+    /**
+     * Number of unread messages
+     */
+    unreadCount: {
+      type: Number,
+      default: null,
+    },
```
francisrupert: Of note, this recipe doesn't account for the logic to display "99+" if the `number` is set to any number over 99, deferring to Product implementation. Just stating it here since I'm sure it'll jump out to someone.

recipes/leftbar/general_row/general_row.vue
```diff
@@ -0,0 +1,173 @@
...
+  },
+};
+</script>
+
+<style lang="less">
+.leftbar-general-row {
+  padding: var(--su6) var(--su8);
+  height: var(--su24);
+  border-radius: var(--su24);
+  display: flex;
```
francisrupert: This is where we'd use `gap` property, e.g. `gap: var(--size-400)` instead of e.g. `class="d-pl8"`

recipes/leftbar/general_row/general_row.vue
```diff
@@ -0,0 +1,173 @@
...
+      }
+    },
+  },
+};
+</script>
+
+<style lang="less">
+.leftbar-general-row {
+  padding: var(--su6) var(--su8);
+  height: var(--su24);
```
francisrupert: Avoid declaring height outright, deferring to careful coordination of the contents to result in a 24px height.

recipes/leftbar/general_row/general_row.vue
```diff
@@ -0,0 +1,173 @@
...
+          Object.keys(LEFTBAR_GENERAL_ROW_CONTACT_CENTER_COLORS).join(', '));
+      }
+    },
+  },
+};
+</script>
+
+<style lang="less">
+.leftbar-general-row {
+  padding: var(--su6) var(--su8);
```
francisrupert: Use `space` variable. `su` variables are deprecated, with a sunset target of... oh idk.

In this case...

```
padding: calc(var(--space-200) * 3) var(--space-400);
```

recipes/leftbar/general_row/general_row.vue
```diff
@@ -0,0 +1,173 @@
...
+  height: var(--su24);
+  border-radius: var(--su24);
```
francisrupert: `calc(var(--space-200) * 8)`

recipes/leftbar/general_row/general_row.vue
francisrupert: Might be the other way around actually.

@bhawkins-dialpad, looks like we need to update `--theme-sidebar-row-color-background-hover` — and I suppose the other theme vars, https://dialpad.design/design/colors/#theme. Have any insight on that?

recipes/leftbar/general_row/general_row.vue
```diff
@@ -0,0 +1,173 @@
...
+  }
+
+  &:active {
+    background-color: var(--theme-sidebar-active-row-color-background);
+  }
+
+  &__icon {
+    width: var(--su16);
+    height: var(--su16);
+  }
```
francisrupert: Remove. Icon size declared by `size` prop of `<dt-icon>`. Which should primarily be 18px (`300`), though dialbot and CC are different.

recipes/leftbar/general_row/general_row.vue
```diff
@@ -0,0 +1,173 @@
...
+
+  &__description {
+    font-size: var(--fs-200);
+    display: flex;
```
francisrupert: Add `line-height: var(--lh-400)`

recipes/leftbar/general_row/general_row.vue
```diff
@@ -0,0 +1,173 @@
...
+    align-items: center;
+    padding-left: var(--su8);
```
francisrupert: Probably can use `gap` property on parent instead of `padding-left`

recipes/leftbar/general_row/general_row_constants.js
```diff
@@ -0,0 +1,33 @@
...
+};
+
+export const LEFTBAR_GENERAL_ROW_CONTACT_CENTER_COLORS = {
+  purple: 'd-fc-purple-300',
+  blue: 'd-fc-blue-300',
+  green: 'd-fc-green-300',
+  red: 'd-fc-red-300',
+  magenta: 'd-fc-magenta-300',
+  gold: 'd-fc-gold-300',
+};
```
francisrupert: @bhawkins-dialpad, have some insight for how CC color blocks are chosen? I vaguely remember reading that they should be a random collection of like 64 colors. But _how_ they're determined may not fall on this recipe at all and maybe these are just prefab ones and actual color options are listed in product implementation.

recipes/leftbar/general_row/general_row_constants.js
```diff
@@ -0,0 +1,33 @@
...
+  DIALBOT: 'dialbot',
+};
+
+export const LEFTBAR_GENERAL_ROW_ICON_MAPPING = {
+  [LEFTBAR_GENERAL_ROW_TYPES.INBOX]: 'inbox',
+  [LEFTBAR_GENERAL_ROW_TYPES.CONTACTS]: 'contacts',
+  [LEFTBAR_GENERAL_ROW_TYPES.CHANNELS]: 'hash',
+  [LEFTBAR_GENERAL_ROW_TYPES.THREADS]: 'thread',
+  [LEFTBAR_GENERAL_ROW_TYPES.LOCKED_CHANNEL]: 'lock',
+  [LEFTBAR_GENERAL_ROW_TYPES.CONTACT_CENTER]: 'stop-filled',
```
francisrupert: [The Figma file](https://www.figma.com/file/oTY56LgQha8ljb4c8w4GtO/DT-Shared%3A-App-Components?node-id=9240%3A7131&t=fUu3sW7wEstPCHg4-11) didn't use an icon, and this might be brittle in the future if the `stop-filled` icon is redrawn to something that doesn't fit this use case.

We may draw an icon that is very specifically a "CC" icon and can be colorized. In the meantime, let's make it a raw square html element that is...

```
width: var(--icon-size-200);
height: var(--icon-size-200);
border-radius: var(--size-200);
```

recipes/leftbar/general_row/general_row.vue
```diff
@@ -0,0 +1,173 @@
...
+      return [
+        'leftbar-general-row',
+      ];
+    },
+
+    descriptionClasses () {
+      return [
+        'leftbar-general-row__description',
+        'd-mr-auto',
+        'd-truncate',
```
francisrupert: Ellipsis isn't appearing

image

francisrupert (changes requested):
Boy this is great stuff, satisfying to see it coming along. I've commented on multiple little things.

Meanwhile, as it relates to custom HTML/CSS vs CSS utilities, the tack we've aiming for lately and that I'd like to see us mature further going forward:

Components and Recipes should be well-structured HTML with carefully designed CSS, while CSS utilities are for quick manipulations and overrides — most likely in product. So for this one (and other Left Sidebar recipes) I'm thinking HTML structured something like this realized through non-overspecific CSS rather than inline css utilities.

```html
<el class="dt-left-sidebar-row dt-left-sidebar-row--general dt-left-sidebar-row--{OTHER_MODIFIERS}">
  <div class="dt-left-sidebar-row__primary">
    <div class="dt-left-sidebar-row__content">
      <div class="dt-left-sidebar-row__type">
        ICON | CC | AVATAR | DIALBOT
      </div>
      <div class="dt-left-sidebar-row__description">
        <div class="dt-left-sidebar-row__label">{Label}</div>
        <div class="dt-left-sidebar-row__meta">
          <div>{Context}</div>
          <div>{Personal}</div>
        </div>
      </div>
    </div>
  </div>
  <div class="dt-left-sidebar-row__secondary">
    <dt-badge />
  </div>
  <div class="dt-left-sidebar-row__action">
    {Call button}
  </div>
</el>
```

And since this is very specifically a recipe, that CSS remaining here and not in dialtone.css makes sense to me.

That said, I don't want to slow this effort down, and since it's a recipe and in Vue we conceivably could change that any time we like. Food for thought, would love yours too.

recipes/leftbar/general_row/general_row.vue
josedialpad: Good catch! Removed https://github.com/dialpad/dialtone-vue/pull/769/commits/7f18e6a399a36d9258ca89da9246b6ecfa8f07b5

recipes/leftbar/general_row/general_row.vue
josedialpad: Yeah that logic should be added in the consumer side. Added a note in the docs https://github.com/dialpad/dialtone-vue/pull/769/commits/592004f6baeb0ed9d090d7184292955169c0caeb

recipes/leftbar/general_row/general_row.vue
braddialpad: hmm yeah making it an `<a>` and then wrapping the `<a>` in a router-link doesn't really make sense since I believe the router link is an `<a>` itself.... Let me look a bit more into how router-link works, there might be some options to help with this.

recipes/leftbar/general_row/general_row.vue
braddialpad: hmm do we want to make this `String` then?

recipes/leftbar/general_row/general_row.vue
braddialpad: `text-overflow:ellipsis;` doesn't appear to work with `display: flex` @josedialpad can you remove `display: flex` from `leftbar-general-row__description` or will this cause problems with the UI?

recipes/leftbar/general_row/general_row.vue
braddialpad: In DP 2.0 the structure is

```html
<li>
  <RouterLink>
    <div>
```

So maybe we just leave this as a div that would be wrapped by routerlink

recipes/leftbar/general_row/general_row.vue
josedialpad: Thanks, updated https://github.com/dialpad/dialtone-vue/pull/769/commits/3f84fb8004075c9941e60d6dc6f6f0a4a16ee1e3

recipes/leftbar/general_row/general_row.vue
josedialpad: Good call, added https://github.com/dialpad/dialtone-vue/pull/769/commits/733bc9fdf4c16c545c61d6b998cd6b6983ed14b5

recipes/leftbar/general_row/general_row.vue
josedialpad: Yeah it should support a string, updated https://github.com/dialpad/dialtone-vue/pull/769/commits/bfc65f31845cd9bcc32fcf47cbf99d4be2cdfe52

recipes/leftbar/general_row/general_row.vue
josedialpad: Removed https://github.com/dialpad/dialtone-vue/pull/769/commits/f79c94dfc3371a35ae425844e79b397a28329394

recipes/leftbar/general_row/general_row.vue
josedialpad: Replaced https://github.com/dialpad/dialtone-vue/pull/769/commits/114abca9ebb3171ee2f626239920657ec76f62a2

recipes/leftbar/general_row/general_row.vue
josedialpad: Updated using `space` var though in figma it has 28px of radius so I've used `calc(var(--space-200) * 14)` https://github.com/dialpad/dialtone-vue/pull/769/commits/8eea5b89373452928b91b50a6c0706959dcf2665

recipes/leftbar/general_row/general_row.vue
josedialpad: Good call, updated the size for both icon types using the `size` prop https://github.com/dialpad/dialtone-vue/pull/769/commits/4e85a7d7d35b94b2aa1908ef248e0eaed342523c

recipes/leftbar/general_row/general_row.vue
josedialpad: Added https://github.com/dialpad/dialtone-vue/pull/769/commits/de13cd3eec7ab9fff27036038d4df114061f29d1

recipes/leftbar/general_row/general_row.vue
francisrupert: Heh, actually it's basically a "Pill" radius, i.e. `.d-bar-pill` ([CSS Utility](https://dialpad.design/utilities/borders/radius.html#pills)) which has a value of `100em`. 

Since we're doing it in the CSS (preferred) rather than as a CSS Utility, go ahead and make it `100em` (which basically just maxes out the roundedness.

It will be replaced by a Radius Design Token in the future.

recipes/leftbar/general_row/general_row.vue
josedialpad: Removed the display flex from the description https://github.com/dialpad/dialtone-vue/pull/769/commits/4c5fe4fa070767b7eebe9e29ae10f490c3f152d7

recipes/leftbar/general_row/general_row_constants.js
josedialpad: Hm yes we probably need a dedicated icon for the CC. Updated to use HTML for now https://github.com/dialpad/dialtone-vue/pull/769/commits/3679226bf2cf3c56001411cf4b2d48e1d577b0bb

recipes/leftbar/general_row/general_row.vue
josedialpad: gotcha, updated https://github.com/dialpad/dialtone-vue/pull/769/commits/4f58cf2f4d418d11471ce28ac558e9973699faf8

recipes/leftbar/general_row/general_row.vue
josedialpad: Updated with `gap` https://github.com/dialpad/dialtone-vue/pull/769/commits/1a56b1906b710f4ff5ed6fcf5888185fa6186a6f

recipes/leftbar/general_row/general_row.vue
josedialpad: thanks, updated https://github.com/dialpad/dialtone-vue/pull/769/commits/1a56b1906b710f4ff5ed6fcf5888185fa6186a6f

recipes/leftbar/general_row/general_row.vue
josedialpad: Updated to div so in the product you can wrap it with an anchor or router link https://github.com/dialpad/dialtone-vue/pull/769/commits/2ea42d1ddf06c03d9d0b9208136be33eab276b48

braddialpad:
Taking this over from Jose on Tuesday since he's gone for a week.

francisrupert:
@braddialpad,

Awesome. 

An update: the hover/focus I did assumed all rows have an `.dt-leftbar-row__action`. I updated the Less on [this  codepen](https://codepen.io/francisrupert/pen/QWVNvGX) to have a `--no-action`. 

Since the majority all row types (not just this recipe) will have an action (call button), I suppose a boolean prop defaulting to having no action... unless that shouldn't exposed as a prop and this "General" recipe would include the `--no-action` class name and not be exposed. And thus the other row recipes would have it exposed as an option to toggle (though I don't know how this recipe will play with those other ones). Your call.

recipes/leftbar/general_row/general_row.vue
```diff
@@ -0,0 +1,152 @@
+<template>
+  <div
+    :class="leftbarGeneralRowClasses"
+    data-qa="dt-leftbar-row"
+  >
+    <a
+      class="dt-leftbar-row__primary"
+      data-qa="dt-leftbar-row-link"
+      v-on="$listeners"
+    >
```
francisrupert: I happened to default to `<a>`, though it supports `<button>` too — and implicitly `<el role="button">`. What's your call on how it should be demo'ed in these recipes?

Meanwhile, omitting `href` removes focusability. if you do decide to retain as `<a>`, include a dummy `href="#link-to-nowhere"`.

And looking forward to the other recipes in [DT-934](https://dialpad.atlassian.net/browse/DT-934), you'll probably want to put a note in there to emulate the keyboard focus script in that codepen.

recipes/leftbar/general_row/general_row_constants.js
```diff
@@ -0,0 +1,32 @@
+export const LEFTBAR_GENERAL_ROW_TYPES = {
+  INBOX: 'inbox',
+  CONTACTS: 'contacts',
+  CHANNELS: 'channels',
+  THREADS: 'threads',
+  LOCKED_CHANNEL: 'locked channel',
+  CONTACT_CENTER: 'contact center',
+  DIALBOT: 'dialbot',
```
francisrupert, line 8: erk

image

recipes/leftbar/general_row/general_row_constants.js
```diff
@@ -0,0 +1,32 @@
...
+};
+
+export const LEFTBAR_GENERAL_ROW_ICON_MAPPING = {
+  [LEFTBAR_GENERAL_ROW_TYPES.INBOX]: 'inbox',
+  [LEFTBAR_GENERAL_ROW_TYPES.CONTACTS]: 'contacts',
+  [LEFTBAR_GENERAL_ROW_TYPES.CHANNELS]: 'hash',
+  [LEFTBAR_GENERAL_ROW_TYPES.THREADS]: 'thread',
+  [LEFTBAR_GENERAL_ROW_TYPES.LOCKED_CHANNEL]: 'lock',
+  'channel bold': 'hash-bold',
```
francisrupert: If there are unreads, the locked channel icon is `lock-filled`

image

image

recipes/leftbar/general_row/leftbar_general_row_icon.vue
```diff
@@ -0,0 +1,103 @@
...
+}
+
+.leftbar-general-row__dialbot-icon {
+  background-color: var(--purple-400);
+  width: 100%;
+  height: 100%;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+}
```
francisrupert: May as well replace this with what's [in my codepen](https://codepen.io/francisrupert/pen/QWVNvGX).

francisrupert (changes requested):

recipes/leftbar/general_row/general_row.vue
francisrupert: Aside: are [these really three different recipes](https://dialpad.atlassian.net/browse/DT-934) though?

recipes/leftbar/style/leftbar_row.less
```diff
@@ -0,0 +1,206 @@
+.dt-leftbar-row {
+
+  // TODO: reconcile Theme colors, which probably means use them but
+  //       update https://dialpad.design/design/colors/#theme in dialtone.css
+  //       to reflect the new design
+
+  //  ============================================================================
+  //  $   CSS CUSTOM PROPERTIES
+  //  ----------------------------------------------------------------------------
+  --leftbar-row-color-foregrond: var(--fc-primary);
```
francisrupert: `foregrond` => `foreground`

recipes/leftbar/style/leftbar_row.less
```diff
@@ -0,0 +1,206 @@
+.dt-leftbar-row {
+
+  // TODO: reconcile Theme colors, which probably means use them but
+  //       update https://dialpad.design/design/colors/#theme in dialtone.css
+  //       to reflect the new design
```
francisrupert, line 5: Logged as its own task https://dialpad.atlassian.net/browse/DT-988.

recipes/leftbar/style/leftbar_row.less
```diff
@@ -0,0 +1,206 @@
...
+  //  $   VARIANTS
+  //  ----------------------------------------------------------------------------
+  //
+  //  <div class="dt-leftbar-row dt-leftbar-row--{$VARIANT}">
+  //    ...
+  //  </div>
+  //
+  &--has-unread {
+    --leftbar-row-description-font-weight: var(--fw-bold);
+    --leftbar-row-alpha-color-foreground: var(--leftbar-row-color-foregrond);
```
francisrupert: `foregrond` => `foreground`

recipes/leftbar/style/leftbar_row.less
```diff
@@ -0,0 +1,206 @@
...
+  // </div>
+  //
+  &__primary {
+    // contains extra style properties to support both <button> and <a>
+    display: flex;
+    align-items: center;
+    flex: 1;
+    width: 100%;
+    text-align: left;
+    color: var(--leftbar-row-color-foregrond);
```
francisrupert: `foregrond` => `foreground`

recipes/leftbar/general_row/general_row_constants.js
braddialpad: gross

recipes/leftbar/style/leftbar_row.less
```diff
@@ -0,0 +1,206 @@
...
+    line-height: var(--leftbar-row-status-line-height);
+    padding-bottom: var(--space-100);
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+  }
+
+  &__meta-context {
+    //
+  }
```
braddialpad, line 198: What is this empty `meta-context`?

recipes/leftbar/style/leftbar_row.less
```diff
@@ -0,0 +1,206 @@
...
+  //  ============================================================================
+  //  $   CSS CUSTOM PROPERTIES
+  //  ----------------------------------------------------------------------------
+  --leftbar-row-color-foreground: var(--fc-primary);
+  --leftbar-row-color-background: transparent;
+  --leftbar-row-radius: 100em;
+  --leftbar-row-alpha-color-foreground: var(--fc-secondary);
+  --leftbar-row-alpha-width: calc(var(--size-300) * 10);
+  --leftbar-row-alpha-height: calc(var(--size-300) * 9);
+  --leftbar-row-omega-min-width: var(--leftbar-row-alpha-width);
```
braddialpad, line 16: are greek letters like this just a way of referring to elements that "have no semantic name"?

recipes/leftbar/general_row/leftbar_general_row_icon.vue
braddialpad: Whoops missed this, should be there now.

recipes/leftbar/general_row/general_row_constants.js
braddialpad: got it, should be there now.

recipes/leftbar/general_row/general_row.vue
braddialpad: I just changed this to a button for now since it seems there isn't really a way to fire an event with no action from a link (it is supposed to directly link to something). I'd rather not put `#link-to-nowhere` in there as it changes url history.

recipes/leftbar/general_row/general_row_constants.js
braddialpad: should be fixed.

recipes/leftbar/general_row/general_row.vue
braddialpad: > Aside: are [these really three different recipes](https://dialpad.atlassian.net/browse/DT-934) though?

We decided to do this pretty much solely for the purpose of simplicity, only exposing properties the consumer is likely to need. It's definitely possible for it to be a single component. As it is now I would likely use general row as a base for the two others.

braddialpad:
Updated!

recipes/leftbar/general_row/general_row.vue
francisrupert: Cool, figured it'd be extended by others. I'm sure the other ones will get done right quick.

recipes/leftbar/style/leftbar_row.less
francisrupert: Yeah, abstracted. `alpha` and `omega` for "first" and "last" is a common CSS naming convention, especially BEM.

```html
<el class="dt-leftbar-row__primary"...>
  <div class="dt-leftbar-row__alpha"> ... </div>
  <div class="dt-leftbar-row__label"> ... </div>
  <div class="dt-leftbar-row__omega"> ... </div>
</el>
```

recipes/leftbar/style/leftbar_row.less
francisrupert: That's the "In a meeting" portion. It doesn't have any specific style, though I still include it — future-proofing. In Product, its conditional style would apply `d-fc-critical || d-fc-warning || etc`.

```html
<div class="dt-leftbar-row__status">
  <span class="dt-leftbar-row__meta-context d-fc-critical">In a meeting</span>
  <span class="dt-leftbar-row__meta-custom" id="status_meeting">Working from SF</span>
</div>
```

recipes/leftbar/general_row/general_row.vue
```diff
@@ -3,7 +3,7 @@
     :class="leftbarGeneralRowClasses"
     data-qa="dt-leftbar-row"
   >
-    <a
+    <button
       class="dt-leftbar-row__primary"
```
francisrupert, line 7: `type="button"` attribute?

recipes/leftbar/general_row/leftbar_general_row_icon.vue
```diff
@@ -45,59 +44,41 @@ export default {
...
+      return ICON_MAPPING[this.type];
     },
 
     contactCenterIconClasses () {
       return [
-        'd-icon',
-        'leftbar-general-row__contact-center-icon',
-        LEFTBAR_GENERAL_ROW_CONTACT_CENTER_COLORS[this.color],
+        'd-bar2',
+        'dt-leftbar-row-icon-cc--size',
```
francisrupert: `dt-leftbar-row-icon-cc--size` → `dt-leftbar-row__icon-cc`

BEM convention `__` is a _child_ of `dt-leftbar-row` (albeit not a direct child). No need to be specifc to `size` as it may include more properties.

In fact, you can remove `d-bar2` and add `border-radius: var(--size-200)` to it. 

Aside: ultimately with Design Tokens all radii would replace those, i.e. `border-radius: var(--dt-size-radius-200)`

recipes/leftbar/style/leftbar_row.less
```diff
@@ -203,4 +203,9 @@
       color: var(--fc-tertiary);
     }
   }
+
+  &-icon-cc--size {
+    width: calc(var(--size-300) * 3.5);
+    height: calc(var(--size-300) * 3.5);
+  }
```
francisrupert: See previous comment about `dt-leftbar-row-icon-cc--size`

francisrupert (changes requested):

recipes/leftbar/general_row/general_row.vue
braddialpad: I added a `v-bind="$attrs"` there so any attributes set on the component will go to this element. Would we really want to allow for setting something like `<div role="button">` though?

francisrupert: I of course bias toward `<button>` in place of `role="button"` at all opportunities. I don't have the knowledge for how product eng teams default to one or the other, or why; I defer to you on that. Though I imagine Vue natively reimplements the `<button>` native browser functionality.

One bugaboo about using `<button>` is nesting block-level HTML elements is invalid (took a while for the validation standard to allow block-level items in `<a>`). Browsers forgive this and it's effectively not an issue, and I see it all the time. 

That is but one common reason for choosing `role="button"` over `<button>`. I suppose one fix is to convert every child `<div> ` element with `<span>`, but... some components used within default to `<div>` like Avatar (which I suppose can be `<span>`. It's a bit of a spiraling consideration for something seemingly simple.

francisrupert (approved):
Approved. Look forward to seeing this efficiently driving [DT-959](https://dialpad.atlassian.net/browse/DT-959), [DT-960](https://dialpad.atlassian.net/browse/DT-960), and [DT-988](https://dialpad.atlassian.net/browse/DT-988)

PR #770, opened 2023-02-17
fix(popover): prevent viewport from cutting off modal popover

braddialpad:
# fix(popover): prevent viewport from cutting off modal popover

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

It is possible for a modal popover to be "cut off" by the viewport if there is a large amount of content in it or the browser window is small. This change uses and intersectionObserver to detect when this happens and set maxHeight on the popover. This should be compatible with the existing maxHeight prop as this will only trigger and set the height when the dialog has crossed the viewport threshold.

This is only necessary for modal popovers as with non modal popovers you can still scroll the window.

## :bulb: Context

Fixes an issue noticed with the popover during product leftbar changes.

## :pencil: Checklist

## :crystal_ball: Next Steps

Try out in product

## :link: Sources

https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API


iropolo:
How are the steps to reproduce it?

Im trying with `contentWidth` on `anchor` and some content on `content` slot.
But when I try to reduce the viewport the popover just close.

braddialpad:
> 

If the anchor scrolls out of view the popover will close (and this is intended)

Try popping out the window by clicking this, and then just copy pasting a bunch of content in the DOM manually

Screenshot 2023-02-17 at 1 04 33 PM

juliodialpad:
I'm not able to reproduce this either, whenever I change the viewport size the popover closes, I've tried setting the `hideOnClick` prop to false to prevent this.


https://user-images.githubusercontent.com/87546543/220168137-a1994321-6e6b-4945-b629-19bdc89f4d55.mov

components/popover/popover.vue
```diff
@@ -607,6 +628,18 @@ export default {
    *     METHODS    *
    ******************/
   methods: {
+
+    hasIntersectedViewport (entries, observer) {
+      const dialog = entries[0];
+
+      // if dialog is cut off from the viewport
+      if (!dialog.isIntersecting) {
```
iropolo: Suggested change:

```
hasIntersectedViewport (entries) {
      const dialog = entries?.[0];

      // if dialog is cut off from the viewport
      this.isOutsideViewport = !dialog?.isIntersecting;
    },
```

- `observer` parameter is not being used
- `entries?.[0];` on this way, we avoid throw an error if `entries` is undefined
- `this.isOutsideViewport = !dialog?.isIntersecting;` code simplification

iropolo:
> I'm not able to reproduce this either, whenever I change the viewport size the popover closes, I've tried setting the `hideOnClick` prop to false to prevent this.

Hello Julio!!

You will need to add more content to trigger the scroll or cut even more the viewport


https://user-images.githubusercontent.com/89984179/220174228-b4998ed2-7447-4824-9ede-282e6c844614.mov

iropolo:
I would suggest in the future remove this tippy (it also uses popper wich I dont preffer) and instead build our own popper based on `floating ui` maybe.

braddialpad:
> I would suggest in the future remove this tippy (it also uses popper wich I dont preffer) and instead build our own popper based on `floating ui` maybe.

Yep this might be a good idea for the future

components/popover/popover.vue
braddialpad: great thanks!

braddialpad:
Investigating broken unit tests

braddialpad:
Tests should be fixed now. Added a global Resize/Intersection observer mock since this errors out if you have components using it.

iropolo (approved):
LGTM, nice work 👍

PR #772, opened 2023-02-18
fix(popover, tooltip): support shadow dom

braddialpad:
# fix(popover, tooltip): support shadow dom

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Had this change released on beta for Kare team a while back. Integrating it into the main release now since it seems to be working for them. 

Use [getRootNode](https://developer.mozilla.org/en-US/docs/Web/API/Node/getRootNode) instead of `document.body` to reference the body. Changed default appendTo prop to be string 'body' which causes `getRootNode` to be used to get the "nearest" body.

Added appendTo to dropdown as well for consistency

## :bulb: Context

Because we were appending to `document.body` it was causing a number of issues in the kare shadow dom containers with popovers/tooltips rendering at the app body rather than shadow dom body. Errors being caused by looking for components in `document.body`.

## :pencil: Checklist

- [x] I have added tests

### DIFF
--- a/components/dropdown/dropdown.stories.js
+++ b/components/dropdown/dropdown.stories.js
@@ -76,6 +76,14 @@ export const argTypesData = {
       },
     },
   },
+  appendTo: {
+    defaultValue: 'body',
+    table: {
+      defaultValue: {
+        summary: 'body',
+      },
+    },
+  },
   padding: {
     defaultValue: 'none',
     control: {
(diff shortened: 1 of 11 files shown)


joaodialpad (approved):
LGTM, is there some beta version available for me to test this?

juliodialpad (approved):
This looks really good, don't forget to create the Vue 3 version of this.

PR #778, opened 2023-02-24
feat(leftbar): add group row recipe

juliodialpad:
# Add group row recipe

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Added `group-row` recipe, based on `general-row`.

Updated general row, adding a slot called `left` so we can use it to put the avatar inside it, it also gives `general-row` component a little freedom to extend it. 

## :pencil: Checklist

- [x] I have added tests

Screenshot 2023-02-24 at 12 27 46 p m


recipes/leftbar/general_row/general_row.vue
```diff
@@ -3,7 +3,7 @@
     :class="leftbarGeneralRowClasses"
     data-qa="dt-leftbar-row"
   >
-    <button
+    <dt-button
```
francisrupert: I intentionally didn't use `dt-button` to avoid CSS redundancies or overriding CSS. I inspected the rendered markup and it's adding structures that changes its integrity and could introduce unintended effects.

`d-btn` has sooo many styles, and there's no reason for them to be bundled in.

/c @braddialpad

recipes/leftbar/general_row/general_row.vue
```diff
@@ -12,11 +12,13 @@
...
-          :color="color"
-          data-qa="dt-leftbar-row-icon"
-        />
+        <slot name="left">
+          <dt-recipe-leftbar-general-row-icon
+            :type="getIcon"
+            :color="color"
+            data-qa="dt-leftbar-row-icon"
+          />
+        </slot>
```
francisrupert, line 21: Slot is pretty much what I thought of this as for General, and supports everything else about other Leftbar Recipes. Defer to @braddialpad.

recipes/leftbar/general_row/general_row.vue
```diff
@@ -45,11 +47,13 @@ import { LEFTBAR_GENERAL_ROW_TYPES, LEFTBAR_GENERAL_ROW_CONTACT_CENTER_COLORS }
 import { DtBadge } from '@/components/badge';
 import DtEmojiTextWrapper from '@/components/emoji_text_wrapper/emoji_text_wrapper.vue';
 import DtRecipeLeftbarGeneralRowIcon from './leftbar_general_row_icon.vue';
+import DtButton from '@/components/button/button.vue';
```
francisrupert: Independent of `d-btn` style concerns I mentioned in previous comment, is there functionality that using `<dt-button>` provides?

recipes/leftbar/group_row/group_row.vue
```diff
@@ -0,0 +1,88 @@
...
+    },
+
+    /**
+     * Initials to display on the avatar if avatarSrc is not provided or
+     * alt attr if avatarSrc is provided.
+     */
+    avatarInitials: {
+      type: String,
+      default: '',
+    },
```
francisrupert: Perhaps more a bug of Avatar's Group variant, but if the Group Avatar doesn't have an image the initials are not displayed at that size (Avatar size XS). 

image

Should be like so...

image
 
...which I get from this combo https://vue.dialpad.design/deploy-previews/pr-778/?path=/story/components-avatar--initials&args=size:xs;group:2

francisrupert (changes requested):
Awesome to see this quickly get built off General Row.

recipes/leftbar/general_row/general_row.vue
braddialpad: Yep agree

recipes/leftbar/group_row/group_row.vue
```diff
@@ -0,0 +1,88 @@
+<template>
+  <dt-recipe-general-row
+    :description="names"
+    :unread-count="unreadCount"
+    :selected="selected"
+    v-on="$listeners"
```
braddialpad, line 6: Not necessary, `v-on="$listeners"` is on the root element by default.

recipes/leftbar/group_row/group_row.vue
```diff
@@ -0,0 +1,88 @@
...
+    avatarSrc: {
+      type: String,
+      default: '',
+    },
+
+    /**
+     * Initials to display on the avatar if avatarSrc is not provided or
+     * alt attr if avatarSrc is provided.
+     */
+    avatarInitials: {
```
braddialpad, line 50: Should this be required? Sounds like it should be based on this comment.

recipes/leftbar/group_row/group_row_variants.story.vue
```diff
@@ -0,0 +1,79 @@
+<template>
+  <div class="d-stack16 d-w464">
```
braddialpad: Super minor nit, but why not use our new stack component and then use gap to space it out instead of padding?

recipes/leftbar/general_row/general_row.vue
braddialpad: I'm not sure I understand the question here, but I think this is a good pattern.

recipes/leftbar/general_row/general_row.vue
josedialpad: Besides supporting the different variants, the icon and its placement are the main functionality. But I don't think we'd need these in this case.

recipes/leftbar/group_row/group_row.vue
josedialpad: Yeah it was a bug in the Avatar group since with variant it changes the size to `xs` with CSS but the component size remains with the size passed with the `size` prop. Updated and fixed this.
Can you validate that [the logic here](https://github.com/dialpad/dialtone-vue/pull/778/commits/09abfa9bcaf1b5a89c60877486d1d7f3cb9a5c7c#diff-90b30c664802d78094183e71b1acbb8385122479a20e6052588cd5a6d422dff0R201) is okay?

recipes/leftbar/general_row/general_row.vue
josedialpad: Rollbacked to use the native `button` element https://github.com/dialpad/dialtone-vue/pull/778/commits/e8d00b8a4eaaf001585fe927a12f0f2f39cc2a61

recipes/leftbar/group_row/group_row.vue
josedialpad: Yeah it shouldn't be necessary. But by removing this, the click event isn't logged in the Actions in the SB panel. So I'm guessing it's not passing down the events to the root?

recipes/leftbar/group_row/group_row.vue
josedialpad: Yeah makes sense, updated https://github.com/dialpad/dialtone-vue/pull/778/commits/392fdcb3d3ba9fd73853b38f494053e8194c569b

recipes/leftbar/group_row/group_row.vue
braddialpad: Hmm wonder if it's different if the root is a vue component? it's fine just leave it then.

recipes/leftbar/general_row/general_row.vue
francisrupert: My comment was just me saying "cool cool cool". No question.

recipes/leftbar/general_row/general_row.vue
francisrupert: 👍 Yep, rip it out.

recipes/leftbar/group_row/group_row.vue
francisrupert: Logic looks sound to me. I played around with a few combinations of `size` and `group` to try to break it, e.g. https://vue.dialpad.design/deploy-previews/pr-778/?path=/story/components-avatar--initials&args=size:lg;group:2

Keep playing with those values to see if you can you break it.

recipes/leftbar/group_row/group_row_variants.story.vue
francisrupert: In fact, we should start using `<dt-stack>` as much as possible to stress-test it out of Beta.

josedialpad: good call, updated variant stories in the leftbar to use `dt-stack` https://github.com/dialpad/dialtone-vue/pull/778/commits/4e50a0857ab80e3bd3720e3c1503285e185976e3

braddialpad (approved):
Looks good by me!

francisrupert (approved):

PR #779, opened 2023-02-24
fix(combobox): popover close on scroll

iropolo:
# fix(combobox): popover close on scroll

- [x] Fix

## :book: Description

https://dialpad.atlassian.net/browse/DT-1003

Popover of combobox closes when try to scroll.

Now we are using a different way to catch if its clicking outisde our element.


recipes/comboboxes/combobox_with_popover/combobox_with_popover.vue
```diff
@@ -453,7 +453,7 @@ export default {
 
     onFocusOut (e) {
       // Check if the focus change was to another target within the combobox component
-      const isComboboxStillFocused = Object.keys(this.$refs).some(refName => {
+      const isComboboxStillFocused = e.relatedTarget === null || Object.keys(this.$refs).some(refName => {
```
braddialpad: hmm don't think this works because now it will never close when it "unfocuses". In most cases we want it to, just in the case of the scrollbar we don't. I think the problem is that the scrollbar exists on an element parent to all these refs.

Object.keys(this.$refs) is probably not a great way of doing this check. We should get the root element of the dialog which we can do with: `this.$refs.popover.closest('.tippy-box')` and then we also need to check the input `this.$refs.input`

replace `Object.keys(this.$refs)` with something like this

```
const focusElements = [this.$refs.popover.closest('.tippy-box'), this.$refs.input]
focusElements.some(refName .....
```

braddialpad (changes requested):

braddialpad (approved):

PR #780, opened 2023-02-27
feat(scroller): initial mvp

iropolo:
# feat(scroller): initial mvp

- [x] Feature

## :book: Description

Initial Scroller component released and useable based on the below acceptance criteria.

- Consumer can pass in the scrollable content.
- Scrolling to the top or bottom of the list fires an event.
- The consumer has a way to jump immediately to the top or bottom of the list.
- The consumer has a way to jump to a specific list item in the list.

## :pencil: Checklist

- [x] I have added tests


components/scroller/scroller.vue
```diff
@@ -0,0 +1,642 @@
+<template>
```
braddialpad: I'm guessing this is all forked library code? We cannot dump all this directly in the component, we should maintain a separate forked repo with the library code, make our modifications to it and publish it separately. This will allow us to merge in any updates and make it much easier to see exactly what we changed from the original code. This component is much too complex, it should be using the external library rather than implementing it.

components/scroller/scroller_default.story.vue
```diff
@@ -0,0 +1,59 @@
+<template>
+  <div>
+    <button @click="scrollToItem(100)">
+      Go to index 100
+    </button>
```
braddialpad, line 7: Is it possible to scroll to the items id rather than index?

components/scroller/scroller_default.story.vue
```diff
@@ -0,0 +1,59 @@
...
+    <button @click="scrollToItem(items.length - 1)">
+      Go to end
+    </button>
+
+    <dt-scroller
+      ref="scroller"
+      v-slot="{ item }"
+      class="scroller"
+      :items="items"
+      :item-size="32"
```
braddialpad: I'm a bit confused what item-size is here, is that the height of the row?

components/scroller/scroller_default.story.vue
braddialpad: or is this actually the id? since it looks like id equals index

components/scroller/scroller_default.story.vue
```diff
@@ -0,0 +1,59 @@
...
+      Go to index 100
+    </button>
+    <button @click="scrollToItem(0)">
+      Go to start
+    </button>
+    <button @click="scrollToItem(items.length - 1)">
+      Go to end
+    </button>
+
+    <dt-scroller
```
braddialpad, line 15: I think we may have to use the [dynamic scroller implementation](https://github.com/Akryum/vue-virtual-scroller/blob/master/packages/vue-virtual-scroller/README.md#dynamicscroller). For the feed list I don't think we are going to know sizes in advance.

iropolo:
Scope change, will use fork rather than implementation.
It will allow us to follow the library updates more effectively.

iropolo:
Fork not possible, will continue with implementation

components/scroller/scroller.vue
```diff
@@ -0,0 +1,299 @@
...
+            el.scrollTop = el.scrollHeight + 5000;
+            if (this.$_undefinedSizes === 0) {
+              this.$_scrollingToBottom = false;
+            } else {
+              requestAnimationFrame(cb);
+            }
+          });
+        };
+        requestAnimationFrame(cb);
+      });
```
braddialpad: Was this code you wrote yourself or part of the library?

components/scroller/scroller.vue
```diff
@@ -0,0 +1,299 @@
...
+        });
+      }
+      return result;
+    },
+  },
+
+  watch: {
+    items () {
+      this.forceUpdate();
+    },
```
braddialpad: Why do we have to force update here?

components/scroller/scroller_default.story.vue
```diff
@@ -0,0 +1,313 @@
...
+   display:block;
+  position:absolute;
+  top:0;
+  left:0;
+  height:100%;
+  width:100%;
+  overflow:hidden;
+  pointer-events:none;
+  z-index:-1
+ }
```
braddialpad: A bit confused what we are doing here. We are looking for attribute `data-v-b329ee4c`? Seems auto generated what is that?

components/scroller/scroller_default.story.vue
```diff
@@ -0,0 +1,313 @@
...
+</script>
+
+<style scoped>
+.scroller {
+  height: 200px;
+  width: 100%;
+  max-width: calc(var(--size-300) * 157);
+  margin: 20px 0;
+  border: 1px solid red;
+}
```
braddialpad: Would the consumer of the scroller be expected to set all these css properties? We might want to make these props if they're going to have to be set regularly. I'd think the width and the height of the scroller would be a very common thing to set.

components/scroller/scroller.vue
```diff
@@ -0,0 +1,299 @@
...
+    /*
+      * The tag to use for the items.
+     */
+    itemTag: {
+      type: String,
+      default: 'div',
+    },
+
+    /*
+      * The minimum size of an item.
```
braddialpad: let's specify that this is height if direction = vertical and width if direction = horizontal. Just "size" is a bit confusing.

components/scroller/scroller_default.story.vue
```diff
@@ -0,0 +1,313 @@
...
+  'lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec auctor, nisl eget ultrices aliquam, nunc nisl aliquet nunc, eget aliquam nisl ni loremlorem ipsum dolor sit amet, consectetur adipiscing elit. Donec auctor, nisl eget ultrices aliquam, nunc nisl aliquet nunc, eget aliquam nisl ni loremlorem ipsum dolor sit amet, consectetur adipiscing elit. Donec auctor, nisl eget ultrices aliquam, nunc nisl aliquet nunc, eget aliquam nisl ni loremlorem ipsum dolor sit amet, consectetur adipiscing elit. Donec auctor, nisl eget ultrices aliquam, nunc nisl aliquet nunc',
+  'lorem ipsum dolor sit amet, consectetur adipiscing elit. Donec auctor, nisl eget ultrices aliquam',
+  'lorem ipsum dolor sit amet',
+];
+
+const dynamicItems = ref([
```
braddialpad: Can probably just do a for loop to create this array.

components/scroller/scroller.vue
```diff
@@ -0,0 +1,299 @@
...
+                // @TODO remove when contentRect is deprecated
+                width = entry.contentRect.width;
+                height = entry.contentRect.height;
+              }
+              entry.target.$_vs_onResize(entry.target.$_vs_id, width, height);
+            }
+          }
+        });
+      });
+    }
```
braddialpad: Also seems like library code? Just saying that due to the references in it so correct me if I'm wrong.

components/scroller/scroller.test.js
```diff
@@ -0,0 +1,66 @@
+import { assert } from 'chai';
+import { shallowMount } from '@vue/test-utils';
+import DtScroller from './scroller.vue';
+
+// Constants
+// const baseSlotsData = {
+//   default: `<div data-qa="default-content">Test 1</div> <div>Test 2</div>`,
+// };
+
+describe('DtStack Tests', function () {
```
braddialpad: DtStack -> DtScroller

braddialpad (commented):
The public interface of the component looks pretty good. My main concern is it looks like there are some things in the scroller.vue component that are referencing variables directly in library code, but I'd like to hear your thoughts on it.

components/scroller/scroller_default.story.vue
iropolo: Oh ye, sure, it was just a quick stub to test the scroller, I will update it on the next commit

components/scroller/scroller.vue
iropolo: Yes, your right I will update it with the same on the library

 `display height (or width in horizontal mode) of the items in pixels used to calculate the scroll size and position.`

components/scroller/scroller_default.story.vue
iropolo: The only required CSS for the scroller to work properly is the height... 
You could also use height 100% here if the parent already has a fixed height.

For the item in the list you should use ` :item-size="32"` and also if you want the size with a CSS class.
Like Im doing it, with ` :item-size="32"` and `class="user" with height: 25px;`

If you dont want to set any size, you should set the scroller with `:dynamic="true"` and put a minimum height value with `:min-item-size="54"`

components/scroller/scroller_default.story.vue
iropolo: Ah yes, this is a leftover. I removed it.

I used it because it was on the CSS of the library... it was here because the vue-resize package generate a element with 'resize-observer' and this data `data-v-b329ee4c`... so in the library they just hide it.

You can check it here https://jsfiddle.net/Akryum/zqygwf3r/

components/scroller/scroller.test.js
iropolo: Ye I didnt start with the tests yet until the scroller works as we expect

components/scroller/scroller.vue
iropolo: Yes this is library code, Im still trying to trigger it to see if it necessary but I would like to remove it, it is used on scroller-item.. I need to test if it is useful for the correct behaviour

components/scroller/scroller.vue
iropolo: I removed `forceUpdate`, it is not necessary

components/scroller/scroller.vue
iropolo: No, this is from the library and it is used if it is necessary to have sth like a "Chat Mode".
I didn't tried it I would remove it with `getItemSize `

braddialpad: So if it's from the library I'm thinking it should be in `core-scroller.vue` right? `scroller.vue` should be completely separated from library code.

components/scroller/scroller_default.story.vue
braddialpad: I think the data-v-b329ee4c may change whenever they release a new build. Does it break anything if we remove this code?

components/scroller/scroller.vue
braddialpad: I'm not sure we should just be removing things if we don't know what they do, but it should be in the same file as the rest of the library code.

components/scroller/scroller.test.js
braddialpad: yeah kinda figured.. just so you don't forget

components/scroller/scroller_default.story.vue
iropolo: It not because I removed resize functionality and we are not using 'resize-observer' anymore

components/scroller/scroller.vue
iropolo: I would not like to have unnecessary or "magic" code in our component, if it is not necessary I would remove it.

If in the further feedback we found a bug or miss some functionality we can work it later.

components/scroller/scroller.vue
iropolo: ye... mhmm let me see what I can do...
At this moment our scroller.vue its their dynamic scroller from the library, and our core-scroller its their recycle-scroller.

And Im just triggering one or the other if they set `dynamic` true

components/scroller/scroller_default.story.vue
braddialpad: great, we can remove it then.

components/scroller/scroller.vue
braddialpad: > ye... mhmm let me see what I can do... At this moment our scroller.vue its their dynamic scroller from the library, and our core-scroller its their recycle-scroller.
> 
> And Im just triggering one or the other if they set `dynamic` true

The ideal way to do this would be to put the dynamic scroller code in it's own file within`modules` as well. 

Logic to decide which is triggered or not based on whether dynamic = true would remain in the scroller.vue

components/scroller/scroller.vue
braddialpad: Yep no problem, just making sure we know exactly what it does before removing.

components/scroller/scroller_default.story.vue
braddialpad: Alright, I'm thinking we set a "width" and "height" prop on `scroller.vue` that defaults to 100% since this is likely something that would be commonly set.

see [Dropdown](https://vue.dialpad.design/?path=/story/components-dropdown--default) props maxHeight and maxWidth for reference.

iropolo: mhmmm okay it will be sth like scrollerWidth and scrollerHeight, alright

components/scroller/scroller.vue
iropolo: Agree, let me rethink this and I will separate `dynamic` to another component in `modules`

components/scroller/scroller.stories.js
```diff
@@ -56,6 +56,18 @@ export const argTypesData = {
         'Is required for the initial render of items in DYNAMIC size mode.',
   },
 
+  scrollerWidth: {
+    control: { type: null },
+    description: 'The width of the scroller. ' +
+        'Can be a number (in pixels) or a string (in CSS units).',
```
braddialpad: You shouldn't have to repeat this in the description, storybook will automatically use your comment on the prop. I think the reason it is not working is because the syntax of your comment is not correct. It needs to be like so:

```js
/**
 * comment here
 */
```

note there is two * at the top

components/scroller/modules/scroller-item.vue
```diff
@@ -0,0 +1,232 @@
+<script>
```
braddialpad, line 5: Let's put a comment at the top stating that this is code from an external library and link to the library it's from. Also state that we have modified it for our own specific use. Put the same comment in the other library code files.

components/scroller/scroller.stories.js
```diff
@@ -0,0 +1,102 @@
+import DtScroller from './scroller';
+import BaseScrollerMdx from './scroller.mdx';
+import ScrollerDefault from './scroller_default.story';
+import { createTemplateFromVueFile } from '@/common/storybook_utils';
+export const argTypesData = {
```
braddialpad, line 8: add

```js
import { action } from '@storybook/addon-actions';

export const argsData = {
  onScrollStart: action('scroll-start'),
  onScrollEnd: action('scroll-end'),
};
```

pass it into the default export object with `args: argsData,`

components/scroller/scroller.stories.js
```diff
@@ -0,0 +1,102 @@
...
+  },
+
+  scrollerWidth: {
+    control: { type: null },
+    description: 'The width of the scroller. ' +
+        'Can be a number (in pixels) or a string (in CSS units).',
+  },
+
+  scrollerHeight: {
+    control: { type: null },
```
braddialpad: Why null controls for everything?

components/scroller/scroller.stories.js
```diff
@@ -0,0 +1,102 @@
+import DtScroller from './scroller';
+import BaseScrollerMdx from './scroller.mdx';
+import ScrollerDefault from './scroller_default.story';
+import { createTemplateFromVueFile } from '@/common/storybook_utils';
+export const argTypesData = {
+  default: {
+    control: { type: null },
+    description: 'Slot for main content',
```
braddialpad: It is not clear what you are supposed to put into the default slot.

components/scroller/scroller_default.story.vue
```diff
@@ -0,0 +1,123 @@
+<template>
+  <div>
+    <h3>Scroller</h3>
```
braddialpad, line 3: We should only put one scroller on the default page that can be changed via the props that are set in the controls. If we want to show more than one example with static props like you are doing here, make a `scroll_variants.story.vue` page.

braddialpad (commented):
This looks much better, nice work! I approve this draft and you can move on to the next phase.

components/scroller/scroller.stories.js
iropolo: Mhmm I added it but I cannot trigger the events on the `action` tab in storybook.. it can be because the event its emmited from `core-scroller` component and not on `scroller`?

components/scroller/scroller.stories.js
iropolo: ye it was in TO DO, but now let me build it, I will build also the variant (dynamic-size)

components/scroller/scroller_default.story.vue
iropolo: ye it was in TO DO, but now let me build it, I will build also the variant (dynamic-size)

components/scroller/scroller.stories.js
iropolo: Mhmm.. what about

```
description: 'Slot for main content that will be rendered inside the scroller.' +
      'Default scoped slot props are:\n\n ' +
      'item: item being rendered in a view.\n\n' +
      'index: reflects each items position in the items array\n\n' +
      'active: whether or not the view is active.',
```

components/scroller/scroller.test.js
```diff
@@ -0,0 +1,87 @@
+import { assert } from 'chai';
+import { mount } from '@vue/test-utils';
+import DtScroller from './scroller.vue';
+
+global.requestAnimationFrame = cb => cb();
```
iropolo, line 5: Mhmm.. I saw we use 

`global.requestAnimationFrame = sinon.spy();`

But it not work... Will investigate this

iropolo: I updated it with 

`  before(function () {
    global.requestAnimationFrame = sinon.spy();
  });
`
But as you can see, Im not calling 
`global.cancelAnimationFrame = sinon.spy();`
and also 
`config.global.renderStubDefaultSlot = true;`

They are necessary to call? or it is okay on this way?

components/scroller/scroller_dynamic.story.vue
```diff
@@ -0,0 +1,76 @@
...
+            class="image"
+          >
+        </div>
+        <div class="text">
+          {{ item.message }}
+        </div>
+      </template>
+    </dt-scroller>
+  </div>
+</template>
```
braddialpad, line 33: Bit of odd behaviour with this example. When you scroll the scrollbar with the mouse, the scrollbar does not stay in the position where your cursor is. Guessing this has something to do with the dynamic loading.

components/scroller/scroller.stories.js
braddialpad: definitely better, but it's not totally clear what some of these properties really mean.

item - this is the data for an item? like a javascript object containing properties for the item?
index - this is obvious so it's fine
active - what does the view being active mean? is this like enabled vs disabled?

A dev with zero knowledge of the library this is using should be able to understand how to use this just by reading it

components/scroller/scroller.test.js
braddialpad: unless you are getting errors when running the tests this is fine.

components/scroller/scroller.stories.js
```diff
@@ -0,0 +1,118 @@
...
+    control: { type: null },
+    description: 'Emitted when the last item is rendered.',
+  },
+
+};
+
+export const argsData = {
+  onScrollStart: action('scroll-start'),
+  onScrollEnd: action('scroll-end'),
+};
```
braddialpad: The way storybook works here is a little confusing, but normally when you have events you have to do them like this to get them to display in storybook properly. Can see some of our other existing files for examples.

```js
// Action Event Handlers
onScrollStart: {
  table: {
    disable: true,
  },
},
onScrollEnd: {
  table: {
    disable: true,
  },
},

'scroll-start': {
  description: 'Emitted when the first item is rendered.',
  table: {
    type: { summary: 'event' },
  },
},
'scroll-end': {
  description: 'Emitted when the last item is rendered.',
  table: {
    type: { summary: 'event' },
  },
},
```

components/scroller/scroller.stories.js
```diff
@@ -0,0 +1,118 @@
...
+
+export default {
+  title: 'Components/Scroller',
+  component: DtScroller,
+  parameters: {
+    docs: {
```
braddialpad, line 108: add to parameters object

```
controls: {
  sort: 'requiredFirst',
},
```

this sorts the props in storybook in our standardized order which is required first and then alphabetical.

components/scroller/scroller.stories.js
```diff
@@ -0,0 +1,118 @@
...
+};
+
+const Template = (args) => createTemplateFromVueFile(args, ScrollerDefault);
+
+export const Default = Template.bind({});
+Default.args = {
+  default: 'Scroller',
+};
+
+const VariantsTemplate = (args) => createTemplateFromVueFile(args, ScrollerDynamic);
```
braddialpad: Can just call this DynamicTemplate

braddialpad (commented):
Leaving a few comments I will review more thoroughly on monday.

components/scroller/scroller_dynamic.story.vue
iropolo: Same behaviour in library...  need to investigate it, I recommend to take it later in a different ticket

braddialpad: Yeah 100% not a blocker for this PR

components/scroller/scroller.stories.js
iropolo: Item is the item itself... it will be an object if you are using an array of objects
active return if the item is visible for the user or not


image

I could update it to sth like:

'item: the item itself, it will be an object if your list is an array with objects.\n\n' +
'index: reflects each items position in the list array\n\n' +
'active: return if the item is visible for the user or not.',

components/scroller/scroller.stories.js
iropolo: Nice

components/scroller/scroller.stories.js
iropolo: Agree

components/scroller/scroller.stories.js
braddialpad: > I could update it to sth like:
> 
> 'item: the item itself, it will be an object if your list is an array with objects.\n\n' + 'index: reflects each items position in the list array\n\n' + 'active: return if the item is visible for the user or not.',

Definitely the right idea. Just to refine the wording a bit..

we can use jsdoc for types:

### default slot

Markup to display each item from the items array.

**item {*}:** the item currently being rendered from within your `items` array.
**index {number}:** the index of this item in the `items` array.
**active {boolean}:** whether this item is currently visible to the user.

components/scroller/scroller.vue
```diff
@@ -0,0 +1,163 @@
...
+
+    /**
+      * The direction of the scroller.
+      * Can be either 'vertical' or 'horizontal'.
+     */
+    direction: {
+      type: String,
+      default: 'vertical',
+      validator: (value) => ['vertical', 'horizontal'].includes(value),
+    },
```
braddialpad: Our core components need to be jsdoc'd properly with @values when it is limited to specific values or it won't display correctly with the "Vue API" on the dialpad.design site.

example for this one
```
/**
* The direction of the scroller.
* @values vertical, horizontal
*/
```

components/scroller/scroller.vue
```diff
@@ -0,0 +1,163 @@
...
+      default: null,
+    },
+
+    /**
+      * Minimum size used if the height (or width in horizontal mode) of a item is unknown.
+      * Is required for the initial render of items in DYNAMIC size mode.
+     */
+    minItemSize: {
+      type: [Number, String],
+    },
```
braddialpad: We should validate this as required when dynamic is true.

See `validateProps ()` in `badge.vue` for a good example of how to do these validations involving multiple props.

components/scroller/scroller.vue
```diff
@@ -0,0 +1,163 @@
...
+
+    /**
+      * Display height (or width in horizontal mode) of the items in pixels
+      * used to calculate the scroll size and position.
+     */
+    itemSize: {
+      type: Number,
+      default: null,
+    },
```
braddialpad: We should validate this as required when dynamic is false.

See validateProps () in badge.vue for a good example of how to do these validations involving multiple props.

components/scroller/scroller.vue
```diff
@@ -0,0 +1,163 @@
...
+    /**
+      * Minimum size used if the height (or width in horizontal mode) of a item is unknown.
+      * Is required for the initial render of items in DYNAMIC size mode.
+     */
+    minItemSize: {
+      type: [Number, String],
+    },
+  },
+
+  emits: ['scroll-start', 'scroll-end'],
```
braddialpad: Core scroller emits this right? so we have to explicitly emit this in this component as well if we want the consumer to get it.

within the `<component>` template add

```
<component
  @scroll-start="$emit($event)"
  @scroll-end="$emit($event)"
>
```

remember that `v-bind="$attrs"` only works on native events, not custom ones.

components/scroller/scroller.vue
braddialpad: add to the description that this is required if dynamic is false

components/scroller/scroller.vue
```diff
@@ -0,0 +1,163 @@
...
+    scrollerHeight: {
+      type: [String, Number],
+      default: '100%',
+    },
+
+    /**
+      * Indicates if the items are dynamic.
+      * If true, the items will be wrapped in a DtScrollerItem component.
+      * This is required for dynamic items to be able to react to changes in their size.
+     */
```
braddialpad: The consumer doesn't need to know about these internal details (DtScrollerItem and such). Lets change this to add the details the consumer really cares about.

"Indicates if the items need to react to changes in their size. If disabled the itemSize prop is required and you will get improved performance. If enabled the minItemSize prop is required and you will have reduced performance but the ability to reactively size list items."

components/scroller/scroller_default.story.vue
```diff
@@ -0,0 +1,66 @@
...
+      :scroller-width="$attrs.scrollerWidth"
+      :min-item-size="$attrs.minItemSize"
+      :list-tag="$attrs.listTag"
+      :item-tag="$attrs.itemTag"
+      :direction="$attrs.direction"
+      key-field="id"
+    >
+      <template #default="{ active }">
+        <div class="user">
+          active: {{ active }}
```
braddialpad: Might still want to show the index here as well.

components/scroller/scroller.vue
```diff
@@ -0,0 +1,163 @@
+<template>
+  <component
```
braddialpad: Add `tabindex=0` as scrollable regions should always be focusable according to this

https://dequeuniversity.com/rules/axe/4.4/scrollable-region-focusable?application=axe-puppeteer

components/scroller/scroller_default.story.vue
```diff
@@ -0,0 +1,66 @@
...
+        </div>
+      </template>
```
braddialpad: hmm can we try an example using dt-list-item? I expect this is what would be used within the scroller the majority of the time.

components/scroller/scroller.vue
iropolo: Added @values true, false on `dynamic` prop
Added @values vertical, horizontal on `direction` prop

iropolo: Should we additionally add 
@type
@required
@default 
@example

Or @values when it is limited to specific values its enough?

Example on `Items` prop:

```
@type {Array}
@required
@default []
@example items: [ 'item1', 'item2', 'item3' ]
@example items: [ { id: 1, name: 'item1' }, { id: 2, name: 'item2' }, { id: 3, name: 'item3' } ]
```

components/scroller/scroller_default.story.vue
iropolo: Oh yes, I was testing the {active}, Its already updated with the previous template

components/scroller/scroller.vue
braddialpad: I believe we don't need type/required/default because we get it directly from the vue prop. Example is kinda nice though, I'll leave it up to you if you think it's necessary.

components/scroller/scroller.stories.js
braddialpad: see https://github.com/dialpad/dialtone-vue/pull/780#discussion_r1134544113 I think that will fix your issue here.

components/scroller/scroller_default.story.vue
iropolo: Mhmm I dont see necessary to wrap a bunch of code if there is no special usage.
dt-list-item is used on dynamic-size to handle size computations of the element.

There is not necessary to use it if you just want to use a simple list.

braddialpad: what about keyboard navigation?

components/scroller/scroller_dynamic.story.vue
braddialpad: After looking into it a bit, I think this is an issue that is not possible to fix. Even if you scroll up the chat in slack it has this same issue. The scrollbar itself is going to change if the items inside it are changing, not much we can do about it. 

Don't worry about this one.

components/scroller/scroller_default.story.vue
iropolo: We dont have any keyboard navigation here... either in the library...
Im guessing because there is no way you want to navigate between 1000 items with keyboard...

But... it is possible for the developer to build a manual keyboard navigation.
For example: 

```
switch (e.key) {
      focusedItemIndex.value++;
      break;
    case 'ArrowUp':
      focusedItemIndex.value--;
      break;
    case 'Home':
      focusedItemIndex.value = 0;
      break;
    case 'End':
      focusedItemIndex.value = props.items.length - 1;
      break;
    case 'Escape':
      opened.value = false;
      focusedItemIndex.value = null;
      break;
    case 'Enter':
      onSelect();
      break;
  }
```

iropolo: I dont exactly why you ask for keyboard navigation on dtItemList discussion, do you have sth in mind?

components/scroller/scroller.vue
iropolo: image

Done... also with this you can navigate with keyboard (the scrollable content, not specific item)

braddialpad: good enough for this MVP, thanks!

components/scroller/scroller_default.story.vue
braddialpad: We have an existing mixin that works with our dt-list-items. Anyways it's not really part of the requirements for this MVP so I don't think we need to worry about it for now, though I know they will want it in the future.

components/scroller/scroller.vue
iropolo: Mhmm....
This emits are from scroller.vue, Im injecting it in core-scroller (its fired from core-scroller but the emit its in scroller.vue)

Previously it was on core-scroller, but I updated it on this way to expose it in OUR component and because it was more easily to test it.

You can check it if you listen it in scroller_default
Also on Vue tools

image

iropolo: Or maybe I'm not understanding correct what you mean.

components/scroller/scroller.stories.js
iropolo: Still not work, we can follow the discussion there and resolve this conflict to void multithread

https://github.com/dialpad/dialtone-vue/pull/780#discussion_r1134701244

components/scroller/scroller.stories.js
```diff
@@ -0,0 +1,134 @@
...
+import BaseScrollerMdx from './scroller.mdx';
+import ScrollerDefault from './scroller_default.story';
+import ScrollerDynamic from '@/components/scroller/scroller_dynamic.story.vue';
+import { createTemplateFromVueFile } from '@/common/storybook_utils';
+import { action } from '@storybook/addon-actions';
+
+export const argTypesData = {
+  default: {
+    control: { type: null },
+    description: 'Markup to display each item from the items array.' +
```
braddialpad: missing `\n` here

components/scroller/scroller.vue
braddialpad: Just that it's not triggering an event in storybook, I will check this out.

components/scroller/scroller_default.story.vue
```diff
@@ -0,0 +1,66 @@
...
+      class="scroller"
+      :items="items"
+      :item-size="$attrs.itemSize"
+      :scroller-height="$attrs.scrollerHeight"
+      :scroller-width="$attrs.scrollerWidth"
+      :min-item-size="$attrs.minItemSize"
+      :list-tag="$attrs.listTag"
+      :item-tag="$attrs.itemTag"
+      :direction="$attrs.direction"
+      key-field="id"
```
braddialpad: No reason to not just use `$attrs.keyField` here

components/scroller/scroller.vue
```diff
@@ -0,0 +1,186 @@
...
+      type: Number,
+      default: null,
+    },
+
+    /**
+      * Minimum size used if the height (or width in horizontal mode) of a item is unknown.
+      * Is required for the initial render of items in DYNAMIC size mode.
+     */
+    minItemSize: {
+      type: [Number, String],
```
braddialpad: Requires default prop for the linter (can just be null or undefined)

components/scroller/scroller.vue
braddialpad: See https://github.com/dialpad/dialtone-vue/pull/812

components/scroller/scroller.vue
braddialpad, line 1: Lets not change this now since we're just about ready to release this, but shouldn't this component be in composition API?

components/scroller/scroller.vue
iropolo: Great.
It was necessary to put on the story
`@scroll-start="$attrs.onScrollStart"
@scroll-end="$attrs.onScrollEnd"`

To work properly

components/scroller/scroller.vue
iropolo: Yes let me update it

components/scroller/scroller.vue
braddialpad: We should still remove the provide emit thing we've got going on here. That's not a normal way to emit an event.

iropolo: Yes, now the test works with this addition on scroller.vue
`
@scroll-start="$emit('scroll-start')"
@scroll-end="$emit('scroll-end')"
`

braddialpad: nice, looks good.

components/scroller/index.js
```diff
@@ -0,0 +1 @@
+export { default as DtScroller } from './DtScroller.vue';
```
braddialpad, line 1: Just to be consistent with the naming conventions of our other components, the file name should be 'Scroller.vue' but the export should be DtScroller

braddialpad (approved):
Lookin good, nice work!

components/scroller/index.js
braddialpad: Resolving after discussion due to issues with composition API and component names.

PR #786, opened 2023-03-01
feat(leftbar-general-row): additional styles

braddialpad:
# feat(leftbar-general-row): additional styles

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

There were some styles needed that we only just discovered after looking into the product code. Added two new types for "digital engagement" and "assigned". Added prop for a "muted" channel which reduces opacity. Added prop for "active voice chat" which displays an animated icon when true.

Also updated CC colors to match the available CC colors in product

Just used the styling that was already existing in product for muted and active voice chat. Let me know if updated styling is needed.

## :bulb: Context

Adds styles that were needed in product leftbar.

## :pencil: Checklist

- [x] I have added tests

## :crystal_ball: Next Steps

- Integrate into product
- some changes needed to group row as well.

### DIFF
--- a/recipes/leftbar/general_row/general_row.vue
+++ b/recipes/leftbar/general_row/general_row.vue
@@ -29,11 +29,15 @@
         </dt-emoji-text-wrapper>
       </div>
       <div class="dt-leftbar-row__omega">
-        <dt-icon
+        <div
           v-if="activeVoiceChat"
-          size="300"
-          name="activity"
-        />
+          class="dt-leftbar-row__active-voice"
+        >
+          <dt-icon
+            size="300"
+            name="activity"
+          />
+        </div>
         <dt-badge
           v-else-if="hasUnreadMessages"
           kind="count"
--- a/recipes/leftbar/general_row/general_row.vue
+++ b/recipes/leftbar/general_row/general_row.vue
@@ -30,7 +30,13 @@
       </div>
       <div class="dt-leftbar-row__omega">
         <div
-          v-if="activeVoiceChat"
+          v-if="dndText"
+          class="dt-leftbar-row__dnd"
+        >
+          {{ dndText }}
+        </div>
+        <div
+          v-else-if="activeVoiceChat"
           class="dt-leftbar-row__active-voice"
         >
           <dt-icon
@@ -136,6 +142,15 @@ export default {
     activeVoiceChat: {
       type: Boolean,
       default: false,
+    },
+
+    /**
+     * Acronym used to represent "Do not Disturb" state. If entered will display the entered text over
+     * unreadCount and activeVoiceChat.
+     */
+    dndText: {
+      type: String,
+      default: '',
     },
   },
 
--- a/recipes/leftbar/general_row/general_row.vue
+++ b/recipes/leftbar/general_row/general_row.vue
@@ -29,8 +29,13 @@
         </dt-emoji-text-wrapper>
       </div>
       <div class="dt-leftbar-row__omega">
+        <dt-icon
+          v-if="activeVoiceChat"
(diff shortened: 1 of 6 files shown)


recipes/leftbar/style/leftbar_row.less
```diff
@@ -210,3 +225,35 @@
     height: calc(var(--size-300) * 3.5);
   }
 }
+
+.opacity-pulsate {
```
francisrupert, line 243: Great substitution. I've logged [DT-1008](https://dialpad.atlassian.net/browse/DT-1008) to refine it later.

recipes/leftbar/general_row/general_row.vue
```diff
@@ -29,8 +29,13 @@
         </dt-emoji-text-wrapper>
       </div>
       <div class="dt-leftbar-row__omega">
+        <dt-icon
+          v-if="activeVoiceChat"
+          size="300"
+          name="activity"
+        />
```
francisrupert: In case we go with another icon than `activity` (or even to future-proof it in case we don't use an icon), wrap the `<dt-icon>` with a `class="dt-leftbar-row__active-voice"` and use that to style it (TBD via [DT-1008](https://dialpad.atlassian.net/browse/DT-1008)) (see other review in `leftbar_row.less` below).

recipes/leftbar/style/leftbar_row.less
```diff
@@ -158,6 +167,11 @@
     padding-left: var(--space-400);
     padding-right: var(--space-400);
     visibility: var(--leftbar-row-omega-visibility);
+
+    .d-icon--activity {
+      color: var(--green-400);
+      .opacity-pulsate();
+    }
```
francisrupert: Rather than make it specific to `d-icon--activity`, create a new style, `.dt-leftbar-row__active-voice`. So that'd be...

```less
&__active-voice {
  color: var(--fc-success);
  .opacity-pulsate();
}
```

recipes/leftbar/general_row/general_row_default.story.vue
```diff
@@ -5,6 +5,8 @@
     :description="description"
     :unread-count="unreadCount"
     :selected="selected"
+    :muted="muted"
```
francisrupert, line 8: You mentioned earlier it would remain a `<button>`, and can't remove its actionability. I'm _assuming_ it's not actionable though. Conditionally add `disabled` attribute?

Naturally we'd prefer it not just _look_ disabled, but functionally be disabled.

francisrupert: And if you _can_ toggle the `disabled` attribute, then you would update...

`.dt-leftbar-row__primary:hover` 
to
`.dt-leftbar-row__primary:not([disabled]):hover`

...and...

`.dt-leftbar-row__primary:active`
to
`.dt-leftbar-row__primary:not([disabled]):active`.

Oh, and add...

`.dt-leftbar-row__primary:disabled { cursor: default; }` so it's not the button's default pointer cursor.

francisrupert (commented):
Really good edge case catches.

recipes/leftbar/general_row/general_row_default.story.vue
braddialpad: Resolving this post our discussion in slack

braddialpad:
Requested changes made.

francisrupert (approved):
Nailed it.

braddialpad:
Added a dnd prop as well using the current styling for DND in firespotter. More of a placeholder until we've figured out what we want to do with it.

PR #796, opened 2023-03-03
feat(leftbar-contact-row): add contact row

braddialpad:
# feat(leftbar-contact-row): add contact row

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Leftbar contact row

Still some work to do on polishing up the documentation and writing tests and accessibility testing, but the component is ready for design review.

## :bulb: Context

Leftbar row item for a single contact

## :pencil: Checklist

- [ ] I have added tests

## :crystal_ball: Next Steps

Product integration

## :link: Sources

Francis' original codepen https://codepen.io/francisrupert/pen/QWVNvGX

### DIFF
--- a/recipes/leftbar/contact_row/contact_row.stories.js
+++ b/recipes/leftbar/contact_row/contact_row.stories.js
@@ -25,7 +25,6 @@ export const argTypesData = {
   },
 
   avatarPresence: {
-    defaultValue: null,
     control: {
       type: 'select',
       options: [...PRESENCE_STATES_LIST],
--- a/recipes/leftbar/contact_row/contact_row.stories.js
+++ b/recipes/leftbar/contact_row/contact_row.stories.js
@@ -4,6 +4,7 @@ import DtRecipeContactRow from './contact_row';
 import DtRecipeContactRowMdx from './contact_row.mdx';
 import DtRecipeContactRowDefaultTemplate from './contact_row_default.story.vue';
 import DtRecipeContactRowVariantsTemplate from './contact_row_variants.story.vue';
+import { PRESENCE_STATES_LIST } from '@/components/presence/presence_constants';
 const defaultImage = require('./../../../components/avatar/person.png');
 
 // Default Prop Values
@@ -21,6 +22,14 @@ export const argTypesData = {
 
   avatarSrc: {
     defaultValue: defaultImage,
+  },
+
+  avatarPresence: {
+    defaultValue: null,
+    control: {
+      type: 'select',
+      options: [...PRESENCE_STATES_LIST],
+    },
   },
 
   // Slots
--- a/recipes/leftbar/contact_row/contact_row.stories.js
+++ b/recipes/leftbar/contact_row/contact_row.stories.js
@@ -101,7 +101,6 @@ const VariantsTemplate = (args, { argTypes }) => createTemplateFromVueFile(
 
 // Stories
 export const Default = DefaultTemplate.bind({});
-Default.args = {};
 
 export const Variants = VariantsTemplate.bind({});
 Variants.args = {};
--- a/recipes/leftbar/contact_row/contact_row.stories.js
+++ b/recipes/leftbar/contact_row/contact_row.stories.js
@@ -1,28 +1,22 @@
 import { action } from '@storybook/addon-actions';
 import { createTemplateFromVueFile } from '@/common/storybook_utils';
-import DtRecipeGroupRow from './group_row';
-import DtRecipeGroupRowMdx from './group_row.mdx';
-import DtRecipeGroupRowDefaultTemplate from './group_row_default.story.vue';
-import DtRecipeGroupRowVariantsTemplate from './group_row_variants.story.vue';
+import DtRecipeContactRow from './contact_row';
+import DtRecipeContactRowMdx from './contact_row.mdx';
+import DtRecipeContactRowDefaultTemplate from './contact_row_default.story.vue';
+import DtRecipeContactRowVariantsTemplate from './contact_row_variants.story.vue';
 const defaultImage = require('./../../../components/avatar/person.png');
 
 // Default Prop Values
 export const argsData = {
   onClick: action('click'),
+  onCall: action('call'),
 };
 
 export const argTypesData = {
   // Props
-  groupCount: {
-    defaultValue: 8,
-  },
 
-  names: {
-    defaultValue: 'Jaqueline Nackos, Lori Smith',
-  },
-
-  avatarInitials: {
-    defaultValue: 'JN',
+  name: {
+    defaultValue: 'Jaqueline Nackos',
   },
 
   avatarSrc: {
@@ -38,8 +32,21 @@ export const argTypesData = {
     },
   },
 
+  onCall: {
+    table: {
+      disable: true,
+    },
+  },
+
   click: {
-    description: 'Native click event on the row',
+    description: 'Native click event',
+    table: {
+      type: { summary: 'event' },
+    },
+  },
+
+  call: {
+    description: 'Call button clicked',
     table: {
       type: { summary: 'event' },
     },
@@ -52,8 +59,8 @@ const decorator = () => ({
 
 // Story Collection
 export default {
-  title: 'Recipes/Leftbar/Group Row',
-  component: DtRecipeGroupRow,
+  title: 'Recipes/Leftbar/Contact Row',
+  component: DtRecipeContactRow,
   args: argsData,
   argTypes: argTypesData,
   decorators: [decorator],
@@ -63,7 +70,7 @@ export default {
       sort: 'requiredFirst',
     },
     docs: {
-      page: DtRecipeGroupRowMdx,
+      page: DtRecipeContactRowMdx,
     },
     options: {
       showPanel: true,
@@ -75,12 +82,12 @@ export default {
 const DefaultTemplate = (args, { argTypes }) => createTemplateFromVueFile(
   args,
   argTypes,
-  DtRecipeGroupRowDefaultTemplate,
+  DtRecipeContactRowDefaultTemplate,
 );
 const VariantsTemplate = (args, { argTypes }) => createTemplateFromVueFile(
   args,
   argTypes,
-  DtRecipeGroupRowVariantsTemplate,
+  DtRecipeContactRowVariantsTemplate,
 );
 
 // Stories
--- a/recipes/leftbar/contact_row/contact_row.stories.js
+++ b/recipes/leftbar/contact_row/contact_row.stories.js
@@ -47,14 +47,12 @@ export const argTypesData = {
   },
 
   click: {
-    description: 'Native click event',
     table: {
       type: { summary: 'event' },
     },
   },
 
   call: {
-    description: 'Call button clicked',
     table: {
       type: { summary: 'event' },
     },
--- a/recipes/leftbar/contact_row/contact_row.stories.js
+++ b/recipes/leftbar/contact_row/contact_row.stories.js
@@ -5,7 +5,7 @@ import DtRecipeContactRowMdx from './contact_row.mdx';
 import DtRecipeContactRowDefaultTemplate from './contact_row_default.story.vue';
 import DtRecipeContactRowVariantsTemplate from './contact_row_variants.story.vue';
 import { PRESENCE_STATES_LIST } from '@/components/presence/presence_constants';
-const defaultImage = require('./../../../components/avatar/person.png');
+const defaultImage = require('@/components/avatar/person.png');
 
 // Default Prop Values
 export const argsData = {
@@ -30,8 +30,6 @@ export const argTypesData = {
       options: [...PRESENCE_STATES_LIST],
     },
   },
-
-  // Slots
 
   // Action Event Handlers
   onClick: {
--- a/recipes/leftbar/contact_row/contact_row.vue
+++ b/recipes/leftbar/contact_row/contact_row.vue
@@ -58,6 +58,8 @@ import { DtRecipeGeneralRow } from '@/recipes/leftbar/general_row';
 import DtEmojiTextWrapper from '@/components/emoji_text_wrapper/emoji_text_wrapper.vue';
 import DtAvatar from '@/components/avatar/avatar.vue';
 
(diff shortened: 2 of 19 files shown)


recipes/leftbar/contact_row/contact_row.vue
```diff
@@ -0,0 +1,163 @@
...
+        case 'active':
+          return 'var(--theme-presence-color-background-available)';
+        case 'busy':
+          return 'var(--theme-presence-color-background-busy-unavailable)';
+        case 'away':
+          return 'var(--theme-presence-color-background-busy)';
+        default:
+          return undefined;
+      }
+    },
```
francisrupert: @bhawkins-dialpad have some insight here? This is for the "Context" string of content.

I couldn't find specific documentation on this beyond the Status section of [DP 2.0 Left sidebar](https://www.figma.com/file/6n4RCYk2y2rYx5c4hL088f/DP-2.0-Left-sidebar?node-id=1%3A13438&t=FLFTQw3BurZgb54A-11).

That had the "Away" text color as the old Orange 500, which effectively used to be the "Warning" color, and it doesn't specifically match the Presence's "Away" dot color.

My instinct is the Context text string should use `class="d-fc-critical | d-fc-success | d-fc-warning"` and the theme presence colors are not in play for them. Too risky for a theme to have a really light Presence dot color that won't hold up for a11y contrast; as it is, the theme's "away" color fails miserably.

image

recipes/leftbar/general_row/general_row.vue
```diff
@@ -53,6 +55,27 @@
...
+          @click.stop="$emit('call', $event)"
+        >
+          <template #icon>
+            <dt-icon
+              name="phone"
+              size="200"
+            />
+          </template>
+        </dt-button>
+      </div>
```
francisrupert: `.dt-leftbar-row__action` is incorrectly nested within `.dt-leftbar-row__primary` (`<button>` within a `<button>`), must sit at same level as `.dt-leftbar-row__primary`. 

Before/after screenshot of what that should look like.

image

francisrupert: **Hiding `.dt-leftbar-row__omega` on hover and focus:**

https://user-images.githubusercontent.com/1165933/222622396-362d5521-0d21-4ac4-bb9e-0ed6980abc3c.mov

When I move that structure to the correct location, hover works as intended via CSS, though a script will be necessary to do the same when keyboard-focusing the button. Yesterday I added that AC point in [DT-959](https://dialpad.atlassian.net/browse/DT-959), and pointed to my codepen prototype JS.

recipes/leftbar/contact_row/contact_row.vue
```diff
@@ -0,0 +1,163 @@
...
+        {{ name }}
+      </dt-emoji-text-wrapper>
+      <div class="dt-leftbar-row__status">
+        <span
+          data-qa="dt-leftbar-row-presence-text"
+          class="dt-leftbar-row__meta-context"
+          :style="{ color: presenceColor }"
+        >
+          {{ presenceText }}
+        </span>
```
francisrupert: Add `v-if="presenceText`

Suppress this outright if there's no `presenceText` string. That is, keeping...

```html
<span class="dt-leftbar-row__meta-context"></span>
```
...in there when blank results in the `•` rendering, which must only appear if _both_ `.dt-leftbar-row__meta-context` and `dt-leftbar-row__meta-custom.` are present.

image

recipes/leftbar/contact_row/contact_row.vue
```diff
@@ -0,0 +1,163 @@
...
+        <dt-emoji-text-wrapper
+          v-if="userStatus"
+          element-type="span"
+          data-qa="dt-leftbar-row-user-status"
+          class="dt-leftbar-row__meta-custom"
+        >
+          {{ userStatus }}
+        </dt-emoji-text-wrapper>
```
francisrupert: Umm..?


image

image

recipes/leftbar/general_row/general_row.vue
```diff
@@ -160,14 +185,22 @@ export default {
...
       default: '',
     },
+
+    /**
+     * Whether the row should have a call button. Usually only applicable to individual contact rows.
+     */
+    hasCallButton: {
+      type: Boolean,
+      default: false,
+    },
```
francisrupert, line 201: Still present when set to `false`.

francisrupert (commented):
Gooood stuff man. Apologies if I pointed out anything you're still working on.

francisrupert:
image

I've PR'd (https://github.com/dialpad/dialtone/pull/839) into `dialtone` to correct this border color to match selected state background.

francisrupert:
https://user-images.githubusercontent.com/1165933/222634158-f8d6f847-01f1-420d-ab63-3e00da973993.mov

I think I transposed the `:hover` and `:active` bg color in [Theme vars](https://dialpad.design/design/colors/#theme). I'll resolve Friday.

bhawkins-dialpad:
We do not need to display the "Active/Available" presence text we only need to display text for  for "Away", "In a meeting","DND", etc..
image

bhawkins-dialpad:
The row hover background color `--theme-sidebar-row-color-background-hover` looks darker than the design it looks like its currently using `--black-300` when it should be using `--black-200`. Maybe the pressed/active and the hover states got swapped some how? The pressed/active state should be using the darker `--black-300` style. The selected state background color is correct.
image

bhawkins-dialpad:
The first letter of the presence text should be capitalized.
image

francisrupert:
> Maybe the pressed/active and the hover states got swapped some how?

@bhawkins-dialpad 

Yeah, that's on me - https://github.com/dialpad/dialtone-vue/pull/796#issuecomment-1452969785. It's a Dialtone CSS thing for I'll PR in at some point. No action for @braddialpad on that one.

braddialpad:
> We do not need to display the "Active/Available" presence text we only need to display text for for "Away", "In a meeting","DND", etc.. image

The consumer can pass in whatever text they want here, but I'll change the example.

recipes/leftbar/general_row/general_row.vue
braddialpad: Thanks, both these issues should be fixed now.

recipes/leftbar/contact_row/contact_row.vue
braddialpad: got it, should be fixed.

recipes/leftbar/contact_row/contact_row.vue
braddialpad: lol, added a size prop to our emoji wrapper. Should work well enough for now.

recipes/leftbar/general_row/general_row.vue
braddialpad: fixed.

braddialpad:
> > We do not need to display the "Active/Available" presence text we only need to display text for for "Away", "In a meeting","DND", etc.. image
> 
> The consumer can pass in whatever text they want here, but I'll change the example.

Fixed now.

recipes/leftbar/contact_row/contact_row.vue
francisrupert: Confirmed between @bhawkins-dialpad and I. Go with switching between `class="d-fc-critical | d-fc-success | d-fc-warning"`.

iropolo:
Phone icon persist after click it, should it disappear after unfocus?

image

recipes/leftbar/general_row/general_row.vue
francisrupert: The focus behavior is _almost_ there. I can see that you're removing the `class="dt-leftbar-row__omega"` from the DOM, but it needs to remain in there to hold the space beneath the call button which allows for the ellipsis to be properly positioned. 

Look at the [Codepen's](https://codepen.io/francisrupert/pen/QWVNvGX) `actionButtons.forEach()`. On focus/blur of the action button, toggle `class="dt-leftbar-row--action-focused"` on the `class="dt-leftbar-row"`.

**Current**

https://user-images.githubusercontent.com/1165933/222808902-7cad633a-ddac-4e31-b950-6e6e2bf4c0a1.mov


**Expected**

https://user-images.githubusercontent.com/1165933/222809888-82b30634-8d99-4251-a2a6-660a4fd3f51f.mov

braddialpad:
> Phone icon persist after click it, should it disappear after unfocus?
> 
> image

Hmm good find. The problem is that it in fact IS still focused because it was clicked. I will also hide on mouseout event that should fix it.

braddialpad:
> > Phone icon persist after click it, should it disappear after unfocus?
> > image
> 
> Hmm good find. The problem is that it in fact IS still focused because it was clicked. I will also hide on mouseout event that should fix it.

Should be good now.

recipes/leftbar/general_row/general_row.vue
braddialpad: Got it. Should be good now. I also added `--leftbar-row-omega-min-width: var(--leftbar-row-alpha-width);` to `&--action-focused` as the width was changing on hover, but not on focus.

recipes/leftbar/contact_row/contact_row.vue
braddialpad: Colors modified.

recipes/leftbar/contact_row/contact_row.stories.js
```diff
@@ -0,0 +1,106 @@
...
+  },
+
+  avatarPresence: {
+    control: {
+      type: 'select',
+      options: [...PRESENCE_STATES_LIST],
+    },
+  },
+
+  // Slots
```
iropolo: Seems not be necessary

recipes/leftbar/contact_row/contact_row.stories.js
```diff
@@ -0,0 +1,106 @@
+import { action } from '@storybook/addon-actions';
+import { createTemplateFromVueFile } from '@/common/storybook_utils';
+import DtRecipeContactRow from './contact_row';
+import DtRecipeContactRowMdx from './contact_row.mdx';
+import DtRecipeContactRowDefaultTemplate from './contact_row_default.story.vue';
+import DtRecipeContactRowVariantsTemplate from './contact_row_variants.story.vue';
+import { PRESENCE_STATES_LIST } from '@/components/presence/presence_constants';
+const defaultImage = require('./../../../components/avatar/person.png');
```
iropolo: Could be replaced with 

const defaultImage = require('@/components/avatar/person.png');

recipes/leftbar/contact_row/contact_row_constants.js
```diff
@@ -0,0 +1,5 @@
+export const DEFAULT_CONSTANTS = null;
```
iropolo: Where we are using DEFAULT_CONSTANTS?

recipes/leftbar/contact_row/contact_row_variants.story.vue
```diff
@@ -0,0 +1,89 @@
...
+        />
+      </div>
+    </dt-stack>
+  </div>
+</template>
+
+<script>
+import DtRecipeContactRow from './contact_row';
+import { DtStack } from '@/components/stack';
+const defaultImage = require('./../../../components/avatar/person.png');
```
iropolo: Could be replaced with 

const defaultImage = require('@/components/avatar/person.png');

iropolo: I can see this before the image loads correctly (dont mind the dot status, im checking on localhost and I dont have the latest dialtone css updated)
image

Final load:
image


but on the pr preview its seems to load the fallback image, cool

image


At the end its works correctly on both, you can test it with incognito mode and slow 3g on network tab

recipes/leftbar/contact_row/index.js
```diff
@@ -0,0 +1,2 @@
+export { default as DtRecipeContactRow } from './contact_row.vue';
+export {} from './contact_row_constants';
```
iropolo: Same on the previous comment, if we are not using DEFAULT_CONSTANTS this will not be necessary

recipes/leftbar/contact_row/contact_row_variants.story.vue
francisrupert: Eek, I have zero interest right now in viewing the DP app with incognito mode and slow 3g.

recipes/leftbar/general_row/general_row.vue
francisrupert: Perfect!

recipes/leftbar/contact_row/contact_row.vue
francisrupert: Yep, perfect. Just to mention another reason we must use these: there will be dark mode versions of `d-fc-critical, d-fc-success, d-fc-warning`, so there's that. Anyways.

francisrupert:
Great updates @braddialpad. I approve at least from UI perspective, but will refrain from marking it approved given the remaining Vue-specific feedback from @iropolo.

I still have to reconcile the hover/active/selected bg colors, https://github.com/dialpad/dialtone-vue/pull/796#issuecomment-1452969785, but that doesn't block this task.

recipes/leftbar/contact_row/contact_row_constants.js
braddialpad: We are not it just gets created by the generator. Removed the export and the const. Will leave the file there in case any constants need to be added in the future.

recipes/leftbar/contact_row/contact_row_variants.story.vue
braddialpad: hmm yeah seems like a local thing

recipes/leftbar/contact_row/contact_row.stories.js
braddialpad: yep thanks

recipes/leftbar/contact_row/contact_row.stories.js
braddialpad: removed

recipes/leftbar/contact_row/index.js
braddialpad: removed.

iropolo (approved):
Nice work here 👍

PR #797, opened 2023-03-03
fix(popover): protect against an undefined focus element

braddialpad:
# fix(popover): protect against an undefined focus element

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Fix for error mentioned in
https://dialpad.atlassian.net/browse/DT-1013

Do not cause error if focus element no longer exists.

## :bulb: Context

If navigating to a new page after clicking something in a popover it will error out because it tries to focus an element that no longer exists. This will cause an error to no longer happen in that case.

## :pencil: Checklist

## :crystal_ball: Next Steps

Verify with Yusef the fix works.


iropolo (approved):
LGTM

PR #803, opened 2023-03-08
fix(avatar): initials are not available until image loads

juliodialpad:
# Fix (Avatar): Initials are not available until image loads

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Changed the behavior to have the initials always displayed and put the image over so when the image is loading, initials are shown.

Fixed behavior on storybook, vue 2 version didn't detected the slot changes properly.

## :bulb: Context

Even after passing the initials, initials are not shown untill the image loads, this is replicable on storybook.

## :pencil: Checklist

- [x] I have added tests

Uploading Screen Recording 2023-03-07 at 9.10.59 p.m..mov…

### DIFF
--- a/components/avatar/avatar.vue
+++ b/components/avatar/avatar.vue
@@ -12,7 +12,7 @@
       <!-- @slot Slot for avatar content -->
       <slot v-if="showDefaultSlot" />
       <span
-        v-else-if="showInitials"
+        class="d-ps-absolute d-zi-base"
         :class="AVATAR_KIND_MODIFIERS.initials"
       >
         {{ formattedInitials }}
@@ -155,6 +155,7 @@ export default {
       imageLoadedSuccessfully: null,
       slottedInitials: '',
       formattedInitials: '',
+      defaultSlotContent: null,
     };
   },
 
@@ -172,7 +173,7 @@ export default {
     },
 
     showDefaultSlot () {
-      return this.kind !== 'initials' && this.imageLoadedSuccessfully !== false;
+      return this.kind !== 'initials' || this.imageLoadedSuccessfully === true;
     },
 
     showInitials () {
@@ -203,20 +204,26 @@ export default {
     },
   },
 
-  mounted () {
-    this.init();
+  watch: {
+    defaultSlotContent: {
+      deep: true,
+      immediate: true,
+      handler: function () {
+        this.init();
+      },
+    },
   },
 
   updated () {
-    if (this.kind === 'initials') {
-      this.slottedInitials = this.$slots.default[0].text || this.$slots.default[0].textContent;
-      this.formatInitials(this.slottedInitials);
-    }
+    this.defaultSlotContent = this.$slots.default ? this.$slots.default[0] : null;
+    this.imageLoadedSuccessfully = this.defaultSlotContent?.elm?.complete;
   },
 
   methods: {
     init () {
-      const firstChild = this.$refs.canvas.firstChild;
+      this.defaultSlotContent = this.$slots.default ? this.$slots.default[0] : null;
+      const firstChild = this.defaultSlotContent?.elm ? this.defaultSlotContent.elm : this.defaultSlotContent;
+      this.formatInitials(this.initials);
 
       if (!firstChild) {
         return;
@@ -227,14 +234,17 @@ export default {
     },
 
     kindHandler (el) {
+      if (!this.$el) return;
+      const avatarImage = this.$el.querySelector('img');
+      const iconElement = this.$el.querySelector('svg');
       switch (this.kind) {
         case 'image':
-          el.classList.add('d-avatar__image');
+          avatarImage?.classList.add('d-avatar__image', 'd-zi-base1');
           this.validateImageAttrsPresence();
           this.setImageListeners(el);
           break;
         case 'icon':
-          el.classList.add(AVATAR_KIND_MODIFIERS.icon);
+          iconElement?.classList.add(AVATAR_KIND_MODIFIERS.icon);
           break;
         case 'initials':
           this.slottedInitials = el.text || el.textContent;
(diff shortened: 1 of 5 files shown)


components/avatar/avatar.vue
```diff
@@ -203,20 +204,26 @@ export default {
...
+    this.defaultSlotContent = this.$slots.default ? this.$slots.default[0] : null;
+    this.imageLoadedSuccessfully = this.defaultSlotContent?.elm?.complete;
   },
 
   methods: {
     init () {
-      const firstChild = this.$refs.canvas.firstChild;
+      this.defaultSlotContent = this.$slots.default ? this.$slots.default[0] : null;
+      const firstChild = this.defaultSlotContent?.elm ? this.defaultSlotContent.elm : this.defaultSlotContent;
+      this.formatInitials(this.initials);
```
braddialpad: Why was the original firstChild method with $refs.canvas not working here? seems a bit like a more complicated way of doing the same thing, but I'm sure you had a reason for it.

components/avatar/avatar.vue
```diff
@@ -12,7 +12,7 @@
       <!-- @slot Slot for avatar content -->
       <slot v-if="showDefaultSlot" />
       <span
-        v-else-if="showInitials"
+        class="d-ps-absolute d-zi-base"
```
braddialpad, line 16: I think zi-base may cause a problem if we render within something that doesn't have a default z-index, like a popover for example. Let's try this to make sure.

components/avatar/avatar.vue
```diff
@@ -203,20 +204,26 @@ export default {
...
+      },
+    },
   },
 
   updated () {
-    if (this.kind === 'initials') {
-      this.slottedInitials = this.$slots.default[0].text || this.$slots.default[0].textContent;
-      this.formatInitials(this.slottedInitials);
-    }
+    this.defaultSlotContent = this.$slots.default ? this.$slots.default[0] : null;
```
iropolo: Would replace it with `this.defaultSlotContent = this.$slots.default?.[0] ? this.$slots.default[0] : null;`

iropolo: Mhmmm... `this.defaultSlotContent = this.$slots.default?.[0] ?? null;` is even better

braddialpad (commented):
This component is starting to feel a bit overcomplicated, though I don't think we can fix some of the issues without changing the public interface and making a breaking change. Was thinking it may simplify things if we used props instead of slots for some of this stuff since accessing the markup within the slots is fairly cumbersome and has reactivity issues.

components/avatar/avatar.vue
```diff
@@ -308,8 +324,8 @@ export default {
     },
 
     validateImageAttrsPresence () {
-      const isSrcMissing = !this.$refs.canvas.firstChild.getAttribute('src');
-      const isAltMissing = !this.$refs.canvas.firstChild.getAttribute('alt');
+      const isSrcMissing = !this.$slots.default[0].elm.getAttribute('src');
+      const isAltMissing = !this.$slots.default[0].elm.getAttribute('alt');
```
iropolo: Mhm... If we are using our local `defaultSlotContent` and also updating it in `updated()` hook... why we are accessing directly to the $slots here?, could we use `defaultSlotContent` ...? Or Im missing sth...

Just asking I didnt switch the branch

components/avatar/avatar.vue
juliodialpad: This wasn't working correctly because the $ref.canvas content was changing with the 'kind' property, sometimes overriding the checks I was doing later so the component didn't detected the changes properly.

This way, we're checking the actual default slot content property to decide what we should render.

components/avatar/avatar.vue
juliodialpad: Sure, I was thinking the same, I just wanted to have the image over the initials because the initials are always being displayed so when the image is loading, the initials are behind.

 But I'll need to test several cases to check that z-index would work correctly in all of them.

components/avatar/avatar.vue
juliodialpad: Right, I missed that one, nice catch.

components/avatar/avatar.vue
braddialpad: Might work if you just move the slot after the initials in the DOM order.

components/avatar/avatar.vue
braddialpad: as I said, knew you had a good reason 😄

juliodialpad:
> This component is starting to feel a bit overcomplicated, though I don't think we can fix some of the issues without changing the public interface and making a breaking change. Was thinking it may simplify things if we used props instead of slots for some of this stuff since accessing the markup within the slots is fairly cumbersome and has reactivity issues.

You bet, this component is too complicated now IMO, but yeah, making a change in the public interface would be a breaking change and a large migration.

Automatically detecting the slot content seem good at the beginning but now its hard to maintain this component.

components/avatar/avatar.vue
juliodialpad: I didn't knew I could do it that way, thanks!

iropolo (approved):
LGTM, I'm agree also to start to rethink this component in the future.

components/avatar/avatar.vue
juliodialpad: Doesn't work because initials have `d-ps-absolute`.

Tested in popover and modal and everything works fine so we shouldn't have any problems.

juliodialpad:
Fixed issues I found while testing avatar over popover and modal to check that `d-zi-base` didn't break anything.

Fixed tests as well.

braddialpad (approved):

juliodialpad:
While working on Vue 3 version, found a lot of problems with the way I made it on Vue 2 and realized that there was a better way of handling that logic.

Refactored to keep almost the same logic in both Vue 2 and Vue 3 versions.

It should be cleaner to understand what's happening and also fixed some issues I found.

The z-index should work fine as the [stacking-context](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Positioning/Understanding_z_index/The_stacking_context) of the avatar contains it.

Please give another review @braddialpad @iropolo

braddialpad (approved):
This is cleaner, nice job.

PR #804, opened 2023-03-08
feat(leftbar): fixes and additions to leftbar row components

braddialpad:
# feat(leftbar): fixes and additions to leftbar row components

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

The following changes have been made to leftbar components:

- add new general row types, quickstart, coaching group, coaching center
- add isTyping animation - placeholder styling for now, will be updated when the design team has finished with it
- fixed bug where call button would remain displaying after clicking on it and mousing out of the row
- removed focus-visible polyfill from storybook since all browsers now support it.
- added tooltip text props for elements that had them in the original implementation
- added title for row
- added screen reader support and tested screen reader read-out made sense.

## :bulb: Context

Fixes a number of issues and missing content from the leftbar rows

## :pencil: Checklist

## :crystal_ball: Next Steps

Implement new features into firespotter branch


recipes/leftbar/general_row/general_row.vue
```diff
@@ -14,11 +14,18 @@
...
           <dt-recipe-leftbar-general-row-icon
+            :class="{ 'd-o50': isTyping }"
             :type="getIcon"
             :color="color"
             data-qa="dt-leftbar-row-icon"
           />
         </slot>
+        <div
+          v-if="isTyping"
+          class="dt-leftbar-row--typing d-bgc-black-600"
```
francisrupert: Cleaner BEM: `dt-leftbar-row__is-typing`.

recipes/leftbar/style/leftbar_row.less
```diff
@@ -82,6 +82,40 @@
     --leftbar-row-omega-min-width: var(--leftbar-row-alpha-width);
   }
 
+  &--typing {
+    height: 24px;
+    width: 24px;
```
francisrupert: Ideally I would like to reuse `var(--avatar-size-shape)` to sync the size, but that's localized to `.d-avatar`. If you can't figure a easy way to sync that size, can just do...

```less
&--typing {
  --typing-size-shape: calc(var(--size-300) * 3)

  height: var(--typing-size-shape);
  width: var(--typing-size-shape);
...
```

recipes/leftbar/style/leftbar_row.less
```diff
@@ -82,6 +82,40 @@
...
+    align-items: center;
+    justify-content: center;
+    gap: 2px;
+    border-radius: var(--br-pill);
+    opacity: 0.75;
+  }
+
+  &--typing span {
+    transition: all 500ms ease;
+    background: var(--white);
```
francisrupert: Color combination TBD via @bhawkins-dialpad per https://dialpad.slack.com/archives/C04QQ1GKT96/p1678310674717059. Just ensure design token usage to future-proof dark theme.

recipes/leftbar/style/leftbar_row.less
braddialpad: Hmm yeah don't think we can do this without moving variables or a mixin outside of the avatar and referencing it in both. Will do your suggestion for now.

braddialpad:
@juliodialpad @iropolo ready for developer review.

components/popover/tippy_utils.js
```diff
@@ -57,21 +56,17 @@ export const getPopperOptions = ({
...
 
-export const getAnchor = (anchorWrapper, tabIndex = '0') => {
+export const getAnchor = (anchorWrapper) => {
   const anchor = anchorWrapper?.children[0];
   if (!anchor) return createAnchor(anchorWrapper);
-  if (!findFirstFocusableNode(anchor)) {
-    anchor.setAttribute('tabindex', tabIndex);
-  }
   return anchor;
 };
```
braddialpad, line 71: This was causing issues with focus in the contact row. not sure why we force tabindex=0 here, it should be focusable if the consumer makes it focusable.

juliodialpad: Maybe this was a validation we needed when first created popover, will need to check if this doesn't break anything.

juliodialpad (changes requested):
The code is looking really good, clean and easy to follow up, just found one potential improvement.

https://user-images.githubusercontent.com/87546543/224130335-6e3ea3ca-9bb8-405a-9b56-975f4c873f67.mov

When hovering with the mouse, the 'action' button is shown and while focusing with keyboard it is not until you tab to it. I think the button should be shown since the moment you start focusing the row, right?

Hmm, I think the unread count tooltip is not accessible because when you hover over, the action button shows up. Is that the right behavior?

juliodialpad:
image

Avatar presence default value should be the actual value, not the constant.

juliodialpad:
Screenshot 2023-03-09 at 1 10 00 p m

If presence text is too large it overflows the entire row and displaces the user status. Is this something we want to validate or is it ok to leave it this way?

braddialpad:
> Uploading Screen Recording 2023-03-09 at 1.00.35 p.m..mov…

Looks like your upload didn't work

juliodialpad:
> > Uploading Screen Recording 2023-03-09 at 1.00.35 p.m..mov…
> 
> Looks like your upload didn't work

Sorry, updated

braddialpad:
> Screenshot 2023-03-09 at 1 10 00 p m
> 
> If presence text is too large it overflows the entire row and displaces the user status. Is this something we want to validate or is it ok to leave it this way?

Since presence text is not user entered and only uses specific preset values that are pretty short it's probably okay? Though I'll call on @francisrupert to confirm.

components/popover/tippy_utils.js
braddialpad: yeah seems okay so far.. haven't really tested that much though will do some more testing.

braddialpad:
> When hovering with the mouse, the 'action' button is shown and while focusing with keyboard it is not until you tab to it. I think the button should be shown since the moment you start focusing the row, right?

I believe this was actually intended right @francisrupert?

braddialpad:
> Hmm, I think the unread count tooltip is not accessible because when you hover over, the action button shows up. Is that the right behavior?

Yeah you cannot ever access the unread count tooltip on a contact item, only on group / general due to this. It will still be read out via screen reader though.

components/popover/tippy_utils.js
braddialpad: I think this was from before when we used to have a tabIndex prop. it would set the tabIndex on the anchor to this prop. This prop was removed long ago so I think it is fine to remove.

Tested a bunch in dt-vue, doesn't seem to have broken anything.

braddialpad:
Going to merge this since I believe the remaining issues are intentional behaviour. If design says they are not, will make a new PR.

juliodialpad:
> Going to merge this since I believe the remaining issues are intentional behaviour. If design says they are not, will make a new PR.

Sounds good for me, code looks good.

juliodialpad (approved):

PR #807, opened 2023-03-13
fix: add dialtone-icons to peerDependencies

juliodialpad:
# Add dialtone-icons to peerDependencies

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Added dialtone-icons to peerDependencies

## :bulb: Context

Dialtone icons was not a part of peer dependencies so the Dialtone-vue users needed to install dialtone-icons in order to get the icon component working. This should fix the issue.

## :pencil: Checklist

- [x] I have reviewed my changes- [ ] I have added tests

## :crystal_ball: Next Steps

Remove dialtone-icons dependency from the products.

### DIFF
--- a/package.json
+++ b/package.json
@@ -137,6 +137,7 @@
   },
   "peerDependencies": {
     "@dialpad/dialtone": ">=7.19",
+    "@dialpad/dialtone-icons": "latest",
     "vue": ">=2.6"
   },
   "engineStrict": true,
--- a/package-lock.json
+++ b/package-lock.json
@@ -5736,21 +5736,6 @@
         "node": ">=6"
       }
     },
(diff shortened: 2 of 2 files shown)


package.json
braddialpad, line 1: `"@dialpad/dialtone-icons": "latest",` is already part of the normal "dependencies" though. This means it should be installed as part of the package. IE when you install @dialpad/dialtone-vue @dialpad/dialtone-icons is automatically installed as well.

juliodialpad: Hmm, so maybe the "latest" tag is the issue here?

So adding the tag version instead should fix it, right?

braddialpad: Yes I believe so.

juliodialpad: Will close this and create a PR with that updates then, thanks!

PR #814, opened 2023-03-14
fix(leftbar-group-row): add conditional img render

braddialpad:
# fix(leftbar-group-row): add conditional img render

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Was causing an error when group row had a falsy avatarSrc. This should fix it.


juliodialpad (approved):
LGTM thanks!

PR #828, opened 2023-03-16
feat(badge): add decoration prop

jimut-dialpad:
# PR Title

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Add decoration prop in badge component.

## :bulb: Context

[https://dialpad.atlassian.net/browse/DT-1028](https://dialpad.atlassian.net/browse/DT-1028)

## :pencil: Checklist

- [x] I have added tests

Screenshot 2023-03-16 at 8 18 30 PM


package-lock.json
MengzhenGao, line 1: @jimut-dialpad Not sure why there's such big changes in `package-lock` but we probably don't need to push it?

components/badge/badge.vue
```diff
@@ -97,12 +102,25 @@ export default {
       default: 'default',
       validator: (type) => Object.keys(BADGE_TYPE_MODIFIERS).includes(type),
     },
+
+    /**
+     * Decoration for the badge
```
MengzhenGao: Just a suggestion, is it better to mention here the decoration prop can only be used with kind: `label` and type `default` and no `iconLeft` and `iconRight` so eng can easily know such limitations? 

image

components/badge/badge.vue
```diff
@@ -121,6 +139,24 @@ export default {
...
+      if (this.kind !== 'label' || this.type !== 'default') {
+        console.error('DtBadge error: decoration prop can only be used with kind: \'label\' and type: \'default\'.');
+      }
+
+      if (this.hasIcons()) {
+        console.error('DtBadge error: decoration prop cannot be used with iconLeft or iconRight.');
+      }
+    },
+
+    hasIcons () {
```
MengzhenGao: not important but `hasIcons` looks more like a computed variable here

components/badge/badge.stories.js
```diff
@@ -54,6 +54,15 @@ export const argTypesData = {
...
     if: { arg: 'type', neq: 'ai' },
   },
+
+  decoration: {
+    defaultValue: 'default',
+    control: {
+      type: 'select',
+      options: Object.keys(BADGE_DECORATION_MODIFIERS),
+    },
+    // TODO: Is there any way to add conditions on more than one control?
```
MengzhenGao: curious what other conditions needed?

package-lock.json
braddialpad: Yeah.. not sure what is going on here but it doesn't seem right.. Possibly a different version of npm? this project uses node v16 npm v8

components/badge/badge.vue
braddialpad: Yes for sure.

components/badge/badge.mdx
```diff
@@ -65,6 +65,10 @@ import { DtBadge } from '@dialpad/dialtone-vue';
 ```jsx
 <dt-badge
   text="badge"
+  decoration="default | black-400 | black-500 | black-900 | red-200 | red-300 | red-400
```
braddialpad: You can just remove "default" here since users will just not include the prop if they don't want it.

components/badge/badge.vue
```diff
@@ -121,6 +139,24 @@ export default {
...
+      if (this.decoration === 'default') return;
+
+      if (this.kind !== 'label' || this.type !== 'default') {
+        console.error('DtBadge error: decoration prop can only be used with kind: \'label\' and type: \'default\'.');
+      }
+
+      if (this.hasIcons()) {
+        console.error('DtBadge error: decoration prop cannot be used with iconLeft or iconRight.');
+      }
+    },
```
iropolo, line 156: `validateDecorationPropCombination() {
  const isDefaultDecoration = this.decoration === 'default';
  const isInvalidDecorationCombination = !isDefaultDecoration && (this.kind !== 'label' || this.type !== 'default');
  const hasIcons = this.hasIcons();

  if (isInvalidDecorationCombination) {
    console.error('DtBadge error: decoration prop can only be used with kind: \'label\' and type: \'default\'.');
  } else if (hasIcons) {
    console.error('DtBadge error: decoration prop cannot be used with iconLeft or iconRight.');
  }
}`

Not necessary to replace but here is a suggestion.

package-lock.json
juliodialpad: By the change on this line, seems like he's using Node v19 and that's why there's a lot of changes

BEFORE
```json
"engines": {
        "node": "8 || 9 || 10 || 11 || 12 || 13 || 14 || 15 || 16 || 17 || 18"
      }
```

AFTER
```json
"engines": {
        "node": "8 || 9 || 10 || 11 || 12 || 13 || 14 || 15 || 16 || 17 || 18 || 19"
      }
```

braddialpad:
Deploy preview failing for some reason... I'm going to try cleaning our github pages branch

components/badge/badge.vue
braddialpad: agree, doesn't need to be a method.

components/badge/badge_constants.js
```diff
@@ -12,3 +12,29 @@ export const BADGE_KIND_MODIFIERS = {
   label: '',
   count: 'd-badge--count',
 };
+
+export const BADGE_DECORATION_MODIFIERS = {
+  default: '',
```
braddialpad: Make this null or undefined rather than empty string or else it will actually insert an empty string into the class list, which isn't that big of deal but it's still a bit messy.

package-lock.json
braddialpad: Yeah you should use the latest LTS rather than current as current is not considered stable. The latest LTS is v18, that should be fine I'd think. 

Guess we should upgrade the project to 18 though.

components/badge/badge.vue
```diff
@@ -121,6 +139,24 @@ export default {
       if (this.type === 'ai' && this.kind === 'count') {
         console.error('DtBadge error: type: \'ai\' with kind: \'count\' is an invalid combination.');
       }
```
braddialpad, line 153: Might as well put this one into it's own function as well just to keep things clean.

components/badge/badge_constants.js
iropolo: mhmm I cannot reproduce the blank space, it is not being added if its default

image

components/badge/badge.mdx
iropolo: If `default` is not necessary then I would recommend to set the default value of decoration prop to `null` and remove `default` also from BADGE_DECORATION_MODIFIERS

components/badge/badge.stories.js
jimut-dialpad: This prop is only valid when kind is label and type is default with no iconLeft and iconRight.
So it would have been better if we could have shown this only when the above conditions were met.
By writing something like
```
if: [
  { arg: 'kind', eq: 'label' },
  { arg: 'type', eq: 'default' },
  { arg: 'iconLeft', eq: '' },
  { arg: 'iconRight', eq: '' }
]
```
But that's not possible.

components/badge/badge_constants.js
jimut-dialpad: Removed this from here. Made default to be undefined.

package-lock.json
jimut-dialpad: I was using node 18.15.0 with npm 9.5.0.
All this change happened because I deleted and recreated the lock file.
Reverted the changes of this file now.

jimut-dialpad:
Thank you so much for the awesome review everyone. 🤩

MengzhenGao (approved):
LGTM! Thanks for taking up this task and contributing to Dialtone!

Not related to this PR but Accessibility Tests has been failing for a while
https://github.com/dialpad/dialtone-vue/actions/workflows/a11y_tests.yml

braddialpad:
> Not related to this PR but Accessibility Tests has been failing for a while https://github.com/dialpad/dialtone-vue/actions/workflows/a11y_tests.yml

Yep just ignore for now, gonna work on fixing these today.

braddialpad (approved):
Looks good Jimut, thanks!

MengzhenGao:
@braddialpad Do you mind releasing it so we can utilize it on some tasks?

braddialpad:
> @braddialpad Do you mind releasing it so we can utilize it on some tasks?

Sure releasing now!

PR #830, opened 2023-03-16
feat(avatar): add canvasClass prop

braddialpad:
# feat(avatar): add canvasClass prop

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Adding canvasClass to avatar so DPM can set a custom size as mentioned in this ticket https://dialpad.atlassian.net/browse/DT-954

### DIFF
--- a/components/avatar/avatar.vue
+++ b/components/avatar/avatar.vue
@@ -7,7 +7,7 @@
   >
     <div
       ref="canvas"
-      class="d-avatar__canvas"
+      :class="[canvasClass, 'd-avatar__canvas']"
     >
       <!-- @slot Slot for avatar content -->
       <slot v-if="showDefaultSlot" />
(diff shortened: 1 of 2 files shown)


iropolo (approved):
LGTM

juliodialpad (commented):
This is going to give too much freedom on what you can do with Avatar, wouldn't be better to add like a 'custom-size' prop if we only need to change the size?

Or maybe adding something like 2x, 4x and 8x sizes.

braddialpad:
> This is going to give too much freedom on what you can do with Avatar, wouldn't be better to add like a 'custom-size' prop if we only need to change the size?
> 
> Or maybe adding something like 2x, 4x and 8x sizes.

I actually think it remains more hidden doing it this way. If I add a dedicated size prop they're going to start randomly changing avatar sizes all over the place just because they see that prop 😆

juliodialpad:
> > This is going to give too much freedom on what you can do with Avatar, wouldn't be better to add like a 'custom-size' prop if we only need to change the size?
> > Or maybe adding something like 2x, 4x and 8x sizes.
> 
> I actually think it remains more hidden doing it this way. If I add a dedicated size prop they're going to start randomly changing avatar sizes all over the place just because they see that prop 😆

Sounds fair 😅 go ahead and merge then.

PR #839, opened 2023-03-17
feat(emoji-picker): implement initial mvp

iropolo:
# feat(emoji-picker) stub emoji-picker

- [x] Feature

## :book: Description

[DT-969](https://dialpad.atlassian.net/browse/DT-969)
[DT-1082](https://dialpad.atlassian.net/browse/DT-1082)

New component for vue3 version of Dialtone.
This PR creates the STUB for the emoji-picker component.

[DT-969]: https://dialpad.atlassian.net/browse/DT-969?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DT-1082]: https://dialpad.atlassian.net/browse/DT-1082?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iropolo:
First STUB for the emoji-picker component.

Dont be so rigorous with the code, it will definitely change when I will start to work on it, instead, check how it will be the structure of it and the communication between.

components/emoji_picker/emoji_picker.vue
```diff
@@ -0,0 +1,86 @@
...
+   * This list is necessary to fill the recently used tab
+   * @type {Array}
+   * @default []
+   * @example
+   * <dt-emoji-picker :recentlyUsedEmojis="['😀', '😁']" />
+   */
+  recentlyUsedEmojis: {
+    type: Array,
+    default: () => [],
+  },
```
juliodialpad: This list will just receive an array of Emojis or could it receive shortCodes too as in https://vue.dialpad.design/?path=/story/components-emoji-text-wrapper--default ?

components/emoji_picker/modules/emoji_selector.vue
```diff
@@ -0,0 +1,87 @@
+<template>
+  <div class="d-emoji-picker__selector">
+    <!--    Emoji Selector -->
+    <div
+      v-for="emoji in EMOJI_LIST"
```
juliodialpad: What would that EMOJI_LIST const be?

juliodialpad (commented):
I like this organization, is easy to understand what's happening and so it would be easier to maintain.

Just a couple comments, but looks good so far.

components/emoji_picker/emoji_picker.vue
iropolo: ye... still dunno.. will figure out later lol, Im kind of sure this prop will be needed, but honestly dont fully convince how it will work, but it will sth like you are saying

components/emoji_picker/modules/emoji_selector.vue
iropolo: mhmm... it will be definetly sth from the emoji.json (`node_modules/emoji-toolkit/emoji_strategy.json`)
and/or some utility I will probably create in `common/emoji.js`.

It will be important to have attention on how we will manage this array for the perform impact.
Because we have a bunch of tabset and skins that will affect the search behaviour... not big deal, we can handle it :)

juliodialpad:
Also, is this component going to be **vue3 only**? Otherwise it'd be better to rewrite this with **Options API** to simplify the maintenance when we need to do changes to this component.

iropolo:
> Also, is this component going to be **vue3 only**? Otherwise it'd be better to rewrite this with **Options API** to simplify the maintenance when we need to do changes to this component.

Both scroller component and emoji picker will be vue 3 only :), as I understand they will be DP2.0 components

components/emoji_picker/emoji_picker.vue
braddialpad: Since this stub serves as a tech spec we should probably figure out and write some details about how we are going to do this. A primary purpose of this exercise is that we can determine if an idea is going to work or not before we start actually coding it. This prevents us from having to go back and rewrite things later.

braddialpad:
> > Also, is this component going to be **vue3 only**? Otherwise it'd be better to rewrite this with **Options API** to simplify the maintenance when we need to do changes to this component.
> 
> Both scroller component and emoji picker will be vue 3 only :), as I understand they will be DP2.0 components

Scroller is definitely Vue 3 only... This emoji picker I am not so sure about. Let me talk to a couple of people and verify. 

That said they do have [composition API available for vue 2](https://blog.vuejs.org/posts/vue-2-7-naruto) now, we can probably use this!

components/emoji_picker/emoji_picker.vue
```diff
@@ -0,0 +1,122 @@
...
+
+});
+const emits = defineEmits(
+  /**
+     * It will emit the selected emoji
+     * @event selected-emoji
+     * @param {Object} emoji - The selected emoji from the emoji selector
```
braddialpad, line 140: hmm what exactly does this emit? a shortcode? emoji unicode character?

components/emoji_picker/modules/emoji_selector.vue
```diff
@@ -0,0 +1,87 @@
...
+      :key="emoji"
+      class="d-emoji-picker__selector--emoji-list"
+    >
+      <p
+        class="d-emoji-picker__selector--emoji-list__tab"
```
braddialpad: You shouldn't double nest BEM.

See: https://getbem.com/faq/#css-nested-elements

components/emoji_picker/modules/emoji_data.vue
```diff
@@ -0,0 +1,19 @@
+<template>
+  <div class="d-emoji-picker__hovering">
+    <!--    Display emoji data -->
+  </div>
+</template>
```
braddialpad: What exactly does this component do? Is it like a tooltip?

components/emoji_picker/modules/emoji_selector.vue
```diff
@@ -0,0 +1,87 @@
...
+
+      <div>
+        <dt-emoji
+          v-for="emoji in filteredEmojiList"
+          :key="emoji.code"
+          class="d-emoji-picker__selector--emoji-list__emoji"
+          :code="emoji.code"
+          @mouseover="$emit('emoji-data', emoji)"
+          @mouseleave="$emit('emoji-data', null)"
+          @click="$emit('selected-emoji', emoji)"
```
braddialpad: dt-emoji is not interactive, so it should be wrapped in a button

components/emoji_picker/modules/emoji_search.vue
```diff
@@ -0,0 +1,19 @@
+<template>
+  <div class="d-emoji-picker__search">
+    <dt-input
+      placeholder="Search..."
```
braddialpad: You might already know this but we can't actually put this english text here due to i18n. all language based text needs to be passed in via props.

components/emoji_picker/modules/emoji_skin_selector.vue
```diff
@@ -0,0 +1,16 @@
...
+<script setup>
+const emits = defineEmits([
+  /**
+   * The skin tone that was selected
+   * @event skin-tone
+   * @type {Number}
+   */
+  'skin-tone',
+]);
+</script>
```
braddialpad: Can you comment a few more details on how this one is going to work?

Does it change the look of all the "skin tone based" emojis in the selector?
How does it work if you interact with it via keyboard?
left / right arrows to select skin tone?
Is there a way to access it without tabbing through the entire list of emojis?
Is there any other way to change skin tone when you are selecting an emoji?

components/emoji_picker/emoji_picker.vue
```diff
@@ -0,0 +1,122 @@
...
+      <!--        @emoji-data="updateEmojiData" -->
+      <!--        @selected-emoji="emits('selected-emoji', $event)" -->
+      <!--      /> -->
+    </div>
+    <div className="d-emoji-picker--footer">
+      <!--      <emojiData :emoji-data="emojiData" /> -->
+      <!--      <emojiSkinSelector @skin-tone="updateSkinTone" /> -->
+    </div>
+  </div>
+</template>
```
braddialpad: How would someone trigger the emoji picker from the UI? do we need an anchor slot or is this component something that would be nested inside a DtPopover?

components/emoji_picker/modules/emoji_selector.vue
```diff
@@ -0,0 +1,87 @@
...
+
+  /**
+   * The skin tone to apply to the emoji list
+   * @type {String}
+   * @default null
+   */
+  skinTone: {
+    type: Number,
+    default: null,
+  },
```
braddialpad: How do you plan to filter skin tone using the existing json data? You don't have to write code or anything but a couple sentences about the method you intend on using is useful to figure out at this point.

components/emoji_picker/modules/emoji_tabset.vue
```diff
@@ -0,0 +1,89 @@
...
+  { id: '2', panelId: '2', label: 'Custom', icon: 'Asterisk' },
+  { id: '3', panelId: '3', label: 'Smiley’s and people', icon: 'satisfied' },
+  { id: '4', panelId: '4', label: 'Nature', icon: 'living-thing' },
+  { id: '5', panelId: '5', label: 'Food', icon: 'food' },
+  { id: '6', panelId: '6', label: 'Activity', icon: 'Asterisk' },
+  { id: '7', panelId: '7', label: 'Travel', icon: 'Asterisk' },
+  { id: '8', panelId: '8', label: 'Objects', icon: 'lightbulb' },
+  { id: '9', panelId: '9', label: 'Symbols', icon: 'heart' },
+  { id: '10', panelId: '10', label: 'Flags', icon: 'flag' },
+];
```
braddialpad: Annoyingly, we will have to pass in all of these labels containing english text for i18n.

components/emoji_picker/modules/emoji_selector.vue
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * Filtered emoji list based on the filter
+ */
+const filteredEmojiList = computed(() => {
+  if (props.emojiFilter) {
+    return emojiList.filter((emoji) => emoji.includes(props.emojiFilter));
+  }
+  return emojiList;
+});
```
braddialpad: We're also going to need to filter keywords

components/emoji_picker/emoji_picker.vue
iropolo: I cannot see the big deal here, maybe thats the problem..., what I expect to receive an Object with objects with the emoji object data similiar as what we have currently.

image


Example:

```
recentlyUsedEmojis: {
    "1f469-1f3fb-1f692": {
          "name":"woman firefighter: light skin tone",
          "category":"people",
          "shortname":":woman_firefighter_tone1:",
          "shortname_alternates":[":woman_firefighter_light_skin_tone:"],
          "keywords":["firefighter","firetruck","light skin tone","woman","uc8"],
          "unicode_output":"1f469-1f3fb-200d-1f692"
    } ,...
}
```

If there is some postman or swagger to check how it will be the response or how the developer will handle this..
But at this moment I will expect what I currently have.

---
Code updated

components/emoji_picker/emoji_picker.vue
iropolo: Initially it will retrieve the {Object} emoji, this is all the emoji data.

```
{
   "name":"woman firefighter: light skin tone",
   "category":"people",
   "shortname":":woman_firefighter_tone1:",
   "shortname_alternates":[":woman_firefighter_light_skin_tone:"],
   "keywords":["firefighter","firetruck","light skin tone","woman","uc8"],
   "unicode_output":"1f469-1f3fb-200d-1f692"
}
```

components/emoji_picker/modules/emoji_selector.vue
iropolo: Mhmm ye it was growing ugly, let me clean it a bit

---
Code updated

components/emoji_picker/modules/emoji_data.vue
iropolo: Mhmm seems like the classname was deprecated.
I updated it now with `d-emoji-picker__data`

This component is in charge of this behaviour:

image

---
Code updated

components/emoji_picker/modules/emoji_selector.vue
iropolo: Nice catch 👍🏼 , will update it now

---
Code updated

components/emoji_picker/modules/emoji_search.vue
iropolo: Sure, will start to generate these props 👍🏼 

---
Code updated

components/emoji_picker/modules/emoji_data.vue
braddialpad: ahhh yeah that makes sense why it triggers on mouseover then. Calling this emoji_data might be a bit too generic. Maybe emoji description? highlighted emoji description?

components/emoji_picker/emoji_picker.vue
braddialpad: Got it, we might also want to include the actual emoji unicode character in there.

components/emoji_picker/modules/emoji_selector.vue
```diff
@@ -4,10 +4,10 @@
     <div
       v-for="emoji in EMOJI_LIST"
       :key="emoji"
-      class="d-emoji-picker__selector--emoji-list"
+      class="d-emoji-picker__selector-emoji-list"
```
braddialpad: This is still kinda deeply nesting things in a slightly different way.

Example to fix:

`d-emoji-picker__selector-emoji-list` -> `d-emoji-picker__list`
`d-emoji-picker__selector-emoji-list-tab` -> `d-emoji-picker__tab`
`d-emoji-picker__selector-emoji-list-emoji` -> `d-emoji-picker__emoji`

components/emoji_picker/modules/emoji_selector.vue
```diff
@@ -13,15 +13,20 @@
...
-          @mouseover="$emit('emoji-data', emoji)"
-          @mouseleave="$emit('emoji-data', null)"
+          type="button"
+          class="d-emoji-picker__selector-emoji-list-emoji-button"
+          :aria-label="emoji.name"
           @click="$emit('selected-emoji', emoji)"
-        />
+          @focusin="$emit('emoji-data', emoji)"
+          @focusout="$emit('emoji-data', null)"
+        >
```
braddialpad: hmm don't we still need mouseover and mouseleave in addition to focus?

components/emoji_picker/modules/emoji_skin_selector.vue
iropolo: Sure.... this code is deprecated, because skin tone its not a number, its a composition CODE + LETTER.

image

image

For each mutable skin emoji we have 6 emojis

1. palms up together - 1f64c
2. palms up together: light skin tone - 1f64c-1f3fb
3. palms up together: medium-light skin tone - 1f64c-1f3fc
4. palms up together: medium skin tone - 1f64c-1f3fd
5. palms up together: medium-dark skin tone - 1f64c-1f3fe
6. palms up together: dark skin tone - 1f64c-1f3ff

Each skin tone its a composition 
{EMOJI_UNICODE}-{SKIN_CODE}{SKIN_TONE}
Emoji unicode: 1f64c
Skin code (SAME FOR ALL): 1f3f
Skin tone: b/c/d/e/f

So.. I will update this code to handle this behaviour.

image

components/emoji_picker/emoji_picker.vue
braddialpad: I'd say it's unlikely the consumer will pass in an entire emoji data object into here. Probably just an array of unicode characters?

components/emoji_picker/modules/emoji_skin_selector.vue
braddialpad: This is excellent detail and exactly the type of thing developers would like to see here. Make sure to comment this into your component or include in documentation.

iropolo: Now knowing this, let me answer your questions:

- Does it change the look of all the "skin tone based" emojis in the selector?
Yes, as thats the slack behaviour, the only application I can remember now which has specific skin tone in each emoji its whatsapp

Slack skin tone for all emojis
image


Whatsapp skin tone for each emoji behaviour
image


- How does it work if you interact with it via keyboard?
There is no keyboard navigation specification.
In this case I would replicate the slack emoji picker navigation.

Its work in this way:

1. On emoji picker open it will focus the search input
2. After tab it will jump to the first icon of each tabset.
3. After navigate all the tabset its jump to an specific custom behaviour of "Add Item" (we dont have it) so we will jump directly to the skin tone selector.
4. After tab it will jump to the first Tab of the tabset.
5. After tab it will jump again to the search input.

    When the focus is on the emoji selector, you can start to navigate through all the emojis with arrow keys and select with Enter.
    When the focus is on the Tabset, you can start to navigate through all the Tabs with arrow keys and select with Enter.
    When the focus is on the skin-tone selector you can "Enter" and open the skin selector, then you can navigate through all the skins with arrow keys and select with Enter.


- left / right arrows to select skin tone?
Yes

- Is there a way to access it without tabbing through the entire list of emojis?
I cannot found the shortcode to jump directly to skin tone on Slack behaviour.

- Is there any other way to change skin tone when you are selecting an emoji?
I cannot found the shortcode to to change skin tone on Slack behaviour.
On Whatsapp you can maintain click for 1 second and it will appear the skin selector above.

components/emoji_picker/emoji_picker.vue
iropolo: What I imagine is they will place this on a DtPopover component.

As I understand, that's not our responsibility, we are developing the DtEmojiPicker, they can play with it as they want.
We can place it on a DtPopover on DtInput simulating a chatbox in Variants on Storybook.

components/emoji_picker/modules/emoji_skin_selector.vue
braddialpad: > * Is there a way to access it without tabbing through the entire list of emojis?
>   I cannot found the shortcode to jump directly to skin tone on Slack behaviour.

I think you answered this by saying the emojis are navigated with arrow keys though, that is good enough!


> * Is there any other way to change skin tone when you are selecting an emoji?
>   I cannot found the shortcode to to change skin tone on Slack behaviour.
>   On Whatsapp you can maintain click for 1 second and it will appear the skin selector above.

Interesting, yeah don't think we need to worry about that here. Just the bottom skin selector should be fine.

components/emoji_picker/emoji_picker.vue
braddialpad: Yep I agree this is probably a good way to go.

components/emoji_picker/modules/emoji_skin_selector.vue
iropolo: One more thing I realised doing this research.

We also have GENDER distinctions.
image


And gender with skin tone distinctions.
image


And some like this
image

image


So... let me figure out how I will manage this...

This "UNICODE CODE STORM" will only be affect by skin selector... because we need to filter it, this will not affect the emoji selector because we just need the code and thats it, we can access the emoji doing bracket notation to the object like this `return Emojis['1f3cc-1f3ff-2642']`

braddialpad: Yep and it also gets really crazy when there are multiple people in an emoji with different skin tones for example. Not sure what the plan is on these ones...

components/emoji_picker/modules/emoji_selector.vue
iropolo: Done my research above I still need to figure out this kind of methods, because the complexity and also the perform impact.

What I quickly suggest is to have different emojis arrays for each skin tone.

components/emoji_picker/modules/emoji_tabset.vue
iropolo: Mhmm, you mean expose a prop called `tabSetLabels` on our emoji_picker and then the developer can provide correct i18n translations...?

```
tabSetLabels: {
  recentlyUsed: 'Recently used',
  custom: 'Custom",
  ....
}
```

braddialpad: Yep exactly.

components/emoji_picker/modules/emoji_selector.vue
iropolo: Yes wil add a TODO in the code to not forget 👍🏼

components/emoji_picker/modules/emoji_selector.vue
iropolo: Mhmm, ye ok, hopefully will no have any conflict in the future with the other modules components, but its seems not

---
Code updated

components/emoji_picker/modules/emoji_selector.vue
braddialpad: Hmm think we will actually have to pass the emoji list into emoji_picker in order to handle custom emojis. This could default to the standard emoji_strategy.json data, but they could also pass in their own.

components/emoji_picker/modules/emoji_selector.vue
iropolo: Yes, right, because it will handle the focus on keyboard navigation and the hover with the mouse 👍🏼 

---
Code Updated

components/emoji_picker/emoji_picker.vue
iropolo: Mhmm.. it is possible, its depends on how I will manage show the emojis.
I will put a TODO here to try to simplify this.

---
Code updated

components/emoji_picker/emoji_picker.vue
iropolo: Cool, I will put a TODO to not forget the unicode_character.

In some cases `unicode_output` it will be the same to our `unicode_character`
But in some cases not...

image


image


I dont exactly why, but anyways we will provide both..

---
Code updated

braddialpad: Yeah I found this a bit confusing when I was working on emoji too. It is a "variation selector" https://emojipedia.org/variation-selector-16/

braddialpad (approved):
I think this looks pretty good at this point.

components/emoji_picker/emoji_picker.vue
braddialpad, line 1: Screenshot 2023-04-24 at 5 31 37 PM

Maybe the designers already mentioned this, but the scrollbar seems a bit tight to the search input.

components/emoji_picker/emoji_picker.vue
```diff
@@ -0,0 +1,240 @@
...
+    padding: var(--su4) var(--su4) 0 var(--su4);
+    position: relative;
+    &::after{
+      content: '';
+      position: absolute;
+      right: 0;
+      bottom: 0;
+      left: 0;
+      height: 1px;
+      background-color: var(--black-200) !important;
```
braddialpad: Should we use a surface color here?

components/emoji_picker/emoji_picker.vue
```diff
@@ -0,0 +1,240 @@
...
+   * The list of tabsets to show, it is necessary to be updated with the correct language
+   * It must respect the provided order.
+   * @type {Array}
+   * @default ['Most recently used', 'People', 'Nature', 'Food', 'Activity', 'Travel', 'Objects', 'Symbols', 'Flags']
+   * @example
+   * <dt-emoji-picker :tabSetLabels="['Most recently used', 'People', 'Nature', 'Food', 'Activity', 'Travel', 'Objects', 'Symbols', 'Flags']" />
+   */
+  tabSetLabels: {
+    type: Array,
+    default: () => ['Most recently used', 'Smileys and people', 'Nature', 'Food', 'Activity', 'Travel', 'Objects', 'Symbols', 'Flags'],
```
braddialpad: Should we have some kind of warning if this is left as default? Or maybe we should just remove the default? The problem if we leave it with english text as default is that when people go to use it they will just not realize they have to pass in the i18n version and will just leave it as is.

components/emoji_picker/emoji_picker.vue
```diff
@@ -0,0 +1,240 @@
...
+/**
+ * It will clean the state of the component before unmounting
+ */
+onBeforeUnmount(() => {
+  cleanUp();
+});
+
+function cleanUp () {
+  searchQuery.value = '';
+  selectedTabset.value = {};
```
braddialpad: would this data actually persist after unmounting?

components/emoji_picker/modules/emoji_tabset.vue
```diff
@@ -0,0 +1,144 @@
...
+          :panel-id="tab.panelId"
+          :label="tab.label"
+          class="d-ml2 d-mr2"
+          @click.capture.stop="selectTabset(tab.id)"
+        >
+          <dt-icon
+            size="400"
+            :name="tab.icon"
+          />
+        </dt-tab>
```
braddialpad: hmm wonder if we could use dt-stack here rather than margin for spacing? or maybe that breaks the tab component...

components/emoji_picker/modules/emoji_tabset.vue
```diff
@@ -0,0 +1,144 @@
...
+const TABS_DATA = [
+  { label: 'Recently used', icon: 'clock' },
+  { label: 'Smiley’s and people', icon: 'satisfied' },
+  { label: 'Nature', icon: 'living-thing' },
+  { label: 'Food', icon: 'food' },
+  { label: 'Activity', icon: 'object' },
+  { label: 'Travel', icon: 'transportation' },
+  { label: 'Objects', icon: 'lightbulb' },
+  { label: 'Symbols', icon: 'heart' },
+  { label: 'Flags', icon: 'flag' },
```
braddialpad, line 88: Should this use `tabSetLabels` prop?

components/emoji_picker/modules/emoji_tabset.vue
```diff
@@ -0,0 +1,144 @@
...
+    type: Boolean,
+    default: false,
+  },
+
+  scrollIntoTab: {
+    type: Number,
+    required: true,
+  },
+
+  isScrollingWithScrollTo: {
```
braddialpad: isCurrentlyScrolling?

components/emoji_picker/emojis/activity-dark.json
braddialpad, line 1: How did you separate these in the categories like this, is it done via some kind of process? This is a LOT of extra json, is it meant to be a replacement for our original emoji_strategy.json file?

components/emoji_picker/modules/emoji_selector.vue
```diff
@@ -0,0 +1,546 @@
...
+        <div
+          class="d-emoji-picker__tab"
+        >
+          <button
+            v-for="emoji in (emojis[tabs[index] + skinTone] ? emojis[tabs[index] + skinTone] : emojis[tabs[index]])"
+            :key="emoji.shortname"
+            type="button"
+            :aria-label="emoji.name"
+            @click="$emit('selected-emoji', emoji)"
+            @focusin="$emit('emoji-data', emoji)"
```
braddialpad: I find `emoji-data` kind of generic for an event name. does something like `highlighted-emoji` work?

braddialpad (commented):
Alright reviewed some of it, will do more tomorrow. Overall looks pretty good, fairly easy to understand event though it's a large amount of code. CSS needs some work to use proper vars and such but that can maybe happen when we move the CSS over to dialtone.

components/emoji_picker/modules/emoji_tabset.vue
```diff
@@ -0,0 +1,144 @@
...
+.d-emoji-picker{
+  &__tabset-list{
+      gap: 0;
+
+      &::after{
+        background-color: var(--black-200) !important;
+      }
+
+      button{
+        padding: var(--su8);
```
braddialpad: `var(--space-400);`

components/emoji_picker/modules/emoji_tabset.vue
iropolo: I tried to use dt-stack on the emoji_selector...but I didn't achieve the pixel perfect of design... Im still dealing with it so, I would prefer to keep it as this for now here also, I added this recommendation on [DT-1162](https://dialpad.atlassian.net/browse/DT-1162)

components/emoji_picker/modules/emoji_tabset.vue
iropolo: Yup seems like I updated the tab labels on the emoji_selector but not here in the emoji_tabset, we need it for the aria-label, lemme update it

components/emoji_picker/modules/emoji_tabset.vue
iropolo: Mhmm, now after the design update, we dont have animation, so... we dont have handleScroll and isScrolling anymore, maybe we can replace `isScrollingWithScrollTo` to `isScrolling` , what you think

components/emoji_picker/emoji_picker.vue
iropolo: It is not necessary I cleaned it

components/emoji_picker/emoji_picker.vue
iropolo: I updated it to required prop and I provided it from the story.
It will be documented in the storybook.

Now if you dont provide a tabSetLabels it will throw an error and not work.

components/emoji_picker/emojis/activity-dark.json
iropolo: Yes, Im not longer using emoji_strategy.json file.
I generated this 23 json files from the emoji_strategy.json.

And yes, I need to write the process I followed to achieve this categories.
I already have an internal READM.md inside the emoji-picker component for this.

I wrote 5 functions to work the emoji_strategy.json and get the necessaries emojis for the correct emoji_picker component functionality. Because for example emoji_strategy has emojis I will not use in this component (for example those which has 3 o 4 skin tone modifiers)

So basically, I have a large JSON file with 3559 objects inside processing that much data on the client-side is not ideal as it can slow down the performance of the component.

To avoid this, I split the original JSON into these 23 smaller JSON files.
By doing this, I can reduce the amount of data that needs to be processed by the client and improve the component's performance.

For example now the people json has only 465, the activity has 117... 

And Im just loading the necessary ones!
When the user change the skin, I will load those... Im not loading all at the beginning

This structure is necessary to avoid processing in the web.

I did this data modeling organizing and structuring in a way that is both efficient and effective to provide the component with the correct and necessary data and achieve better performance.

components/emoji_picker/modules/emoji_selector.vue
iropolo: Seems legit to me, will update it

components/emoji_picker/emoji_picker.vue
iropolo: Working it with Bradley, now should be fixed

components/emoji_picker/modules/emoji_tabset.vue
braddialpad: yep sounds good

components/emoji_picker/modules/emoji_tabset.vue
braddialpad: k no problem then.

components/emoji_picker/emojis/activity-dark.json
braddialpad: This is fantastic and I love it, however if we're going to do it we have to do the following before we merge this in:

- get rid of the import of `emoji_strategy.json` in `emoji.js` or else we are adding a huge amount of duplicated data to this bundle.
- replace all functions that were using `emoji_strategy.json` data to use the new categorized files instead. If we don't do this it is a breaking change.

Once we have the above two complete I think this should be good to merge in.

Additional things we can do, but they can be done later after this is merged:
- add your process that transforms `emoji_strategy.json` into to the repository and run it as part of the build process.
- start using emoji.js functions in the emoji picker code again so we can do things like custom emoji.

iropolo:
Design updates from [DT-1164](https://dialpad.atlassian.net/browse/DT-1164) are done.

[DT-1164]: https://dialpad.atlassian.net/browse/DT-1164?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

components/emoji_picker/emoji_picker.vue
iropolo: I didn't get what you mean here 🤔

braddialpad: `var(--bgc-moderate)` instead of `var(--black-200)`

components/emoji_picker/emojis/activity-dark.json
iropolo: Update:
We created `emojisIndexed` which replicate the emoji_strategy.json structure but with the updated emojis so it is not necessary to replace all function and also it is not necessary to import `emoji_strategy` anymore

braddialpad (approved):
Good to merge once you've fixed the remaining accessibility test issues. You can ignore the bundle size check as we spoke about.

components/emoji_picker/modules/emoji_tabset.vue
```diff
@@ -0,0 +1,154 @@
...
+    >
+      <template #tabs>
+        <dt-tab
+          v-for="tab in tabs"
+          :id="tab.id"
+          :key="tab.id"
+          :panel-id="tab.panelId"
+          :label="tab.label"
+          @click.capture.stop="selectTabset(tab.id)"
+        >
```
braddialpad: Can you override aria-controls to be your scrollable content element? That seems to be the remaining accessibility issue.

iropolo: Yay!, great, thanks, I didn't have a clue on what's even the problem lol.
Im guessing we sure put aria-controls required on dt-tab component or sth like that.

braddialpad (approved):

braddialpad:
Looks good now, I'll merge this since you aren't here today.

PR #842, opened 2023-03-20
fix: configure passthrough attrs in leftbar components

braddialpad:
# fix: configure passthrough attrs in leftbar components

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Passing in data-qa was not overriding the existing data-qa on the element. Made a change so it will now do this and added  passthrough props to group / contact rows as well. Needed to fix failing tests on master


braddialpad:
Merging now due to broken master.

PR #848, opened 2023-03-21
ci: fix visual tests

juliodialpad:
# CI: Fix Percy Visual Tests

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Updated Percy dependencies.
Enabled Javascript and disabled cache to give the snapshots a more consistent behavior.

## :bulb: Context

Percy visual tests were failing long ago but we didn't realized.

## :pencil: Checklist


percy.config.js
```diff
@@ -21,7 +21,9 @@ module.exports = {
     waitForTimeout: 250,
   },
   snapshot: {
-    widths: [2000],
-    minHeight: 2000,
+    enableJavaScript: true,
```
braddialpad, line 24: Is this a new option or something? Really surprised we didn't have this enabled before if not. Also surprised anything even worked at all? lol

juliodialpad: That option always existed, but we didn't needed at first because Storybook was "consistent" in some way with that options set to false (default value) but maybe in some version it stopped working.

And also `percy.config.js` wasn't being called (maybe after a percy dependencies update), so added the --config flag to be sure we are using the config file.

braddialpad (approved):
Great work! thanks for fixing this

PR #852, opened 2023-03-22
feat: adds codeblock component to dialtone vue

cjmollitor-dialpad:
# [DT-1015](https://dialpad.atlassian.net/browse/DT-1015): Adds `DtCodeblock` component to Dialtone Vue

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description
Adds a codeblock component to Dialtone Vue called `DtCodeblock`. The component takes one required prop `text` which is the preformatted text to be rendered inside the codeblock.

## :bulb: Context
For DP2.0, we're currently planning out the implementation for inline rich media in text messages ([link to tech spec](https://github.com/dialpad/web-clients/pull/400)). For most of these types of rich media, we're planning to create and use Dialtone Vue components. The designs and behavior for codeblocks are finalized so we decided to go ahead with implementing the Dialtone-side component that we'll use in DP2.0.

## :pencil: Checklist

- [x] I have added tests

## :crystal_ball: Next Steps
- [ ] Add syntax highlighting (TBD)

image

## :link: Sources
- [DP2.0 Inline Rich Media Tech Spec](https://github.com/dialpad/web-clients/pull/400)
- [Jira ticket](https://dialpad.atlassian.net/browse/DT-1015)

[DT-1015]: https://dialpad.atlassian.net/browse/DT-1015?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


braddialpad:
sweet, will check this out later today.

components/codeblock/codeblock.stories.js
```diff
@@ -0,0 +1,40 @@
...
+const DefaultTemplate = (args, { argTypes }) => createTemplateFromVueFile(
+  args,
+  argTypes,
+  DtCodeblockDefaultTemplate,
+);
+
+// Stories
+export const Default = DefaultTemplate.bind({});
+Default.args = {
+  text: 'function someFunction() {\n  return "some result";',
```
braddialpad: The example is an unclosed function block lol

components/codeblock/codeblock.vue
```diff
@@ -0,0 +1,20 @@
...
+<script>
+export default {
+  name: 'DtCodeblock',
+
+  props: {
+    text: {
+      type: String,
+      required: true,
+    },
+  },
```
braddialpad, line 19: Having this as a prop rather than a slot is totally fine for passing in user entered data, however it might be pretty awkward if you were using this for something like code examples on a documentation site. Imagine copy pasting your code example and then having to add a whole bunch of \n and such manually. Anyways not something we have to change now I don't think. Just thinking about future usage.

scripts/build-docs.js
```diff
@@ -15,6 +15,7 @@ const componentsList = [
   'checkbox.vue',
   'checkbox_group.vue',
   'chip.vue',
+  'codeblock.vue',
```
braddialpad, line 18: good job remembering this, lol

components/codeblock/codeblock.test.js
```diff
@@ -0,0 +1,77 @@
...
+    /*
+     * Test(s) to ensure that the component correctly handles user input
+     */
+  });
+
+  describe('Extendability Tests', function () {
+    /*
+     * Test(s) to ensure that the component can be correctly extended
+     */
+  });
```
braddialpad: Can just delete these

components/codeblock/codeblock.vue
```diff
@@ -0,0 +1,20 @@
+<template>
+  <pre>
+    <code
+      class="d-d-block d-bgc-white d-ws-pre-wrap d-p8 d-my4 d-ba d-bc-black-500 d-bar8 d-fs-100 d-lh-200 d-fc-red-400"
+    >{{ text }}</code>
```
braddialpad: Where did we actually get the styling for this? was it specified by design or just copied from the existing one in ubervoice?

braddialpad: If this is a core component we need to set the styles for this in dialtone and create a documentation page in dialtone (maybe you were already planning on doing this)

components/codeblock/codeblock.mdx
```diff
@@ -0,0 +1,51 @@
...
+
+```jsx
+import { DtCodeblock } from '@dialpad/dialtone-vue';
+```
+
+### Default
+
+```vue
+<dt-codeblock
+  text="'some pre-formatted codeblock'"
```
braddialpad: the inner quotes aren't necessary right?

components/codeblock/codeblock.mdx
```diff
@@ -0,0 +1,51 @@
...
+```js
+function someFunction() {
+  return "some result";
+```
+
+You'll have to do:
+
+```vue
+<dt-codeblock
+  text="'function someFunction() {\n  return "some result";\n}'"
```
braddialpad: the inner quotes aren't necessary right?

components/codeblock/codeblock.stories.js
cjmollitor-dialpad: Damn good catch lol. I'll close this up

components/codeblock/codeblock.vue
cjmollitor-dialpad: Totally agree. Might be a case of YAGNI but one reason I went with the text prop instead of a slot was for if/when we add syntax highlighting. [The way we currently do it](https://github.com/dialpad/firespotter/blob/0ea598ca7a57b5580692bc391031f2b227c2822c/ubervoice/static/js/single/components/feed/rich_media/feed_rich_media_highlight.vue#L3-L10), we format the text with _highlight.js_ and then `v-html` it into the codeblock component. We might be able to find alternative ways of doing this with a slot though. What are your thoughts?

scripts/build-docs.js
cjmollitor-dialpad: Bless the yeoman script lol

components/codeblock/codeblock.mdx
cjmollitor-dialpad: You're right, good catch! I'll change this

components/codeblock/codeblock.vue
cjmollitor-dialpad: > Where did we actually get the styling for this? was it specified by design or just copied from the existing one in ubervoice?

It was just copied from the existing one in ubervoice.

> If this is a core component we need to set the styles for this in dialtone and create a documentation page in dialtone (maybe you were already planning on doing this)

This actually completely slipped my mind. By that, you mean making a new Dialtone class like `d-codeblock` with all these classes baked in right?

francisrupert: @bhawkins-dialpad Will provide design guidance. For example, rather than `d-bgc-white` he'll share [which of these semantic colors](https://dialpad.design/design/colors/#surface) to use.

components/codeblock/codeblock.vue
braddialpad: Might want to think about using a highlight js vue library https://www.npmjs.com/package/vue-highlightjs

francisrupert: That would be awesome. If any library in play or adopted has some theming options, work with @bhawkins-dialpad to choose one that works best per his judgment, and identify a dark one to future-proof for dark mode.

components/codeblock/codeblock.vue
braddialpad: > By that, you mean making a new Dialtone class like `d-codeblock` with all these classes baked in right?

Yep exactly

components/codeblock/codeblock.vue
```diff
@@ -0,0 +1,20 @@
+<template>
+  <pre>
+    <code
+      class="d-d-block d-bgc-white d-ws-pre-wrap d-p8 d-my4 d-ba d-bc-black-500 d-bar8 d-fs-100 d-lh-200 d-fc-red-400"
```
bhawkins-dialpad: A couple styling suggestions
- Set the padding to only `d-p8` remove `d-my4`
- Set the font size to `d-fs-200`
- Set the line-height to `d-lh-400`
- I understand we will be looking at using a library for syntax highlighting but the fall back font color should = `d-fc-secondary`
- Border color = `d-bc-subtle`
- Background = `d-bgc-secondary`

cjmollitor-dialpad: Love the suggestions 👏 thanks Brad!
image

components/codeblock/codeblock.vue
```diff
@@ -0,0 +1,20 @@
+<template>
+  <pre>
+    <code
+      class="d-d-block d-bgc-secondary d-ws-pre-wrap d-p8 d-ba d-bc-subtle d-bar8 d-fs-200 d-lh-400 d-fc-secondary"
```
francisrupert: Add `d-ff-mono` (SF Mono).

It currently renders browser default `monospace`, and Dialtone CSS doesn't apply it to `<code>` element (which it probably should, but that can ultimately be reconciled later in `dialtone` repo.

components/codeblock/codeblock.vue
cjmollitor-dialpad: @bhawkins-dialpad  @francisrupert Looks like highlight.js has [a lot of good theming options for dark mode](https://highlightjs.org/static/demo/), I need to do some research into how we can conditionally choose a theme depending on if we're in dark mode or not.

@braddialpad good find with the Vue library! I saw in the readme for the one linked that it's been superseded by the new [native Highlight.js Vue plugin](https://github.com/highlightjs/vue-plugin). Do you have any objections to using that?

braddialpad: > I saw in the readme for the one linked that it's been superseded by the new [native Highlight.js Vue plugin](https://github.com/highlightjs/vue-plugin). Do you have any objections to using that?

Oh yeah, definitely use that.

francisrupert: Wow, so many options. Would you agree it probably doesn't matter what we choose now, as long as it's dead simple to pick another at any point? If so, don't consider this task blocked by that, and at another time the Dialtone Design team will explore what feels best with the Dialtone color palette. 

Personally, I've changed mine so many times but always come back to [Tomorrow night](https://highlightjs.org/static/demo/#Base16%20/%20Tomorrow%20Night).

braddialpad: Yeah the highlighting is not even part of this story, it will be a future update.

cjmollitor-dialpad: Good points. Just to confirm, is it ok to keep the interface as a text prop for now instead of a slot then?

braddialpad: Yep

cjmollitor-dialpad:
Discussed with @braddialpad offline about order of operations between this work and the `d-codeblock` Dialtone work.

I'm gonna do the following:
1. Finish up this PR and merge in
2. Make a PR in Dialtone to add the `d-codeblock` class
3. Modify this component to use `d-codeblock`

braddialpad (approved):

PR #858, opened 2023-03-24
fix(input): length validation improvement

braddialpad:
# fix(input): length validation improvement

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

A bit of an undesired behaviour found with length validation. If you set no warning but a max length, it still output a warning message when you are at the exact max length. I've modified this so that you can still have this behaviour if desired by setting `max` equal to `warn`. If warn is unset then a message will only display in error.

Also added a new prop retainWarning (default false) to keep a warning message displayed even after focusing out of the input. The current behaviour is to always show errors but remove warning messages on focusing out of the input. This may sometimes not be desired such as in vertically stacked input fields with a save button as shown in the below ticket. Because the warning message disappears on focus out, the height of the field changes and when you attempt to click the save button the save button moves out of your clicked area. Quite frustrating!

## :bulb: Context

Was mentioned in ticket https://dialpad.atlassian.net/browse/DP-67318 by the team working on bullhorn integration. Attempting to get this out for their release.

## :pencil: Checklist

- [x] I have added tests


juliodialpad (approved):
Nice work Brad! works perfectly.

PR #865, opened 2023-03-28
fix(image-viewer): dt image viewer visual testing

AlexLamDialpad:
# PR Title

Dt image viewer visual testing and typo fix

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Adds visual testing and fixes a typo
DP-67072
DT-994

## :pencil: Checklist

- [ ] I have added tests


components/image_viewer/image_viewer.mdx
```diff
@@ -40,5 +40,6 @@ import { DtImageViewer } from '@dialpad/dialtone-vue';
   image-button-class="imageButtonClass"
   aria-label="ariaLabel"
   close-aria-label="closeAriaLabel"
+  force-close="forceClose"
```
braddialpad: forceClose?

components/image_viewer/image_viewer.vue
```diff
@@ -120,6 +120,14 @@ export default {
...
       required: true,
     },
+
+    /**
+     * Forces the image to stay open
+     */
+    forceOpen: {
+      type: Boolean,
+      default: null,
+    },
```
braddialpad: If we're going to do this.. we might as well just put in open.sync (or show.sync?) functionality. This will not only help for these visual tests but also for any developer that needs to trigger this in a different way.

see `dropdown.vue` for an example of this. It would have default behaviour if open is null (open on click), but if set by the consumer it would disable the default open behaviour and allow the consumer to control it instead.

https://v2.vuejs.org/v2/guide/components-custom-events.html#sync-Modifier

The equivalent in vue 3 is `v-model:open `

braddialpad (changes requested):

components/image_viewer/image_viewer.stories.js
```diff
@@ -133,6 +147,7 @@ Default.args = {
   closeAriaLabel: 'Close',
   imageButtonClass: 'd-wmn64 d-hmn64 w-wmx332 d-hmx332',
   ariaLabel: 'Click to open image',
+  open: false,
```
braddialpad: Hmm it doesn't seem to like this... it is immediately opening on the main (non-visual-testing) page. I think we're going to have to create a separate instance of the default template in the visual testing story instead of copying it from this one.

components/image_viewer/image_viewer.stories.js
```diff
@@ -97,6 +100,17 @@ export const argTypesData = {
...
+  },
+
+  'update:open': {
+    table: {
+      disable: true,
+    },
+  },
 };
 
 // Story Collection
```
braddialpad, line 128: Make parameters like this

```
parameters: {
  controls: {
    sort: 'requiredFirst',
  },
  docs: {
    page: DtImageViewerMdx,
  },
  options: {
    showPanel: true,
  },
},
```

components/image_viewer/image_viewer.vue
```diff
@@ -171,27 +186,39 @@ export default {
...
+
+      immediate: true,
+    },
   },
 
   methods: {
-    open () {
-      this.show = true;
+    openModal () {
+      this.isOpen = true;
```
braddialpad, line 216: This should only happen if `this.isOpen` is null. If `true` or `false` the default open behaviour (click the image to open) has been overridden by the consumer so it should not happen anymore.

components/image_viewer/image_viewer.vue
```diff
@@ -171,27 +186,39 @@ export default {
...
+      this.$emit('update:open', true);
```
braddialpad: in addition to update:open you should also trigger a normal "open" event that is documented and meant to be captured by the user.

braddialpad (changes requested):

components/image_viewer/image_viewer.vue
AlexLamDialpad: So disallow open but allow open? We'd never emit `'update:open', true` then right

braddialpad: `update:open` - only for .sync, only emitted when open is not null.
`open` - always emitted whether open is null or not.

see: https://github.com/dialpad/dialtone-vue/blob/staging/components/popover/popover.vue#L790-L793

braddialpad (approved):
Looks good now, sorry for the pain!

PR #868, opened 2023-03-29
feat(tooltip): add 300ms default delay

braddialpad:
# feat(tooltip): add 300ms default delay

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Set 500ms delay on tooltip and added delay prop to enable / disable delay. We are intentionally not allowing users to set an exact ms for delay. Had to code delay manually since we aren't using tippy for trigger events but it was simple enough 🤷‍♂️ .

## :bulb: Context

It was a design decision to make this change since the tooltips in product are displaying much too fast. It is still possible to set delay = false but that will not be the default going forward.

## :pencil: Checklist

## :crystal_ball: Next Steps

Test with designers to see if the delay feels reasonable.

## :link: Sources

https://www.nngroup.com/articles/timing-exposing-content/

### DIFF
--- a/components/tooltip/tooltip.test.js
+++ b/components/tooltip/tooltip.test.js
@@ -48,6 +48,9 @@ describe('DtTooltip tests', function () {
         // this gets around transition async problems. See https://v1.test-utils.vuejs.org/guides/common-tips.html
         transition: transitionStub(),
       },
+      propsData: {
+        delay: false,
+      },
       slots: {
         default: 'Test message',
       },
@@ -158,6 +161,7 @@ describe('DtTooltip tests', function () {
 
       describe('When mouseenter tooltip', function () {
         beforeEach(async function () {
+          await wrapper.setProps({ delay: false });
           await anchor.trigger('mouseenter');
         });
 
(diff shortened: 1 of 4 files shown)


juliodialpad (commented):
The code looks good, but when you leave the anchor, the element waits for the delay to hide the tooltip, is this intended? It doesn't feel right to me.

juliodialpad:
Here's an example of what I'm saying, toggling the `delay` prop.

https://user-images.githubusercontent.com/87546543/228630212-5f57b25c-4966-45fc-8af6-0ff5e5ecda60.mov

braddialpad:
> The code looks good, but when you leave the anchor, the element waits for the delay to hide the tooltip, is this intended? It doesn't feel right to me.

Just made a change to update this, should be there in the latest preview

juliodialpad:
> > The code looks good, but when you leave the anchor, the element waits for the delay to hide the tooltip, is this intended? It doesn't feel right to me.
> 
> Just made a change to update this, should be there in the latest preview

Looks like the GHA are experimenting delays, this is still queued 7 minutes ago https://github.com/dialpad/dialtone-vue/actions/runs/4556393648.

braddialpad:
Yep they're definitely having problems.

juliodialpad (approved):
The preview got updated and looks good to me now, thanks Brad!

PR #877, opened 2023-03-31
feat(recipe-leftbar-unread-chip): add leftbar unread chip recipe

juliodialpad:
# Add left bar unread chip recipe

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

- Added Leftbar to yeoman generator options as we are generating a lot of this recipes
- Fixed a bug found on chip that labelClass prop was only being applied to grouped chips
- Updated caniuse-lite version
- Added Leftbar unread chip recipe

## :bulb: Context

We are creating recipe components to update leftbar look and feel.
This PR includes unread mentions/messages chip recipe

## :pencil: Checklist

- [x] I have added tests

## :crystal_ball: Next Steps

Update this on product

Screenshot 2023-03-30 at 7 43 28 p m

## :link: Sources

https://www.figma.com/file/oTY56LgQha8ljb4c8w4GtO/DT-Shared%3A-App-Components?node-id=9240%3A7131&t=YyektqRcqv0aK6qd-0


braddialpad:
Hmm looks like we might have to make a change to handle the threads case too. Just got notified of this: https://dialpad.atlassian.net/browse/DP-53387

braddialpad:
> Hmm looks like we might have to make a change to handle the threads case too. Just got notified of this: https://dialpad.atlassian.net/browse/DP-53387

Actually.. I think this will already work the way it is since the text is a slot

recipes/leftbar/unread_chip/unread_chip.vue
```diff
@@ -0,0 +1,87 @@
...
+      />
+    </template>
+    <span data-qa="dt-leftbar-unread-chip__label">
+      <slot />
+    </span>
+  </dt-chip>
+</template>
+
+<script>
+import { DtChip, DtIcon } from '@';
```
braddialpad: don't use '@' we had some weird circular dependency issues with that at some point, use `@/components/chip` instead. I know it's a bit less convenient

recipes/leftbar/unread_chip/unread_chip.vue
```diff
@@ -0,0 +1,87 @@
+<template>
+  <dt-chip
+    class="dt-leftbar-unread-chip"
+    :class="`dt-leftbar-unread-chip__${kind}`"
```
braddialpad: ``:class="['dt-leftbar-unread-chip', `dt-leftbar-unread-chip__${kind}`]``

braddialpad (commented):
Looks great! couple small things.

juliodialpad:
> > Hmm looks like we might have to make a change to handle the threads case too. Just got notified of this: https://dialpad.atlassian.net/browse/DP-53387
> 
> Actually.. I think this will already work the way it is since the text is a slot

If it is just text saying "Unread threads" it will work, just need to add that logic into the product, to decide the priority eg.

mentions > threads > messages ?
mentions > messages > threads ?

Or should we show multiple pills?

recipes/leftbar/unread_chip/unread_chip.vue
juliodialpad: Right, I'll change this and have that in mind in the future, thanks!

braddialpad:
> If it is just text saying "Unread threads" it will work, just need to add that logic into the product, to decide the priority eg.
> 
> mentions > threads > messages ? mentions > messages > threads ?
> 
> Or should we show multiple pills?

The threads display on a different view than the messages / mentions so we don't have to worry about this.

braddialpad (approved):

PR #879, opened 2023-03-31
fix: dialtone-icons dependency

juliodialpad:
# Fix dialtone-icons dependency

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Removed dialtone-icons from dependencies and moved into devDependencies and peerDependencies.

## :bulb: Context

Errors are being reported around dialtone-icons not found on clients using dialtone-vue.

## :pencil: Checklist

- [ ] I have added tests

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,6 @@
       "version": "2.74.2",
       "license": "MIT",
       "dependencies": {
-        "@dialpad/dialtone-icons": "0.3.1",
         "@linusborg/vue-simple-portal": "^0.1.5",
         "emoji-regex": "^10.2.1",
         "emoji-toolkit": "^6.6.0",
@@ -21,6 +20,7 @@
         "@commitlint/config-conventional": "^17.4.2",
         "@dialpad/conventional-changelog-angular": "^1.1.1",
         "@dialpad/dialtone": "^7.22.0",
+        "@dialpad/dialtone-icons": "0.3.1",
         "@dialpad/semantic-release-changelog-json": "^1.0.0",
         "@percy/cli": "^1.21.0",
         "@percy/storybook": "^4.3.5",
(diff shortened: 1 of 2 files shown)


braddialpad (commented):
This will work but I'm still questioning whether dialtone-icons should be a peer dependency or not since DtIcon is part of the package and won't run without it. I guess the same is true of dialtone which is also a peer dependency...

Another option is to separate a different build target like we do for emoji. How big is the icons bundle in relation to the dialtone-vue package? If very large I think that will be the best option.

juliodialpad:
> This will work but I'm still questioning whether dialtone-icons should be a peer dependency or not since DtIcon is part of the package and won't run without it. I guess the same is true of dialtone which is also a peer dependency...
> 
> Another option is to separate a different build target like we do for emoji. How big is the icons bundle in relation to the dialtone-vue package? If very large I think that will be the best option.

Dialtone-vue
Screenshot 2023-03-31 at 7 55 43 p m

Dialtone-icons
Screenshot 2023-03-31 at 7 54 20 p m

It almost as big a Dialtone-vue.

Why the peerDependency is not a good option? I think it would work perfectly fine and will not increase bundle size.

braddialpad:
> Why the peerDependency is not a good option? I think it would work perfectly fine and will not increase bundle size.

It only decreases bundle size if you are NOT using dialtone-icons, which most projects will be.

If we create a separate build target for icons like we did for emoji it would:

- still not increase the bundle size of the core package as it's in a different file
- consumer does not have to be responsible for dialtone-icons versioning
- (most importantly) the icons bundle could be dynamically loaded in the app when it needs to be used rather than all loaded up front.

braddialpad:
Think we can just close this one it's working fine in firespotter again. Can create additional tickets for bundle size / code splitting improvements.

juliodialpad:
Agree, closing this for now.

PR #880, opened 2023-03-31
fix: dialtone-icons dependency

juliodialpad:
# Vue 3 version of #879


braddialpad:
Think we can just close this one it's working fine in firespotter again.

juliodialpad:
Agree, closing

PR #889, opened 2023-04-04
fix(recipe-contact-info): remove d-of-hidden

juliodialpad:
# Fix Recipe Contact Info

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Removed `d-of-hidden` class from default_list_item as I don't see any reason to keep it and it is breaking the behavior if it contains a popover anchor inside it.

## :bulb: Context

This is related with https://github.com/dialpad/firespotter/pull/27100, the real issue is that contact-info has a `d-of-hidden` class that prevents the popover anchor element to have the required space to display properly.

## :pencil: Checklist

- [ ] I have added tests


braddialpad (commented):
I agree with this but I'm a bit worried about it breaking something in product. Can we migrate this to dialpad and check instances of dt-list-item?

juliodialpad:
> I agree with this but I'm a bit worried about it breaking something in product. Can we migrate this to dialpad and check instances of dt-list-item?

Sure, I'll migrate and test before merging

braddialpad (approved):

juliodialpad:
Tested all over the places where `DtListItem` was being used and everything looks fine. I think we are safe, merging this.

MikaDialpad:
This was originally added to enable truncation of the content https://github.com/dialpad/dialtone-vue/pull/604. It's still possible when using a custom type, but we could consider adding this conditionally as a prop or use a generic `contentClass` or something to allow passing any classes to this element.

EDIT: Ugh I forgot that there's no interface currently to the default list item as it's being wrapped by the DtListItem, so the suggestion above probably wouldn't work...

PR #888, opened 2023-04-04
feat(modal): add closeOnClick prop

braddialpad:
# feat(modal): add closeOnClick prop

## :hammer_and_wrench: Type Of Change

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

If this prop is false it will prevent closing the modal when the overlay is clicked on

## :bulb: Context

Requested by dev

PR #890, opened 2023-04-04
feat(modal): add closeOnClick prop

braddialpad:
vue 3 https://github.com/dialpad/dialtone-vue/pull/888

### DIFF
--- a/components/modal/modal.vue
+++ b/components/modal/modal.vue
@@ -290,6 +290,15 @@ export default {
     },
 
     /**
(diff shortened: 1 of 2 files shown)


components/modal/modal.vue
```diff
@@ -1,3 +1,4 @@
+
```
juliodialpad: We can remove this new line at the beginning

juliodialpad (changes requested):

PR #893, opened 2023-04-04
docs: add rich text editor tech spec

MikaDialpad:
# docs: add rich text editor tech spec

- [ ] Fix
- [ ] Feature
- [ ] Refactoring
- [x] Documentation

## :book: Description

This is the tech spec for the Rich Text Editor component. It lived originally in the [web-clients repo](https://github.com/dialpad/web-clients/pull/405) from where it was decided to be moved into this repo. The Editor component part has been [approved](https://github.com/dialpad/web-clients/pull/418) and work on the implementation can be started. The extensions section still needs more details and will be worked on after the Editor implementation

## :bulb: Context

We need a Rich Text Editor component for Dialpad 2.0 and most likely for 1.0 as well. We've done a [decision spike](https://github.com/dialpad/web-clients/blob/main/docs/decisions/0004-rich-text-editor.md) earlier and the outcome was that we'll want to use TipTap as the base solution.

## :crystal_ball: Next Steps

We're trying a new tech spec process in DP 2.0 where we divide them into multiple steps, so we could start working on the parts that are already approved. In this specific tech spec the Editor parts have been approved, so we can start building the basic editor while working on the next tech spec, which would in this case be for the link extension. We'll open a separate PR for the new changes so they're easier to review in isolation.

## :link: Sources

https://tiptap.dev/
https://github.com/dialpad/web-clients/blob/main/docs/decisions/0004-rich-text-editor.md

### DIFF
--- a/components/rich_text_editor/rich_text_editor.tech_spec.md
+++ b/components/rich_text_editor/rich_text_editor.tech_spec.md
@@ -1,0 +1,343 @@
+# Dialtone Rich Text Editor
+
+**Originally published:** 16-03-2023
+
+**Original author:** Mika Paulasaari
+
+# Overview
+
+We have identified a need for a rich text editor component. After investigating
+our options we [decided](docs/decisions/0004-rich-text-editor.md) to move
+forward with TipTap, which is a 3rd party rich text editor. This document
+outlines the high level plan for how we can componentize the TipTap editor into
+a Dialtone component and how we can create custom extensions for it using other
+Dialtone components.
+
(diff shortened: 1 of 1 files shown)


MikaDialpad:
Welp, looks like Github added you all as reviewers automatically. There's no action required here at the moment though, but this will be a living document for a while (unless we separate the extensions into their own tech specs?). I can convert this to draft for now.

braddialpad (approved):

PR #896, opened 2023-04-06
fix(popover): losing focus

juliodialpad:
# Fix popover losing focus

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Removed `this.$nextTick()` as we are already waiting for the nextTick on `preventScrolling`.

## :bulb: Context

This behavior was firing the 'opened' event until 2 ticks, so if the client was depending on this to keep the anchor visible, it will hide and close the popover immediately.

## :pencil: Checklist

### DIFF
--- a/components/popover/popover.vue
+++ b/components/popover/popover.vue
@@ -732,8 +732,7 @@ export default {
     *   - when anchor is within another popover: set the popover dialog container to it's non-modal z-index
     *     since it is no longer the active modal. This puts it underneath the overlay and prevents scrolling.
     **/
-    async preventScrolling () {
-      await this.$nextTick();
+    preventScrolling () {
       if (this.modal) {
         const element = this.anchorEl.closest('body, .tippy-box');
         if (element.tagName.toLowerCase() === 'body') {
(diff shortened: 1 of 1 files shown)


braddialpad (approved):
Nice, good find

components/popover/popover.vue
```diff
@@ -783,11 +783,9 @@ export default {
     async onLeaveTransitionComplete () {
       if (this.modal) {
         await this.focusFirstElement(this.$refs.anchor);
-        // await next tick in case the user wants to change focus themselves.
-        await this.$nextTick();
       }
-      this.tip?.unmount();
       await this.enableScrolling();
```
braddialpad: Hmm do we maybe want to pull the nextTick out of enableScrolling so it's more visible?

juliodialpad: Sounds good, I'll change it

braddialpad:
ah damn looks like you set the no-visual-test label but it still ran the visual tests.. I'll fix that

juliodialpad:
> ah damn looks like you set the no-visual-test label but it still ran the visual tests.. I'll fix that

It is because the `visual-test-ready` was added, right? you should only add that that if it doesn't have the 'no-visual-test' tag

braddialpad:
> > ah damn looks like you set the no-visual-test label but it still ran the visual tests.. I'll fix that
> 
> It is because the `visual-test-ready` was added, right? you should only add that that if it doesn't have the 'no-visual-test' tag

Yeah exactly, that logic is in there but looks like it's not working correctly.

braddialpad (approved):

PR #898, opened 2023-04-06
chore: fix more visual test issues

braddialpad:
# chore: fix more visual test issues

## :book: Description

- add a seed to avatar presence story since it seems to be causing tiny 1 pixel visual diffs around the edges.
- change labelling logic to not run if no-visual-test is set.


juliodialpad (approved):
Nice, that's what it was missing, thanks Brad!

juliodialpad (approved):

PR #903, opened 2023-04-11
fix: null protection on some refs

braddialpad:
Some additional null protection on some refs fields, since these seem to be triggering errors in a lot of cases likely because the ref element no longer exists.

### DIFF
--- a/components/popover/popover.vue
+++ b/components/popover/popover.vue
@@ -825,7 +825,7 @@ export default {
         console.warn('Could not find the element specified in dt-popover prop "initialFocusElement". ' +
           'Defaulting to focusing the dialog.');
       }
-      result ? result.focus() : this.$refs.content.$el.focus();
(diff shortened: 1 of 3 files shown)


juliodialpad (approved):
Changes look good, but Deploy preview failed 🤔

braddialpad:
> Changes look good, but Deploy preview failed 🤔

```
! [remote rejected] github-pages-deploy-action/frvc4wxn2 -> gh-pages (cannot lock ref 'refs/heads/gh-pages': is at af0a4afba46dd4a8cafab1a35b263e11d9a1d2eb but expected e61a11805342c11a0ecfe0d03f4701e15adcb323)
```

Hmm might be a race condition type of thing when it's deploying two previews at the same time.

braddialpad:
Worked after re-running.

PR #905, opened 2023-04-11
feat(rich-text-editor): add rich text editor component

MikaDialpad:
# feat(rich-text-editor): add rich text editor component

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [x] Documentation

## :book: Description

Create a Rich Text Editor component. There is a [tech spec](https://github.com/dialpad/dialtone-vue/blob/staging/components/rich_text_editor/rich_text_editor.tech_spec.md) that this implementation follows, so hopefully there aren't many surprises. To make the code review easier tests are coming in a separate PR.

This is not on purpose included in the root index since we don't want TipTap in the code bundle and this component shouldn't be used as standalone anyways. We'll build a Message Input component that uses/wraps this component and that will be built into a separate build target like the emoji.

## :pencil: Checklist

- [ ] I have added tests

## :crystal_ball: Next Steps

- Add tests
- Create a Link extension
- Create Message Input component to use this

image
image
image

## :link: Sources

https://dialpad.atlassian.net/browse/DT-1076
https://github.com/dialpad/dialtone-vue/blob/staging/components/rich_text_editor/rich_text_editor.tech_spec.md
https://tiptap.dev


components/rich_text_editor/rich_text_editor.vue
```diff
@@ -0,0 +1,272 @@
...
+        attrs['aria-readonly'] = true;
+      }
+      return attrs;
+    },
+
+    inputListeners () {
+      return {
+        focusin: event => this.$emit('focusin', event),
+        focusout: event => this.$emit('focusout', event),
+      };
```
MikaDialpad: These work in the Vue 3 version, but not in Vue 2... Not sure what's going on, but it might be some difference in how EditorContent has been implemented in each version..? Or does someone have other ideas? I'm also not super sure how important these events are since we have focus/blur, so maybe we can omit them?

components/rich_text_editor/rich_text_editor.mdx
```diff
@@ -0,0 +1,42 @@
...
+  variety of styles. It enables also mixing text with custom components built as
+  extensions, such as links and user mentions.
+</Subtitle>
+
+<div class="d-notice d-notice--info d-wmx100p">
+<b>NOTICE:</b> This component is not a part of Dialtone-Vue core package and can
+only be used by other Dialtone components. To keep this code out of the
+Dialtone-vue core any Dialtone component that uses this should be built into
+their separate build targets.
+</div>
```
MikaDialpad, line 16: I wanted this to be extra obvious lol

braddialpad: yes very good, haha

components/rich_text_editor/rich_text_editor.vue
braddialpad: I think we can remove these if we already have focus and blur

package.json
```diff
@@ -84,6 +84,11 @@
   "dependencies": {
     "@dialpad/dialtone-icons": "^0.6.0",
     "@linusborg/vue-simple-portal": "^0.1.5",
+    "@tiptap/extension-document": "^2.0.2",
+    "@tiptap/extension-paragraph": "^2.0.2",
+    "@tiptap/extension-text": "^2.0.2",
+    "@tiptap/pm": "^2.0.2",
+    "@tiptap/vue-2": "^2.0.2",
```
MikaDialpad, line 91: I submitted TipTap for a OneTrust review April 6 and what I've seen it takes a few weeks to get approved. Since this wont be bundled in the core package anyways I think we can merge even before the OT approval. It has MIT license, so I'm not expecting any issues to get it approved.

components/rich_text_editor/rich_text_editor_default.story.vue
```diff
@@ -0,0 +1,35 @@
+<template>
+  <dt-rich-text-editor
+    v-model="inputValue"
+    :editable="editable"
+    :input-aria-label="inputAriaLabel"
+    :output-format="outputFormat"
```
MikaDialpad, line 7: I tried to do the same thing for `autofocus` but it wouldn't work for one reason or another. Like the storybook would even strip this param from the URL when set, so I must be missing something. I'll have to look into it later unless someone knows off the bat what the issue is. At least the linter didn't like the `autofocus` name since it assumes it's the native one, so might need to change that to `auto-focus` or something.

components/rich_text_editor/rich_text_editor.vue
```diff
@@ -0,0 +1,250 @@
...
+     * Editor input event
+     * @event input
+     * @type {String}
+     */
+    'input',
+
+    /**
+     * Event to sync the value with the parent
+     * @event input
+     * @type {String}
```
iropolo: Is this correct?

components/rich_text_editor/rich_text_editor.vue
```diff
@@ -0,0 +1,250 @@
...
+    };
+  },
+
+  computed: {
+    extensions () {
+      // These are the default extensions needed jsut for plain text.
+      const extensions = [Document, Paragraph, Text];
+      // TODO: Add logic for extensions passed as props.
+      return extensions;
+    },
```
iropolo, line 139: could we just 

`return [Document, Paragraph, Text]`;

or there is sth Im missing..?

components/rich_text_editor/rich_text_editor.vue
```diff
@@ -0,0 +1,250 @@
...
+      const attrs = {
+        'aria-label': this.inputAriaLabel,
+        'aria-multiline': true,
+        role: 'textbox',
+      };
+      if (!this.editable) {
+        attrs['aria-readonly'] = true;
+      }
+      return attrs;
+    },
```
iropolo, line 151: Code suggestion 

```
return {
    'aria-label': this.inputAriaLabel,
    'aria-multiline': true,
    role: 'textbox',
    ...(this.editable ? {} : {'aria-readonly': true})
  };
```

components/rich_text_editor/rich_text_editor.vue
```diff
@@ -0,0 +1,250 @@
...
+      // These are the default extensions needed jsut for plain text.
+      const extensions = [Document, Paragraph, Text];
+      // TODO: Add logic for extensions passed as props.
+      return extensions;
+    },
+
+    inputAttrs () {
+      const attrs = {
+        'aria-label': this.inputAriaLabel,
+        'aria-multiline': true,
```
braddialpad, line 144: Should this always be true or is it possible to have a "single line mode"

components/rich_text_editor/rich_text_editor.vue
```diff
@@ -0,0 +1,250 @@
...
+      default: true,
+    },
+
+    /**
+     * Descriptive label for the input element
+     */
+    inputAriaLabel: {
+      type: String,
+      required: true,
+    },
```
braddialpad, line 48: Do we also need to support labelled by?

components/rich_text_editor/rich_text_editor.vue
MikaDialpad: Good point, when JSON is the output format it's not

components/rich_text_editor/rich_text_editor.vue
```diff
@@ -0,0 +1,250 @@
...
+     * @type {String|JSON}
+     */
+    'input',
+
+    /**
+     * Event to sync the value with the parent
+     * @event input
+     * @type {String|JSON}
+     */
+    'update:value',
```
braddialpad, line 110: Will a `.sync` actually work here? the output is completely different than what's being passed in right?

braddialpad (commented):
Leaving a few comments, have to head into the office now but I'll review more later.

iropolo:
Im kind of lost here... where is the rich text? cos what I can do here is just overwrite a string 🤔

braddialpad:
> Im kind of lost here... where is the rich text? cos what I can do here is just overwrite a string 🤔 

You may want to look at the previously done tech spec to get an idea of what the requirements are here

https://github.com/dialpad/dialtone-vue/pull/893

MikaDialpad:
> > Im kind of lost here... where is the rich text? cos what I can do here is just overwrite a string 🤔
> 
> You may want to look at the previously done tech spec to get an idea of what the requirements are here
> 
> #893

Sorry for being brief on the context here, I've been mostly chatting with Brad about this so far. So initially this will not include any rich text features really, but it's more or less just a textarea that uses TipTap. Once this is merged we'll start working on some of the rich text features, like a link extension (and eventually emoji, user mention, channel hashtag, codeblock)

iropolo:
> > > Im kind of lost here... where is the rich text? cos what I can do here is just overwrite a string 🤔
> > 
> > 
> > You may want to look at the previously done tech spec to get an idea of what the requirements are here
> > #893
> 
> Sorry for being brief on the context here, I've been mostly chatting with Brad about this so far. So initially this will not include any rich text features really, but it's more or less just a textarea that uses TipTap. Once this is merged we'll start working on some of the rich text features, like a link extension (and eventually emoji, user mention, channel hashtag, codeblock)

Ye, the title sounds confusing to me

components/rich_text_editor/rich_text_editor.stories.js
```diff
@@ -0,0 +1,113 @@
...
+// Default Prop Values
+export const argsData = {
+  value: `I’m not a standalone component, please use Message Input instead 🙏!`,
+  editable: true,
+  inputAriaLabel: 'This is a descriptive label',
+  onBlur: action('blur'),
+  onInput: action('input'),
+  onFocus: action('focus'),
+  onFocusIn: action('focusin'),
+  onFocusOut: action('focusout'),
```
braddialpad: Remove these too

components/rich_text_editor/rich_text_editor_default.story.vue
```diff
@@ -0,0 +1,35 @@
+<template>
+  <dt-rich-text-editor
```
braddialpad, line 2: missing `inputClass` prop here.

components/rich_text_editor/rich_text_editor.vue
```diff
@@ -0,0 +1,250 @@
...
+        // through the parent, so don't do anything here.
+        return;
+      }
+      // Otherwise replace the content (resets the cursor position).
+      this.editor.commands.setContent(newValue, false);
+    },
+  },
+
+  mounted () {
+    this.createEditor();
```
braddialpad, line 185: Should be possible to do this on created rather than mounted right since the editor-content component is what's binding to the DOM not the editor itself? Not that big of deal but might give a slight performance increase.

components/rich_text_editor/rich_text_editor.vue
braddialpad: Hmm see you already thought about this in the value watch

```
if (newValue === this.getOutput()) {
  // The new value came from this component and was passed back down
  // through the parent, so don't do anything here.
  return;
}
```

Not sure I love this, since it's not really how v-model / sync is expected to work and at this point you're not really "syncing" anything

Maybe we have `input` / `update:value` events ALWAYS output as text, and have a different event for the formatted output

components/rich_text_editor/rich_text_editor.vue
braddialpad: Think he's probably just preparing it for the extensions passed as props that will have to be merged into that array.

components/rich_text_editor/rich_text_editor.vue
MikaDialpad: > Will a `.sync` actually work here? the output is completely different than what's being passed in right?

This will expect that the input is in the format that the output is. `v-model` works here and AFAIK it's the same thing as .sync except defaulting to the `value` prop. Not sure how useful it really is to sync the value from the parent except when setting/resetting the initial value for this.

> Not sure I love this, since it's not really how v-model / sync is expected to work and at this point you're not really "syncing" anything

Hmm, maybe the comment is unclear here, but what this early return is doing is just stopping the value from the editor from being passed straight back to it since it would reset the cursor position.
- Editor -> `this.$emit('update:value', '<p>hello</p>')`
- Parent -> local value changes from `<p>hell</p>` to `<p>hello</p>` and as the value here has changed it gets passed right down to the Editor again
- Editor receives `<p>hello</p>`, which matches what the value already is, so there's no need to set it again

Or am I misunderstanding something? I was going back and forth whether to include this in the first place, since the Editor really should be the source of truth for the value. So instead of `value` we could have a prop `initialValue` that makes it clear that it resets the value of the Editor and it wouldn't support v-model, but to get the new value you'd have `@input` or `@output` or whatever

braddialpad: alright so I think what you're saying is whatever format is output, the library input field will also take as input?

ex if you pass `<p>hello</p>` as the input value it only displays as `hello` in the input itself, not `<p>hello</p>`

MikaDialpad: Right, the editor will render any HTML passed to it as a value, but you can't actually type HTML tags in the text as they'll be escaped. The value you pass to it is not actually limited to the selected output format, but for practical reasons they should probably match. Like I can choose to output JSON and pass HTML as the initial value and the editor will not complain.

braddialpad: alright, I think this is okay then.

components/rich_text_editor/rich_text_editor.vue
iropolo: Im not sure about the performance since it will be negligible probably, but Im agree on have it separated just to achive good practice (anyways It is not wrong to have it on `mounted`), because as you said this createEditor() is not related DOM.

On vue 3 we dont have this created hook, it was replaced with setup... if you are writing the component with <script setup> syntax (which I highly recommend) you can place it directly there... if not you could use `onBeforeMount` instead

MikaDialpad: Yeah fair enough, moved this to created() instead

components/rich_text_editor/rich_text_editor.vue
MikaDialpad: Whoops missed these comments. For now we only know of the message input use case, so I'd keep this as is.

components/rich_text_editor/rich_text_editor.vue
MikaDialpad: Hmm, I'm not sure. What I'm seeing is that usually the element that labels something is built into the component as a header or whatever. In this case it'd be an external element ID that would have to be passed, but I feel like it'd be easier to pass a string for the aria-label instead :?

braddialpad: alright, not really a big deal it can be easily added later if needed.

braddialpad (approved):
Looks good by me!

PR #881, opened 2023-03-31
feat: support multiple avatars for `DtRecipeContactInfo` Component

SheldonDialpad:
# PR Title

Support multiple avatars for `DtRecipeContactInfo` Component

## :hammer_and_wrench: Type Of Change

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :pencil: Checklist

- [x] I have reviewed my changes
- [x] I have added all relevant documentation
- [x] I have validated components with a screen reader
- [x] I have considered the performance impact of my change
- [x] I have checked that my change did not significantly increase bundle size

## :crystal_ball: Next Steps

Update Application UI with multiple avatars feature.

## :camera: Screenshots / GIFs

Screen Shot 2023-03-31 at 12 36 47 AM
Screen Shot 2023-03-31 at 12 37 57 AM

## :link: Sources

Jira: [DP-66880](https://dialpad.atlassian.net/browse/DP-66880)

[DP-66880]: https://dialpad.atlassian.net/browse/DP-66880?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #906, opened 2023-04-12
fix(avatar): remove box sizing for overlay text/icon

SheldonDialpad:
# PR Title

remove box sizing for overlay text/icon

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

In our application, there is `box-sizing` style in `commonextra.css` which impact all width calculation if there is border style. 
Screen Shot 2023-04-12 at 12 23 47 PM

This will cause our overlay text/icon fail to cover the avatar.
Screen Shot 2023-04-12 at 12 23 35 PM

To solve this issue, we need to force remove `box-sizing` style in overlay. So we add `d-box-unset` to overlay div.

## :pencil: Checklist

After remove `box-sizing` style in overlay div.
Screen Shot 2023-04-12 at 12 37 31 PM

## :link: Sources

This change will apply to https://dialpad.atlassian.net/browse/DP-66893


juliodialpad (changes requested):
This seems pretty specific for your use-case, don't you think this could potentially break something while other users implement it?

Wouldn't it be better to add an `overlayClasses` prop so you can add custom classes to that overlay?

juliodialpad:
Two more comments:

1. Make sure you're following the commit convention described here: https://github.com/dialpad/dialtone-vue/blob/staging/.github/COMMIT_CONVENTION.md - PR title should be something like: `fix(avatar): remove box sizing for overlay text/icon

2. In this PR: https://github.com/dialpad/dialtone-vue/pull/881 we forgot to add `overlayIcon` and `overlayText` props to `avatar_default.story.vue` so that props currently cannot be tested on the avatar page: https://vue.dialpad.design/deploy-previews/pr-906/?path=/story/components-avatar--initials&args=overlayText:
  - Can you please add that props to the file so we can properly test the behavior there?
  - Also add the list of icons instead of raw text for `overlayIcon` please, you can see an example of that implementation here: https://github.com/dialpad/dialtone-vue/blob/staging/components/icon/icon.stories.js#L15-L24

iropolo:
On code perspective LGTM

components/avatar/avatar.vue
```diff
@@ -220,6 +227,13 @@ export default {
       ];
     },
 
+    overlayClasses () {
+      return [
+        'd-bgc-black-900 d-o70 d-ps-absolute d-w100p d-h100p d-d-flex d-ai-center d-zi-base',
```
juliodialpad:
```suggestion
        'd-bgc-black-900 d-o70 d-ps-absolute d-w100p d-h100p d-d-flex d-ai-center d-zi-base1 d-bar-circle',
```

The z-index needs to be 'd-zi-base1' in order to be able to override the avatarImage in certain cases.
Also we need d-bar-circle here so the overlay is always in circle shape

Before
Screenshot 2023-04-13 at 12 24 05 p m

After
Screenshot 2023-04-13 at 12 24 27 p m

recipes/list_items/contact_info/contact_info.vue
```diff
@@ -23,6 +23,7 @@
               :size="avatarSize"
               :overlay-icon="avatar.icon"
               :overlay-text="avatar.text"
+              overlay-class="d-mn4 d-ba d-baw4 d-bc-white d-bar-pill d-box-unset"
```
juliodialpad:
```suggestion
              overlay-class="d-mn4 d-ba d-baw4 d-bc-white d-box-unset"
```

recipes/list_items/contact_info/contact_info.vue
```diff
@@ -36,6 +37,7 @@
...
```
juliodialpad:
```suggestion
              overlay-class="d-mn4 d-ba d-baw4 d-bc-white d-box-unset"
```

juliodialpad (changes requested):
Looks better now, just a couple issues I found.

Please replace this lines: 
- https://github.com/dialpad/dialtone-vue/blob/DP-66880-fix-box-sizing-issue/recipes/list_items/contact_info/contact_info_variants.story.vue#L263-L264
- https://github.com/dialpad/dialtone-vue/blob/DP-66880-fix-box-sizing-issue/recipes/list_items/contact_info/contact_info_variants.story.vue#L269-L270

e.g
```
{ src: 'static/recipes/list_items/contact_info/avatar.png' },
```

with a require statement, and provide the `initials` prop too, so we can get rid of all the errors

```
{ src: require('./avatar.png'), initials: 'initials' }
```

Those routes work well on local but once deployed it throws errors.

Screenshot 2023-04-13 at 12 40 20 p m

components/avatar/avatar.vue
SheldonDialpad: > The z-index needs to be `d-zi-base1` in order to be able to override the avatarImage in certain cases.

Yeah, I thought as you before. However, we may face to multiple admins listen-in case later. If we use `d-zi-base1`, the overlay icon will cover next avatar.
Screen Shot 2023-04-14 at 8 50 18 AM

components/avatar/avatar.vue
```diff
@@ -220,6 +227,14 @@ export default {
       ];
     },
 
+    overlayClasses () {
+      return [
+        'd-bgc-black-900 d-o70 d-ps-absolute d-w100p d-h100p d-d-flex d-ai-center d-bar-circle',
+        this.showInitials ? 'd-zi-base1' : 'd-zi-base',
+        this.overlayClass,
```
juliodialpad:
```suggestion
        'd-bgc-black-900 d-o70 d-ps-absolute d-w100p d-h100p d-d-flex d-ai-center d-bar-circle d-zi-base',
        this.overlayClass,
```

Update this after I merge #908 into staging

components/avatar/avatar.vue
```diff
@@ -301,12 +316,7 @@ export default {
     },
 
     formatInitials (initials) {
-      if (!initials) {
-        this.formattedInitials = '';
-        return;
-      }
-
-      if (this.validatedSize === 'xs') {
+      if (!initials || this.validatedSize === 'xs') {
```
juliodialpad, line 319: Nice clean up, thanks

components/avatar/avatar.vue
```diff
@@ -350,9 +360,7 @@ export default {
         count++;
       }
       const rng = seedrandom(this.seed);
-      const shuffledColors = Array.from(colors).sort(() => 0.5 - rng());
-
-      return shuffledColors;
+      return Array.from(colors).sort(() => 0.5 - rng()); // shuffle colors
```
juliodialpad, line 363: Nice clean up, thanks!

components/avatar/avatar.vue
```diff
@@ -361,7 +369,7 @@ export default {
       // If alt set to empty string consider it valid, as this is a valid case if the
       // image is already described by something else (ex: visible description)
       // eslint-disable-next-line no-unneeded-ternary
-      const isAltMissing = this.$refs.canvas.firstChild.getAttribute('alt') === null ? true : false;
+      const isAltMissing = this.$refs.canvas.firstChild.getAttribute('alt') === null;
```
juliodialpad, line 372: Lol, wonder why this was like this and no one noticed it.

juliodialpad (approved):
Looking good! just wait for my PR to be merged into staging and update before merging this.

PR #908, opened 2023-04-14
fix(avatar): image size and z-index issues

juliodialpad:
# Fix avatar - image size and z-index issues

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

- Fixed z-index issues caused by displaying both `initials` and `image`, changed back to `d-zi-base`.
- Removed logic that was preventing `d-avatar__image` class being added to avatar.
- Added a decorator to `avatar.stories.js` to contain the avatar.

## :bulb: Context

- Logic preventing `d-avatar__image` class being added #901
- `d-avatar` class recently changed `display: inline-flex` to `display: flex` and it was causing issues on avatar's presence positioning. https://github.com/dialpad/dialtone/pull/852

## :pencil: Checklist

- [ ] I have added tests

## :crystal_ball: Next Steps

Update #906 with this changes.

### DIFF
--- a/components/avatar/avatar.stories.js
+++ b/components/avatar/avatar.stories.js
@@ -96,6 +96,9 @@ const defaultImage = require('./person.png');
 
 // Stories
 export const Default = DefaultTemplate.bind({});
+Default.decorators = [() => ({
+  template: `<div class="d-d-flex"><story /></div>`,
+})];
 Default.args = {
   default: `<img data-qa="dt-avatar-image" src="${defaultImage}" alt="Person Avatar">`,
   initials: 'PS',
--- a/components/avatar/avatar.vue
+++ b/components/avatar/avatar.vue
@@ -38,13 +38,13 @@
     </div>
     <span
       v-if="showGroup"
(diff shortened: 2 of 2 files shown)


components/avatar/avatar.stories.js
```diff
@@ -96,6 +96,9 @@ const defaultImage = require('./person.png');
 
 // Stories
 export const Default = DefaultTemplate.bind({});
+Default.decorators = [() => ({
+  template: `<div class="d-d-flex"><story /></div>`,
```
braddialpad, line 100: Kinda confused why we need this when the wrapper on the component itself already has d-d-flex hmmm....

juliodialpad: It's not actually a flex parent, we just need to contain the avatar or else it will take all the available width, which on storybook breaks the avatar

braddialpad: alright guess we'll see how it goes

braddialpad (approved):

PR #914, opened 2023-04-17
fix: remove dialtone-icons from externals

braddialpad:
remove dialtone-icons from externals so it does not have to be installed independently by the consumer


juliodialpad:
If you remove it from externals it'll increase the bundle size too much, that's why added it as externals.

braddialpad:
Bundle size increase expected. Will separate out into a different build target in the future.

PR #924, opened 2023-04-19
test(rich-text-editor): add tests for rich text editor

MikaDialpad:
# test: add tests for rich text editor

- [ ] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation
- [x] Test

## :book: Description

Tests for [RichTextEditor component](https://github.com/dialpad/dialtone-vue/pull/905)

### DIFF
--- a/components/rich_text_editor/rich_text_editor.test.js
+++ b/components/rich_text_editor/rich_text_editor.test.js
@@ -38,26 +38,17 @@ const _mountWrapper = () => {
   });
 };
 
-async function mountWithProps (newProps) {
-  // Can't use setProps since the Editor instance is created in mounted().
-  // There are helper methods we could use to update the editor options
-  // (editor.setOptions), but we want to test that the passed props are
-  // properly getting through to the editor, so rather re-mount.
-  propsData = { ...propsData, ...newProps };
-  _mountWrapper();
-  await wrapper.vm.$nextTick();
-  _setChildWrappers();
-}
-
 describe('DtRichTextEditor tests', function () {
   // Test Setup
-  beforeEach(function () {
+  beforeEach(async function () {
     propsData = baseProps;
     inputStub = sinon.stub();
     listeners = {
       input: inputStub,
     };
-    mountWithProps();
+    _mountWrapper();
+    await wrapper.vm.$nextTick();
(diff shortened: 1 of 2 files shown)


components/rich_text_editor/rich_text_editor.test.js
```diff
@@ -0,0 +1,197 @@
...
+      it('should pass through the prop to the editor', function () {
+        assert.equal(editor.attributes('contenteditable'), 'false');
+      });
+    });
+  });
+
+  describe('Build Tests', function () {
+    it('should not be included in the core build', async function () {
+      // Ok this one is a bit goofy, but it's a naive attempt to make sure no
+      // one exports this component at the root level.
```
braddialpad, line 183: haha I like this actually

components/rich_text_editor/rich_text_editor.test.js
```diff
@@ -0,0 +1,197 @@
...
+    slots,
+    localVue,
+  });
+};
+
+async function mountWithProps (newProps) {
+  // Can't use setProps since the Editor instance is created in mounted().
+  // There are helper methods we could use to update the editor options
+  // (editor.setOptions), but we want to test that the passed props are
+  // properly getting through to the editor, so rather re-mount.
```
braddialpad: My only issue with this is that it will not capture any reactivity problems. It's a pretty common problem (especially when using 3rd party libraries) that a component will mount successfully with the passed in props, but then doesn't react when props are changed after it's rendered. 

It's fine to do it this way for the initial mount but we should maybe also test props changes with setProps after rendering the component because otherwise all your "watch" code is currently not covered by tests.

braddialpad (commented):
Very nice tests! Just an issue with the watch functions not being covered.

components/rich_text_editor/rich_text_editor.test.js
MikaDialpad: Yeah good point! I think I wrote these tests before I had that watcher set up haha. And to prove your point, the tests did find a bug right away! The aria-readonly was being set backwards, so I jut slipped that fix in this PR if that's ok.

braddialpad (approved):
Looks great thanks!

PR #928, opened 2023-04-20
feat(rich-text-editor): add rich text editor component - vue3

MikaDialpad:
# Vue 3 version of #905 and #924


MikaDialpad:
The only differences are using `modelValue` instead of `value` and some small syntax changes in the test and stories.js. And well using `@tiptap/vue-3` instead of vue-2 version of course

braddialpad (approved):
Thanks!

PR #929, opened 2023-04-20
fix: channels icons size 200 instead of 300

braddialpad:
# fix: channels icons size 200 instead of 300

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Make channels icons size 200 instead of 300, requested by design


recipes/leftbar/general_row/general_row.vue
```diff
@@ -25,7 +25,6 @@
           name="left"
         >
           <dt-recipe-leftbar-general-row-icon
-            :class="{ 'd-o50': isTyping }"
```
juliodialpad, line 28: Removing this class was part of design request too? 🤔

juliodialpad (approved):
If removing the class commented was part of design this LGTM

recipes/leftbar/general_row/general_row.vue
braddialpad: Nah we just made a change a while back to not show the avatar at all rather than show it at 50% opacity. So this class would never be triggered.

francisrupert (commented):
Looks good, though it occurs to me...

Will the "All channels" also be updated since it uses the same one? That one should remain at 18, while only those within the Channels _group_ should bump down to 14.

image

braddialpad:
> Will the "All channels" also be updated since it uses the same one? That one should remain at 18, while only those within the Channels _group_ should bump down to 14.

hmmm did not know that.. Yeah I will have to make an additional change for this to work.

braddialpad:
In fact I think I will just allow setting the icon size independently.

braddialpad:
Added iconSize prop, valid values 200 and 300.

francisrupert (approved):
Looks good and makes sense to control the possible icon sizes this way. I'm guessing you'll need to make the change directly in product codebase specifically for the Channel ones to be `200`.

braddialpad:
> I'm guessing you'll need to make the change directly in product codebase specifically for the Channel ones to be `200`

Yep that's right

PR #933, opened 2023-04-20
fix: visual tests not triggering on approval

braddialpad:
Visual tests not triggering on approval, only triggering on PUSHES after approval... attempting fix here.


juliodialpad (approved):
Approving for testing.

PR #934, opened 2023-04-20
fix: visual tests not triggering on approval

braddialpad:
vue 3


juliodialpad (approved):
Approving for testing

PR #944, opened 2023-04-26
chore: enable only inline component tests

braddialpad:
Lots of visual test flakiness. Going to disable all tests except inline components for right now since its impossible to get a consistent base branch due to flakiness and it keeps triggering false negatives. Will re-enable one by one as we figure out the issues.


braddialpad:
Also added some code to seed some of the avatars that were not yet seeded within the visual tests. This should fix their flakiness as you can see in the current run of the tests.

iropolo (approved):
Seems like you know what you are doing, this PR will not block you anymore 🧙🏻‍♂️

PR #945, opened 2023-04-26
chore: enable only inline component tests

braddialpad:
Vue 3


juliodialpad (commented):
Why are initials displaying instead of images on avatars? I can't find any change related to removing those.

braddialpad:
> Why are initials displaying instead of images on avatars? I can't find any change related to removing those.

Yeah.. this is one of the things that keeps flaking out.. sometimes it shows like that sometimes it doesn't.

braddialpad:
Gonna try removing this 250 wait...

braddialpad:
Seems okay now??

PR #950, opened 2023-05-01
fix(header): fix exports

laurafirespotter:
# Fix exports in header settings menu

- [X] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

I noticed these two places but it still doesn't seem to fix the issue that ubervoice can't find this component.

My IDE doesn't highlight this correctly but does for other components so maybe that's a clue.

Screen Shot 2023-05-01 at 1 50 15 PM
Screen Shot 2023-05-01 at 1 50 29 PM

### DIFF
--- a/index.js
+++ b/index.js
@@ -49,7 +49,7 @@ export * from './recipes/list_items/contact_info';
 export * from './recipes/notices/top_banner_info';
 export * from './recipes/cards/ivr_node';
 export * from './recipes/chips/grouped_chip';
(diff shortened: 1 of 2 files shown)


braddialpad (approved):
Ahhh good find, that was referencing the vue file directly rather than the "index.js" file that includes all the exports.

braddialpad:
> I noticed these two places but it still doesn't seem to fix the issue that ubervoice can't find this component.

How did you test it? npm link?

PR #953, opened 2023-05-01
docs(rich-text-editor): link extension tech spec

MikaDialpad:
# docs(rich text editor): link extension tech spec

- [ ] Fix
- [ ] Feature
- [ ] Refactoring
- [x] Documentation

## :book: Description

Tech spec for the Link extension used by DtRichTextEditor. I didn't touch testing yet since I'm not sure how easy it is to test something like this as the bulk of the functionality will be in the ProseMirror plugin. I'll have to try it out and add a section here later on.

For easier reading: https://github.com/dialpad/dialtone-vue/blob/3b0afb5d4e48c20c6621555cc8ea8f9fd4833cc3/components/rich_text_editor/rich_text_editor.tech_spec.md

## :crystal_ball: Next Steps

Implement the extension and see how we can write tests for it.


MikaDialpad:
FYI @dbecherdialpad

components/rich_text_editor/rich_text_editor.tech_spec.md
```diff
@@ -292,27 +332,198 @@ feed items also. The mentions/channels list dropdowns can probably be purpose
...
+
+### Overview
+
+After a brief [investigation](https://dialpad.atlassian.net/browse/DT-1153) and
+some POC work it became apparent that we can't unfortunately rely on the
+`@tiptap/extension-link` package that provides similar functionality. The
+package wouldn't allow us to customize the behavior of the extension to meet our
+needs, so we'll have to build a custom extension instead. It would for example
+only validate the link after a following whitespace character, which would leave
+all the links at the end of a message not linkified. With a custom extension we
```
braddialpad, line 367: Seems weird it would work like this.. Why would anyone want to leave a link at the end of a message unlinked??

components/rich_text_editor/rich_text_editor.tech_spec.md
```diff
@@ -292,27 +332,198 @@ feed items also. The mentions/channels list dropdowns can probably be purpose
...
+can create rules to linkify text as it is being typed.
+
+Also worth noting that with my current understanding we should construct this as
+a Mark instead of a Node. For example backspacing at the beginning of a Mark and
+merging its contents with the preceeding content seems easier than with a Node,
+where the default behavior is to exit and remove the Node and convert its
+contents back to plain text. The main consequence of this decision is that Marks
+can render only native HTML tags, so in this case an `<a>` tag with `d-link`
+class instead of `DtLink`. To the user they would look indistinguishable though
```
braddialpad, line 376: Should be fine, the only other functionality `DtLink` does is allow you to set kind, which in this situation I'm guessing would always be default.

components/rich_text_editor/rich_text_editor.tech_spec.md
```diff
@@ -292,27 +332,198 @@ feed items also. The mentions/channels list dropdowns can probably be purpose
...
+- Phone Numbers
+  - 7-15 characters long
+  - Without separators, e.g. `7144107035`
+  - With dashes `-`, e.g. `714-410-7035`
+  - With whitespace `\s`, e.g. `714 410 7035`
+  - With parentheses `()`, e.g. `(714) 410-7035`
+  
+  Dialpad backend uses these rules to do an initial validation first and when
+  matched, further uses a more rigorous validation using `phone_number_util`,
+  but for our purposes we'll just replicate the initial validation.
```
braddialpad, line 404: We should include these "phone number checking utils" in dialtone and export them separately so they could be used for anything. One of our near future goals is to start providing utils for things like dates / times / phone numbers

components/rich_text_editor/rich_text_editor.tech_spec.md
```diff
@@ -292,27 +332,198 @@ feed items also. The mentions/channels list dropdowns can probably be purpose
...
+```
+1. “Check out dialpad.c"
+2. “Check out [dialpad.co]”   // link
+3. “Check out [dialpad.com]”  // still a link, just a different match
+```
+
+#### 2. Remove link when a previously valid match doesn’t match anymore
+```
+1. “Check out [dialpad.com]” // link
+2. “Check out dialpad.comm”  // not a valid link anymore
```
braddialpad, line 443: How do you know its valid? Do you actually send a request to the link? or do you have a list of valid TLDs?

components/rich_text_editor/rich_text_editor.tech_spec.md
```diff
@@ -292,27 +332,198 @@ feed items also. The mentions/channels list dropdowns can probably be purpose
...
+backticks are reserved for codeblocks
+  - Except if ONLY preceding or following
+
+Otherwise all link types should go by the following behavior:
+
+#### 1. Add link when a match is found and continue matching while still valid
+```
+1. “Check out dialpad.c"
+2. “Check out [dialpad.co]”   // link
+3. “Check out [dialpad.com]”  // still a link, just a different match
```
braddialpad, line 437: a bit confused by the square brackets here, that means it's rendered as a link? we should imagine they aren't there?

components/rich_text_editor/rich_text_editor.tech_spec.md
```diff
@@ -292,27 +332,198 @@ feed items also. The mentions/channels list dropdowns can probably be purpose
...
+
+  - These use a more loose (and often an incorrect) validator and go by the
+  following rules:
+    - *Preceded* by any character except a letter (`\p{L}`) or a number
+    (`\p{N}`)
+    - *Followed* by `<`, `>`, any whitespace character (`\s`) or one or more
+    repeating `?` or `#`
+      - This rule is way too loose and allows total nonsense URLs to be
+      linkified, but for the best UX we should linkify the same strings as
+      Dialpad backend does.
```
braddialpad, line 485: Fair enough but can we create a task or something to change the backend? this seems kinda ridiculous to still have in the new and improved DP 2.0. I know it's probably not that easy of a fix...

components/rich_text_editor/rich_text_editor.tech_spec.md
```diff
@@ -292,27 +332,198 @@ feed items also. The mentions/channels list dropdowns can probably be purpose
...
+and forwards on a string from a given position and returns the entire word (as
+bound by whitespace), so something like
+[this](https://stackoverflow.com/a/5174867). We can then use this both at the
+beginning of the range and at the end to get the entire range with complete
+words in case the change was something more than just one character change.
+
+- It *would* be way simpler to just look at the entire content and linkify it
+instead of the substring, but that could manifest in performance issues in
+extreme cases (super long messages and/or tons of links in the text) and we want
+this to be very quick since it is run after every keystroke.
```
braddialpad, line 545: This does seem way simpler, perhaps we should do a performance test and see if it makes any significant difference? no need to micro-optimize unless necessary.

components/rich_text_editor/rich_text_editor.tech_spec.md
dbecherdialpad: I don't think this is totally unheard of, I am pretty used to typing a space after pasting a link to have it show up (the mail app does this too). I suspect this behavior might have to do with how the text is parsed and tokenized for perf reasons, which I am a little concerned about with our implementation.

components/rich_text_editor/rich_text_editor.tech_spec.md
```diff
@@ -292,27 +332,198 @@ feed items also. The mentions/channels list dropdowns can probably be purpose
...
+  - With a protocol, e.g. `https://dialpad.com`
+- Email Addresses
+  - Without `mailto:` prefix, e.g. `mika@dialpad.com`
+  - With `mailto:` prefix, e.g. `mailto:mika@dialpad.com`
+  - With param(s), e.g. `mika@dialpad.com?subject=hey`
+- Phone Numbers
+  - 7-15 characters long
+  - Without separators, e.g. `7144107035`
+  - With dashes `-`, e.g. `714-410-7035`
+  - With whitespace `\s`, e.g. `714 410 7035`
```
dbecherdialpad, line 397: I feel like "with whitespace" might end up causing us grief...just a note that if it ends up being a hassle or performance issue, I would be ok NOT highlighting this format in the editor

components/rich_text_editor/rich_text_editor.tech_spec.md
```diff
@@ -292,27 +332,198 @@ feed items also. The mentions/channels list dropdowns can probably be purpose
...
+  - With parentheses `()`, e.g. `(714) 410-7035`
+  
+  Dialpad backend uses these rules to do an initial validation first and when
+  matched, further uses a more rigorous validation using `phone_number_util`,
+  but for our purposes we'll just replicate the initial validation.
+
+### Common Editor Behavior
+
+- All links are invalid (not linkified) if preceded by `#` or `@` characters,
+since they're reserved for user mentions and channel hashtags.
```
dbecherdialpad, line 426: Or: literal numbers (eg #12345)

components/rich_text_editor/rich_text_editor.tech_spec.md
```diff
@@ -292,27 +332,198 @@ feed items also. The mentions/channels list dropdowns can probably be purpose
...
+- Similarly links wrapped with ` ``` ` will not be linkified as the triple
```
dbecherdialpad: Does this maybe more generally mean "we won't linkify things contained in another node" or are there cases where we might want to?

components/rich_text_editor/rich_text_editor.tech_spec.md
dbecherdialpad: List of valid TLDs could get really complicated, not to mention hard to maintain

components/rich_text_editor/rich_text_editor.tech_spec.md
dbecherdialpad: I think just calling out "this is the part that's linkified"

components/rich_text_editor/rich_text_editor.tech_spec.md
```diff
@@ -292,27 +332,198 @@ feed items also. The mentions/channels list dropdowns can probably be purpose
...
+
+#### 3. Add New Links
+
+- As mentioned above, this step acts as a catch-all for new links and "old"
+links that are still valid but were removed in the previous step. However at
+this point we just have the range of the transaction and using that to get the
+substring of the changed content is not enough. For example, if we had content
+`Hey dialpadcom` that was changed to `Hey dialpad.com`, the transaction would
+give us the range for just the changed content, which is the added `.`. What we
+actually need is the entire word, or words, that were affected, so `dialpad.com`
```
dbecherdialpad, line 534: Boy this sure seems really complicated. What's the other approach that just uses a regex?

components/rich_text_editor/rich_text_editor.tech_spec.md
dbecherdialpad: Nah I definitely think we should do it something like this (not parsing all the text every single change), BUT I am curious about what I alluded to above with marks that span whitespace boundaries. How do we know how far back/forward to look? It can't just be the current word since that wouldn't catch phone numbers with spaces.

components/rich_text_editor/rich_text_editor.tech_spec.md
MikaDialpad: Yeah I've definitely seen it being implemented like that in a bunch editors, but I really didn't like the UX of the [built in extension](https://tiptap.dev/api/marks/link) at least. It doesn't for example validate the word if you go from `dialpadcom` -> `dialpad.com` or if you remove space between a valid link and preceding/following word `hey [dialpad.com]` -> `hey[dialpad.com]`. I used Slack as a benchmark here as theirs works really well and it _seems_ to do something similar to what I'm proposing here.

components/rich_text_editor/rich_text_editor.tech_spec.md
MikaDialpad: Ugh I saw your other comment about the word separating characters and I think you're spot on there. I haven't actually tried to implement phone numbers yet, so I don't know how easy it would be to overcome this issue, but I'll definitely try before abandoning any formats.

components/rich_text_editor/rich_text_editor.tech_spec.md
MikaDialpad: Oh that's good to know! I should stress though that the regexes and rules I use come from the current backend implementation, so they're far from perfect, but if we're aiming for parity with DP then they should be fine.

components/rich_text_editor/rich_text_editor.tech_spec.md
MikaDialpad: Great question! I don't think we need to worry about that too much yet, but I think it'll depend on the node. For codeblocks we most likely don't want any linking or any other formatting like bolding, but if one day we wanted to build a vcard extension for example then we probably want the contents to be linkified.

And now testing this again I see that codeblocks don't just work in my beta at all??? So my initial tests were giving wrong results as I saw something like ` ```dialpad.com``` ` shown as plain text and thought the missing whitespace invalidated the codeblock. Anyways, I will not do any special rules for backticks now but eventually when we have node extensions we'll have to probably exclude or include them here.

components/rich_text_editor/rich_text_editor.tech_spec.md
MikaDialpad: Yeah that was my intention. We wont actually render any square brackets :)

components/rich_text_editor/rich_text_editor.tech_spec.md
MikaDialpad: [Tadaa](https://github.com/dialpad/firespotter/blob/c6a7922ac3a5368cb8b96b53c4fb268b5af4f64b/ubervoice/utils/rich_media.py#L108). Spoiler alert, we keep a list of valid TLDs and it is a 6 years old list, so marios.pizza wont be linkified. I do agree with the original author's sentiment though that the TLD regex will capture all but very marginal special domains. Even more importantly, most urls in messages are probably(??) copy-pasted anyways so they have the protocol prefix and hit a different regex that has the opposite problem of validating almost anything typed before and after `://`

components/rich_text_editor/rich_text_editor.tech_spec.md
MikaDialpad: Yeah I totally get your sentiment. I don't _think_ it'd be that difficult since it's all just regex, but also I don't know how high priority issue it really is to anyone lol. We can always try of course.

I would claim though without any actual data that 99% of urls sent as messages are either typed really simple domain names like `dialpad.com` or copy-pasted from the browser's search bar and as such they should mostly be valid unless they messed up some spacing or whatnot.

components/rich_text_editor/rich_text_editor.tech_spec.md
MikaDialpad: So this would still use regex, but the string to match with would be shorter than the alternative which is to just match the entire text content of the editor. I did some benchmarking and for short messages the difference is non-existent, but for reeeally long messages typing might start to lag on low spec machines.

I have a POC with this and it doesn't seem that bad except for the phone number edge case that you brought up. I'll try it out to see how much more complex it makes this.

components/rich_text_editor/rich_text_editor.tech_spec.md
MikaDialpad: Yeah good points. I commented above that I've seen some lagginess when testing the very extreme cases of super longs texts with tons of links while matching the entire text. For DP messaging that might not be that relevant since we have character limits, but if the agent experience team uses this to compose emails then it could become a real life issue.

For my proposed way the text to match isn't affected by how long the message is, so it doesn't really degrade at all. Also I'm thinking that this is just one extension we're implementing, but eventually we will have a bunch more that might do something similar and they can have a compounding effect. I have a POC for how this could work (barring the phone number edge case) which I can show if you're interested.

components/rich_text_editor/rich_text_editor.tech_spec.md
braddialpad: yeah I think we're going to have to do that for now.

components/rich_text_editor/rich_text_editor.tech_spec.md
braddialpad: maybe we should use this so it could be easily maintained just by updating the package? https://www.npmjs.com/package/tlds

components/rich_text_editor/rich_text_editor.tech_spec.md
braddialpad: Alright cool, just wanted to make sure

components/rich_text_editor/rich_text_editor.tech_spec.md
MikaDialpad: Yeah sounds good. Do you have any requirements for the utils for what they should take as inputs and what to output?

braddialpad: probably something like...

```
isPhoneNumber(phoneNumber [string]) {
  // checks whether phone number is valid against all the possible formats
  returns bool
}
```

components/rich_text_editor/rich_text_editor.tech_spec.md
MikaDialpad: I'm down for anything as long as we can match the behavior on DP backend side as well. For now I would rather go with the existing regex so we don't get blocked and just keep it easy enough to swap out with a better regex later if we get the backend updated.

components/rich_text_editor/rich_text_editor.tech_spec.md
MikaDialpad: I added a section to the tech spec mentioning this. I proposed also creating utils for URLs, emails and IP addresses, but let me know if we don't need some of them.

components/rich_text_editor/rich_text_editor.tech_spec.md
braddialpad: fair enough

components/rich_text_editor/rich_text_editor.tech_spec.md
braddialpad: that's great thanks!

components/rich_text_editor/rich_text_editor.tech_spec.md
```diff
@@ -292,27 +332,198 @@ feed items also. The mentions/channels list dropdowns can probably be purpose
...
+- Email Addresses
+  - Without `mailto:` prefix, e.g. `mika@dialpad.com`
+  - With `mailto:` prefix, e.g. `mailto:mika@dialpad.com`
+  - With param(s), e.g. `mika@dialpad.com?subject=hey`
+- Phone Numbers
+  - 7-15 characters long
+  - Without separators, e.g. `7144107035`
+  - With dashes `-`, e.g. `714-410-7035`
+  - With whitespace `\s`, e.g. `714 410 7035`
+  - With parentheses `()`, e.g. `(714) 410-7035`
```
braddialpad, line 398: Would this also include numbers with country code? `1 (714) 410-7035` and also what about the plus sign? (international prefix) `+1 (714) 410-7035`

MikaDialpad: Good callout. Yeah it should but it's not perfect. The regex will basically validate most combinations of the allowed characters, so start with an optional `+` and then a mix of numbers, whitespace, parentheses and dashes and length being between 7 and 15. So for your examples the first `1` and `1+` go over the 15 character limit and only the latter part of the number gets linkified. Same applies for DP backend. Numbers like `+17144107035` and `+1 714 410 7035` fit within 15 chars and would be fully linkified

braddialpad (approved):
All my issues have been resolved, I think the remaining questions are around performance but I think the approach taken here is reasonable and perhaps could be improved through testing if needed.

MikaDialpad:
> All my issues have been resolved, I think the remaining questions are around performance but I think the approach taken here is reasonable and perhaps could be improved through testing if needed.

Thanks for the review! Is there some label to bypass the visual_tests task that's blocking the merge? `no-visual-test` doesn't seem to cut it :)

braddialpad:
Ugh sorry I have to fix that.. I'll just force merge for you for now.

PR #958, opened 2023-05-03
fix(scroller): name refactor, events behaviour and hardcoded attribute

iropolo:
# scroller improvements

- [x] Fix
- [x] Refactoring

## :book: Description

- Name refactoring to achieve our naming conventions.
- Events (scroll-start, scroll-end) now are triggered in a different way but same behaviour
- Test update
- Fixed the key-field hardcoded attribute

This improvements comes from [DT-1141](https://dialpad.atlassian.net/browse/DT-1141)

Last ticket update:

When the user add information to the top of the list we just scroll the user to the previous position. For example, the user is in the position 0 (there is where we update the list) after we update the list with 10 items. we scroll 0 + 10 elements and it will be on the same place. It is imperceptible for the user the scroll.

I also created 3 functions to improve the scroller functionality.

- `scrollToBottom()` (only accessible by the dynamic component) This function will scroll to the bottom of the scroller, this is necessary because the dynamic component has variable size.
 
- `updateItems()` This function is necessary to call when the user do some update to the list (add elements from the top with unshift or update some item).

- `updateItemsFromBottom()` This function is necessary to call when the user push some element to the array list.

[DT-1141]: https://dialpad.atlassian.net/browse/DT-1141?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


iropolo:
Before 
Merge branch 'staging-vue3' into dt-1141-scroller-improvements 3a16ff8
the a11y_tests works, but after this it fails

braddialpad:
> Before Merge branch 'staging-vue3' into dt-1141-scroller-improvements [3a16ff8](https://github.com/dialpad/dialtone-vue/commit/3a16ff8cd6bdb14f2c45c98b81687054ecc799bb) the a11y_tests works, but after this it fails

hmm yes.. I think I have a fix for this 1 sec

braddialpad:
> hmm yes.. I think I have a fix for this 1 sec

Actually.. it ran fine on staging yesterday. I think it's due to this outage.. https://www.githubstatus.com/

braddialpad:
Yep completed successfully on re-run.

components/scroller/scroller.stories.js
```diff
@@ -91,15 +92,8 @@ export const argTypesData = {
     },
   },
```
braddialpad, line 93: add

```
onUserPosition: {
  table: {
    disable: true,
  },
},
```

components/scroller/scroller_default.story.vue
```diff
@@ -12,6 +12,19 @@
       Go to end
     </button>
 
+    <button @click="addItemTop">
+      Add 10 items top
+    </button>
```
braddialpad, line 17: When I add items to top the scroller jumps somewhere near the top. Is that expected to happen? or is it supposed to stay in place?

components/scroller/scroller_dynamic.story.vue
```diff
@@ -5,14 +5,24 @@
...
       Update Element Message
     </button>
+    <br>
+    <br>
+    <button
+      class="autoscrolling"
+      @click="switchAutoScrolling"
+    >
+      Auto scrolling <div :class="{ 'enabled': autoScrolling }" />
+    </button>
```
braddialpad, line 15: Initially starting this it jumps to the bottom but then remains in position as new entries are added.

components/scroller/modules/core_scroller.vue
```diff
@@ -557,6 +561,20 @@ const scrollToItem = (index) => {
 };
 
 const handleScroll = () => {
+  const container = scroller.value;
+
+  emit('user-position', 'middle');
```
braddialpad: Do we need to trigger this on every single scroll movement or only when it changes from "top" to "middle" or "middle" to "bottom" etc. We're outputting the exact same event hundreds of times which might not be desired if they are handling the middle event on the consumer side.

components/scroller/scroller.vue
```diff
@@ -0,0 +1,192 @@
...
+    */
+  scrollerWidth: {
+    type: [String, Number],
+    default: '100%',
+  },
+});
+
+const emits = defineEmits([
+  /**
+   * Emitted when the user scrolls the scroller.
```
braddialpad: Describe when the scroller changes from start/middle/end

components/scroller/scroller.vue
```diff
@@ -0,0 +1,192 @@
...
+  if (scroller.value) scroller.value.scrollToItem(index);
+}
+
+function updateItems () {
+  if (scroller.value) scroller.value._updateVisibleItems(true);
+}
+
+function updateItemsFromBottom () {
+  if (scroller.value) scroller.value._updateVisibleItems(false, true);
+}
```
braddialpad, line 174: It seems kind of "anti-reactive" to have to manually call this. This is due to limitations of the scroller library?

braddialpad: We should probably document these in the mdx once you create it.

braddialpad:
Adding @AlexLamDialpad, Can you check out the preview and see if it does what you need it to do now?

components/scroller/scroller_default.story.vue
iropolo: The expected behaviour is this:
You might load elements when you are in the position 0 (because you are on the top).
When you load elements you dont want to move the user or the elements (similar when you scroll to top in a chat).
It will load 10 elements and in the same time Im moving the user 10 elements below to the place where the event was triggered.

components/scroller/scroller_dynamic.story.vue
iropolo: Yes, thats the expected behaviour.
If you are on the bottom you should autoscroll when new elements are added.

components/scroller/modules/core_scroller.vue
iropolo: hmm I can update it to not emit the event always, just when it changes. let me update it

components/scroller/scroller.vue
iropolo: Before it was a watcher in the core of the library.
Now we need to trigger it manually, I would add this on the mdx.

braddialpad: k sounds good.

braddialpad (approved):
Thanks

PR #972, opened 2023-05-09
feat(feed-items): create feed item emoji row

AlexLamDialpad:
# PR Title

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :pencil: Checklist

- [x] I have added tests

image

## :link: Sources
https://dialpad.atlassian.net/browse/DIALTONE-574

### DIFF
--- a/recipes/conversation_view/emoji_row/emoji_row.stories.js
+++ b/recipes/conversation_view/emoji_row/emoji_row.stories.js
@@ -12,14 +12,16 @@ export const argsData = {
 export const argTypesData = {
   // Props
   reactions: {
-    description: 'Array of reactions to be displayed. Expects an object with these attributes: ' +
-    ' emojiUnicodeOrShortname,' +
-    ' isSelected,' +
-    ' ariaLabel,' +
-    ' tooltip,' +
-    ' num',
-    control: {
-      type: 'object',
+    table: {
+      type: {
+        detail: `[{
+          "emojiUnicodeOrShortname": string,
+          "isSelected": boolean,
+          "ariaLabel": string,
+          "tooltip": string,
+          "num": number,
+        }]`,
+      },
     },
   },
 
--- a/recipes/conversation_view/emoji_row/emoji_row.stories.js
+++ b/recipes/conversation_view/emoji_row/emoji_row.stories.js
@@ -1,0 +1,93 @@
+import { action } from '@storybook/addon-actions';
+import { createTemplateFromVueFile } from '@/common/storybook_utils';
+import DtRecipeEmojiRow from './emoji_row.vue';
+import DtRecipeEmojiRowDefaultTemplate from './emoji_row_default.story.vue';
+
+// Default Prop Values
+export const argsData = {
+  onEmojiClicked: action('emojiClicked'),
+  onEmojiHovered: action('emojiHovered'),
+};
+
+export const argTypesData = {
+  // Props
+  reactions: {
+    description: 'Array of reactions to be displayed. Expects an object with these attributes: ' +
+    ' emojiUnicodeOrShortname,' +
+    ' isSelected,' +
+    ' ariaLabel,' +
+    ' tooltip,' +
+    ' num',
+    control: {
+      type: 'object',
+    },
+  },
+
+  // Action Event Handlers
+  onEmojiClicked: {
+    table: {
+      disable: true,
+    },
+  },
+  onEmojiHovered: {
+    table: {
+      disable: true,
+    },
+  },
+  emojiClicked: {
+    description: 'Emoji clicked event',
+    table: {
+      type: { summary: 'event' },
+    },
+  },
+  emojiHovered: {
+    description: 'Emoji hovered event',
+    table: {
+      type: { summary: 'event' },
+    },
+  },
+};
+
+// Story Collection
+export default {
+  title: 'Recipes/Conversation View/Emoji Row',
+  component: DtRecipeEmojiRow,
+  args: argsData,
+  argTypes: argTypesData,
+  excludeStories: /.*Data$/,
+};
+
+// Templates
+const DefaultTemplate = (args, { argTypes }) => createTemplateFromVueFile(
+  args,
+  argTypes,
+  DtRecipeEmojiRowDefaultTemplate,
+);
+
+// Stories
+export const Default = DefaultTemplate.bind({});
+Default.args = {
+  reactions: [
+    {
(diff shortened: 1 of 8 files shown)


AlexLamDialpad:
Open issues/questions:
1. When tabbing over with voice over, it reads the aria label and then the tooltip, should I hide the tooltip or leave it as is?
2. The hover test does not trigger the tooltip to appear, and therefore doesn't fire the emojiButton event

recipes/conversation_view/emoji_row/emoji_row.vue
```diff
@@ -0,0 +1,121 @@
...
+  ],
+
+  data () {
+    return {};
+  },
+
+  computed: {},
+
+  watch: {},
+
```
braddialpad: Can remove the empty stuff.

recipes/conversation_view/emoji_row/emoji_row.vue
```diff
@@ -0,0 +1,121 @@
...
+              'd-h24 d-bar16 d-py0 d-fs-200 d-mb0 d-t',
+              !!reaction.isSelected
+                ? `d-fc-purple-400 d-bgc-purple-400 d-bgo10 d-bc-purple-400 d-bco50 ' +
+                  'h:d-bgc-purple-400 h:d-bco25 a:d-bgc-purple-400 a:d-bco25`
+                : 'd-fc-black-600 d-bgc-black-100 h:d-bgc-white h:d-bc-black-600 a:d-bgc-white a:d-bc-black-600',
+            ]"
+            :aria-label="reaction.ariaLabel"
+            :attrs="attrs"
+            @click="emojiClicked(reaction)"
+          >
```
braddialpad: hmmm color contrast problem when it's selected.. might want to bring it up with design.

Screenshot 2023-05-09 at 4 05 00 PM

recipes/conversation_view/emoji_row/emoji_row.stories.js
```diff
@@ -0,0 +1,98 @@
+import { action } from '@storybook/addon-actions';
+import { createTemplateFromVueFile } from '@/common/storybook_utils';
+import DtRecipeEmojiRow from './emoji_row.vue';
+import DtRecipeEmojiRowDefaultTemplate from './emoji_row_default.story.vue';
+
+// Default Prop Values
+export const argsData = {
+  onEmojiClicked: action('emojiClicked'),
+  onEmojiHovered: action('emojiHovered'),
```
braddialpad: These actions aren't being output in the storybook.. I'm actually not sure why.. looks like you did it correctly. Are they maybe not actually emitting from the component?

braddialpad:
> Open issues/questions:
> 
> 1. When tabbing over with voice over, it reads the aria label and then the tooltip, should I hide the tooltip or leave it as is?

If the button is described by the tooltip you should probably use `aria-labelledby` instead of aria-label

> 2. The hover test does not trigger the tooltip to appear, and therefore doesn't fire the emojiButton event

Is it actually triggering? I can't get it to trigger in the storybook actions either.

recipes/conversation_view/emoji_row/emoji_row.vue
```diff
@@ -0,0 +1,121 @@
...
+          const reaction = reactions[i];
+          REACTIONS_ATTRIBUTES.forEach((attribute) => {
+            if (reaction[attribute] === undefined || reaction[attribute] === null) {
+              validInput = false;
+            }
+          });
+          if (!validInput) {
+            break;
+          }
+        }
```
braddialpad:
```js
for (reaction of reactions) {
  let validInput = REACTIONS_ATTRIBUTES.every((attribute) => reaction[attribute] ?? false);
  if (!validInput) return false;
}
return true;
```

recipes/conversation_view/emoji_row/emoji_row.vue
```diff
@@ -0,0 +1,121 @@
...
+     * Provide an array of reaction objects to be shown.
+     */
+    reactions: {
+      type: Array,
+      default: () => [],
+      validator: (reactions) => {
+        let validInput = true;
+        if (!Array.isArray(reactions)) {
+          return false;
+        }
```
braddialpad: Don't think you need this because it's already constrained by `type: Array`

braddialpad (changes requested):
Awesome! stoked for this stuff.

recipes/conversation_view/emoji_row/emoji_row.stories.js
iropolo: Its a naming issue

in the stories.js the events are registered like this:
onEmojiClicked: action('`emojiClicked`'),
onEmojiHovered: action('`emojiHovered`'),

But on emoji_row_default.story.vue are like this:
@`emoji-clicked`="onEmojiClicked"
@`emoji-hovered`="onEmojiHovered"

If you update it like
@`emojiClicked`="onEmojiClicked"
@`emojiHovered`="onEmojiHovered"

It will work but linter will piss off because events must be hyphenated and he is right.

My suggestions here is update the event names to be hyphernated.

recipes/conversation_view/emoji_row/emoji_row.vue
AlexLamDialpad: Talked to Bradley, will be bumping this to fc-purple-500

AlexLamDialpad:
> > Open issues/questions:
> > 
> > 1. When tabbing over with voice over, it reads the aria label and then the tooltip, should I hide the tooltip or leave it as is?
> 
> If the button is described by the tooltip you should probably use `aria-labelledby` instead of aria-label
> 
> > 2. The hover test does not trigger the tooltip to appear, and therefore doesn't fire the emojiButton event
> 
> Is it actually triggering? I can't get it to trigger in the storybook actions either.

In the product we want it to read the aria-label which has reduced text instead of the tooltip that may show up to 50 names. I went ahead and wrapped the tooltip content with span aria-hidden

braddialpad:
> In the product we want it to read the aria-label which has reduced text instead of the tooltip that may show up to 50 names. I went ahead and wrapped the tooltip content with span aria-hidden

k no prob

recipes/conversation_view/emoji_row/emoji_row.stories.js
braddialpad: looks like it's working now

braddialpad (approved):
All issues resolved. Looks good!

PR #981, opened 2023-05-15
feat(confact-info): add `showAvatar` and `avatarIcon` props

tonydialpad:

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

In recent contact update, if the contact has no `first name`, `last name` or `image_url`, the avatar will display nothing. To fix this, this pull request adds `avatarIcon` to support icon display in avatar when there is nothing to display in avatar. Moreover, we want to hide avatar in some use cases, so `showAvatar` property is added to control that.

## :pencil: Checklist

- [x] I have added tests

## :crystal_ball: Next Steps

Make a new release for `dialtone-vue`.

Screenshot 2023-05-15 at 2 45 48 PM

## :link: Sources

JIRA: https://dialpad.atlassian.net/browse/DP-70835
Vue3 PR: https://github.com/dialpad/dialtone-vue/pull/983


recipes/list_items/contact_info/contact_info.stories.js
```diff
@@ -108,9 +108,11 @@ export const Default = {
   render: DefaultTemplate,
 
   args: {
+    showAvatar: true,
```
juliodialpad: This is not needed as `showAvatar` prop is `true` by default

recipes/list_items/contact_info/contact_info.stories.js
```diff
@@ -108,9 +108,11 @@ export const Default = {
...
     avatarSrc: avatarImage,
     avatarSeed: 'seed',
     avatarInitials: 'JL',
+    avatarIcon: 'user',
```
juliodialpad: This is not needed as `avatarIcon` prop is `user` by default

recipes/list_items/contact_info/contact_info.vue
juliodialpad, line 1: This file is having a lot of `dt-avatar` components with almost the same properties over and over, do you think it could be simplified?

The only thing that's changing on some of them is just the slot content

juliodialpad (changes requested):

recipes/list_items/contact_info/contact_info.stories.js
tonydialpad: Hi @juliodialpad , after I remove this line, I found the `avatarIcon` field become empty string in the panel:
Screenshot 2023-05-16 at 11 50 42 AM

recipes/list_items/contact_info/contact_info.vue
tonydialpad: Okay, done, please review again. Thanks!

recipes/list_items/contact_info/contact_info.stories.js
juliodialpad: That's intended, that means that if no `avatarIcon` prop is set 'user' will be used as default.

Actually, please update that field to use the list of icons. You can see an example of that over [icon.stories.js](https://github.com/dialpad/dialtone-vue/blob/83c3c3ad37968ee5beaf953c70b986529b626199/components/icon/icon.stories.js#L6)

recipes/list_items/contact_info/contact_info.stories.js
```diff
@@ -182,6 +183,7 @@ export const Variants = {
     avatarInitials: 'NW',
     avatarColor: 'orange-500',
     avatarSize: 'lg',
+    avatarIcon: 'user',
```
juliodialpad: This isn't necessary too

recipes/list_items/contact_info/contact_info.vue
```diff
@@ -29,48 +29,41 @@
...
-              overlay-class="d-mn4 d-ba d-baw4 d-bc-white d-box-unset"
-              :avatar-class="['d-ba d-baw4 d-bc-white d-bar-pill', { 'd-mln24': index > 0 }]"
-            >
-              {{ avatar.initials }}
+              <div v-else-if="avatar.initials">
+                {{ avatar.initials }}
+              </div>
+              <div v-else>
+                <dt-icon :name="avatarIcon" />
+              </div>
```
juliodialpad: Don't need to wrap the icon on a `div` that'd prevent the avatar to correctly define the slot content type

recipes/list_items/contact_info/contact_info.vue
```diff
@@ -29,48 +29,41 @@
...
-          :seed="avatarSeed"
-          :presence="presence"
-        >
-          {{ avatarInitials }}
+          <div v-else-if="avatarInitials">
+            {{ avatarInitials }}
+          </div>
+          <div v-else>
+            <dt-icon :name="avatarIcon" />
+          </div>
```
juliodialpad: Same here, remove the wrapper `div`

juliodialpad (changes requested):
Nice! It looks a lot cleaner now, thanks!

Just a couple minor changes

recipes/list_items/contact_info/contact_info.stories.js
tonydialpad: Done, please review:
Screenshot 2023-05-18 at 12 16 19 PM

juliodialpad (approved):
Nice work, thanks!

PR #987, opened 2023-05-17
feat(emoji-picker):  keyboard navigation

iropolo:
# feat(emoji-picker):  keyboard navigation

- [x] Feature
- [x] Refactoring

## :book: Description

This is the addition of keyboard navigation tot he emoji picker component

https://dialpad.atlassian.net/browse/DIALTONE-578


iropolo:
Since we have [DIALTONE-269](https://dialpad.atlassian.net/browse/DIALTONE-269) to tackle improvements, I would implement those in that ticket, here I achieve keyboard navigation goals.

We also have [DIALTONE-268](https://dialpad.atlassian.net/browse/DIALTONE-268) for documentation, so I would explain this keyboard navigation there.

I tryed to migrate accesibility to composables but there is `scrollToTab` which I need also migrate to a composable, and inside it I have computed `tabLabels` so... It took me some time but I would like to move on with this and then continue it later.




[DIALTONE-269]: https://dialpad.atlassian.net/browse/DIALTONE-269?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[DIALTONE-268]: https://dialpad.atlassian.net/browse/DIALTONE-268?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

components/emoji_picker/modules/emoji_selector.vue
```diff
@@ -444,13 +478,202 @@ function setTabLabelObserver () {
...
+      // Jump to the next tab
+      // handle end of tab
+      if (!focusEmoji(indexTab + 1, 0)) {
+      // when reach the end of the list, jump to the first emoji of the first tabset
+        focusEmoji(0, 0);
+      }
+    }
+  }
+
+  if (event.key === 'Tab') {
```
braddialpad, line 615: Pressing shift + tab should move this in the other direction

components/emoji_picker/modules/emoji_selector.vue
```diff
@@ -65,15 +69,20 @@
...
+              'hover-emoji': (index === 0 && hoverFirstEmoji),
+            }"
             @click="$emit('selected-emoji', emoji)"
             @focusin="$emit('highlighted-emoji', emoji)"
             @focusout="$emit('highlighted-emoji', null)"
-            @mouseover="$emit('highlighted-emoji', emoji)"
-            @mouseleave="$emit('highlighted-emoji', null)"
+            @mouseover="hoverEmoji(emoji)"
+            @mouseleave="hoverEmoji(null)"
+            @keydown="event => handleKeyDownFilteredEmojis(event, index, emoji)"
```
braddialpad, line 85: the blue focus ring should not display on click, only when using keyboard for navigation (focus-visible style behaviour)

components/emoji_picker/modules/emoji_selector.vue
```diff
@@ -444,13 +478,202 @@ function setTabLabelObserver () {
...
+  if (event.key === 'ArrowRight') {
+    if (!focusEmoji(indexTab, indexEmoji + 1)) {
+      // Jump to the next tab
+      // handle end of tab
+      if (!focusEmoji(indexTab + 1, 0)) {
+      // when reach the end of the list, jump to the first emoji of the first tabset
+        focusEmoji(0, 0);
+      }
+    }
+  }
```
braddialpad: The logic in these two if statements is pretty similar but obviously in the other direction, is it possible to isolate this in a single function? maybe we pass in a single param `right` or `left`? can we do something similar for up/down?

components/emoji_picker/modules/emoji_selector.vue
```diff
@@ -444,13 +478,202 @@ function setTabLabelObserver () {
...
+      // if there is no previous row, we should jump to emoji in the sampe position of the previous tab
+      const previousTab = indexTab - 1 < 0 ? 0 : indexTab - 1;
+      const emojisInPreviousTab = emojiRefs.value[previousTab].length;
+      const lastEmojiPosition = emojisInPreviousTab - (emojisInPreviousTab % EMOJIS_PER_ROW) + position;
+
+      if (!focusEmoji(previousTab, lastEmojiPosition)) {
+        // if there is no emoji in this position, jump to the last emoji of the row
+        focusEmoji(indexTab - 1, emojiRefs.value[indexTab - 1].length - 1);
+      }
+    }
```
braddialpad, line 577: We are doing a lot of work in here related to calculating emoji index. Would it make things cleaner if we added a mapping of sorts for the tab start and end indexes?

something like this off the top of my head:

```js
const indexTabMap = [
  // indexTab 0
  { indexStart: 0, indexEnd: 43 },
  // indexTab 1
  { indexStart: 44, indexEnd: 106 },
  ...
]
```

components/emoji_picker/modules/emoji_selector.vue
iropolo: It will also jump to the input search when it is in the first tab

components/emoji_picker/modules/emoji_selector.vue
iropolo: Since the expected behaviour is to close the emojipicker on emoji selection I dont see this necessary.

braddialpad: This isn't always the case since it's pretty common to keep the emoji picker open when shift is held and an emoji is clicked (test on slack for example)

iropolo: hmm that seems to be another accessibility thing to handle because that behaviour will probably not be possible.
Because on `click` you emit a `close` event but you need to catch the `shift` to not emit it or ...

I dont know how the user can replicate that case in our emoji picker

components/emoji_picker/modules/emoji_selector.vue
iropolo: I updated the code with an improvement on thise arrowRight and arrowLeft.

I would like to do it the same with up and down but they are more complex...

components/emoji_picker/modules/emoji_selector.vue
iropolo: I create a composable to handle this in another file, it is not what you suggest but its a point of start...

I would like to tackle these on the improvement task..

But as you suggest, I can "hardcode" more our emojiPicker functionality, because we know which emojis have and how many... so.. If we want ... we can do this things more "easy"..

You may ask why I say "hardcode MORE our emojiPicker" the only thing it is necessary to have is the `width: 372px;` on our `.d-emoji-picker{` class, this because we force to have have 9 emoji per row, so all the following calculations will be OK, because we have our constant `EMOJIS_PER_ROW = 9`

components/emoji_picker/modules/emoji_selector.vue
braddialpad: I'm not sure we need to handle anything on our side, on the product side they could capture a shift key press event in their wrapper component and then decide whether they want to close it or not when the "selected-emoji" event triggers

components/emoji_picker/modules/emoji_selector.vue
braddialpad: yeah for sure I think it's fine to have this hardcoded for now. Could be a future upgrade to make width changable if it is actually needed by product for some reason.

And yeah this is not crucial at all for right now, could definitely be a future task

braddialpad (approved):
All my remaining issues are mostly code cleanup things that should not block this PR but could be fixed in the future, so approving this.

components/emoji_picker/modules/emoji_selector.vue
iropolo: Okay...
This is because I added this css style

` &:focus{
        box-shadow: var(--bs-focus-ring);
      }`

On the button in `.d-emoji-picker{` class... this is necessary to have because when you click on a tab this focus the first emoji of the tab, then you can navigate it directly with keyboard. (This is a design request)

Your request to remove it from the click seems like to be a side effect and I need to investigate how I can replace it witouth lost the previous desire behaviour.

braddialpad: can you just change `&:focus` to `&:focus-visible`?

https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible

iropolo: Seems to not work with focus-visible 🤔

PR #988, opened 2023-05-17
fix: set node engine minimum

braddialpad:
# fix: set node engine minimum

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

While we use node 18, it doesn't mean it's actually required to use our library. Used [ls-engines](https://www.npmjs.com/package/ls-engines) to determine what our actual node required version is and set it to that.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -82,7 +82,7 @@
         "yorkie": "^2.0.0"
       },
       "engines": {
-        "node": ">= 18"
+        "node": ">= 12.20"
       },
       "peerDependencies": {
         "@dialpad/dialtone": ">=7.26",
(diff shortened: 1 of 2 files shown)


juliodialpad (commented):
Isn't it better to keep >=16? 

I don't think we need to set it as low as 12.

This might bring us issues in the future if we add something that's newer that 12, right?

braddialpad:
> Isn't it better to keep >=16?
> 
> I don't think we need to set it as low as 12.
> 
> This might bring us issues in the future if we add something that's newer that 12, right?

It's not really accurate, and even setting it at 16 we might have that same issue in the future. However, we could probably use this library to automate a simple check for whenever our minimum version changes. I'll try doing this.

braddialpad:
Merging for now to unblock ubervoice. Will create subsequent PR for the automated check.

PR #991, opened 2023-05-18
feat(rich-text-editor): link extension

MikaDialpad:
# Add Link extension for DtRichTextEditor

- [ ] Fix
- [x] Feature
- [] Refactoring
- [x] Documentation

## :book: Description

Add a link extension for the rich text editor. The link extension will linkify content as it is typed on the editor, but only visually as the links are not interactive or clickable. The types of links it handles are URLs with and without the protocol, IP addresses, email addresses and phone numbers. The original idea was to use TipTap's link extension but it ended up not working for our purposes, so a custom build was necessary.

## :pencil: Checklist

- [ ] I have added tests

## :crystal_ball: Next Steps

Tests will be added in the next step in order to keep the size of this PR down and to give me some time to think where to test what.

image

## sources

- https://github.com/dialpad/dialtone-vue/blob/staging/components/rich_text_editor/rich_text_editor.tech_spec.md#links


MikaDialpad:
I know this is a lot to review and most of the code is specific to TipTap and ProseMirror, but let me know if you have questions and I can provide more info.

common/utils.js
```diff
@@ -204,6 +204,124 @@ export function isOutOfViewPort (element) {
   isOut.all = Object.values(isOut).every(val => val);
   return isOut;
 }
+
```
MikaDialpad, line 191: This file has just the regexes, which are equivalent to Dialpad backend, and few of the utils that aren't directly used by this feature but were added by Brad's suggestion

components/rich_text_editor/extensions/link/autolink.js
```diff
@@ -0,0 +1,87 @@
+import {
```
MikaDialpad, line 1: This is mostly boilerplate code for the ProseMirror plugin that is able to manipulate the content more than the TipTap extension could on its own.

components/rich_text_editor/extensions/link/link.js
```diff
@@ -0,0 +1,36 @@
+import {
```
MikaDialpad, line 1: This is the actual extension code, which is mostly showing that all the functionality comes from the ProseMirror plugin

components/rich_text_editor/extensions/link/utils.js
```diff
@@ -0,0 +1,169 @@
+import { getMarksBetween } from '@tiptap/core';
```
MikaDialpad, line 1: These are utils for the autolink plugin and contain the bulk of the logic. I wrote them in a semi-generic fashion, but since they're pretty specific to the autolink functionality I didn't want to go overboard and make everything into a fully generic util. We can pull some out into the common util file if they seem useful for other purposes and/or perhaps in the future we'll want to build similar plugins that could share some of the utils, but for now I don't know about such needs.

MikaDialpad:
Hey @braddialpad just checking if you've had a chance to take a look at this or have any questions?

common/utils.js
```diff
@@ -188,6 +188,124 @@ export function isOutOfViewPort (element) {
...
+// match phone numbers, e.g. "765-8813", "(778) 765-8813" or "+17787658813".
+// create it as a function so the length can be adjusted
+export function getPhoneNumberRegex (minLength = 7, maxLength = 15) {
+  return new RegExp(
+    '(?:^|(?<=\\W))(?![\\s\\-])\\+?(?:[0-9()\\- \\t]' +
+    `{${minLength},${maxLength}}` +
+    ')(?=\\b)(?=\\W(?=\\W|$)|\\s|$)',
+  );
+}
```
braddialpad: We should use [jsdoc comments](https://jsdoc.app/) for functions such as this, looks like you did for the others, just forgot this one.

components/rich_text_editor/extensions/link/link.js
braddialpad: Might as well write the above in the actual file as a comment, as it is a little confusing what the overall purpose of this file is if you're not very familiar with tiptip / prosemirror.

Just so I know I am understanding correctly, this is defining a tiptap plugin, which for the most part is a passthrough to a prosemirror plugin? I'll admit I am still a bit confused between the relationship between `link` and `autolink`

components/rich_text_editor/extensions/link/autolink.js
braddialpad: So did you actually write this code or you mostly took it from an example?

components/rich_text_editor/extensions/link/utils.js
braddialpad: Yep I would agree to keep them here until needed elsewhere.

components/rich_text_editor/extensions/link/utils.js
```diff
@@ -0,0 +1,189 @@
...
+  // If the word doesn't match the regex we can just return it.
+  if (!regex.test(word.text)) {
+    return word;
+  }
+
+  // Depending on the direction take one step to the left or right to find the
+  // preceding or following word.
+  const newIndex = direction === 'left' ? word.from - 1 : word.to + 1;
+
+  // Prevent an infinite loop for the edge cases.
```
braddialpad: sorry to be a terminology snob but I believe the correct term is "base cases" in a recursive function.

components/rich_text_editor/extensions/link/utils.js
```diff
@@ -0,0 +1,189 @@
...
+    to: right + index,
+  };
+}
+
+/**
+ * Helper to check if a word at a given index matches a regex and if true, finds
+ * the previous or next word until the regex doesn't match anymore. Useful to
+ * find for example the entire phone number when it is separated by whitespace.
+ */
+export function getWordAtUntil (text, index, direction, regex) {
```
braddialpad, line 88: pretty interesting function!

components/rich_text_editor/rich_text_editor.vue
```diff
@@ -27,7 +28,7 @@ export default {
      * Value of the input
      */
     value: {
-      type: String,
+      type: [Object, String],
```
braddialpad, line 32: Should we document what the object form of value is?

braddialpad:
> Hey @braddialpad just checking if you've had a chance to take a look at this or have any questions?

Sorry for the delay, was deep in dark mode stuff the last while.

components/rich_text_editor/extensions/link/link.js
MikaDialpad: Sure, I'll add the comment 👍

TipTap extensions like this one mostly defines the look of the node/mark, like how in this case we're rendering an `a` tag with some attributes. TipTap has exposed some ways to control the behavior of the extension also (a.k.a when to actually wrap text with it), but they're just regex rules that don't work for anything more complex. For that they've added a way to add ProseMirror plugins that allow way more detailed control over the extension (but are more complex to write in general). So in this case we really just want to wrap some text with `a` tags, but the logic for when and what to wrap all comes from the ProseMirror plugin.

components/rich_text_editor/rich_text_editor.vue
MikaDialpad: Hmm, yeah I'll add a comment here. Hopefully no-one tries to construct it manually though as it's a specific format created by TipTap/ProseMirror

components/rich_text_editor/extensions/link/autolink.js
MikaDialpad: The link extension that TipTap provides has an "autolink" plugin that I did study to learn how these things work. There are some parts that look similar, but I believe they are mostly boilerplate stuff that most plugins would have no matter what. So while I did write this myself I did take inspiration from the other plugin. Do you think it's worth mentioning here? Shouldn't have issues with the [license](https://github.com/ueberdosis/tiptap/blob/main/LICENSE.md) regardless

MikaDialpad (commented):
Thanks Brad for taking a look! I know this is not easy code to review since it's very specific to TipTap/ProseMirror, but I appreciate another set of eyes on it.

components/rich_text_editor/extensions/link/autolink.js
braddialpad: Yeah wasn't really worried about the licensing, just didn't think there would be a lot that was useful to review if it was pretty standard boilerplate.

braddialpad (approved):
There are definitely some concepts and such I don't fully understand here, however the overall code structure looks well organized, and the preview example seems to work pretty flawlessly. Don't think I have any further changes here.

PR #992, opened 2023-05-19
fix(dt-scroller): fix dynamic scroller not passing fieldKey properly

AlexLamDialpad:
# PR Title

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Field key was still being assigned to id for the dynamic scroller
https://dialpad.atlassian.net/browse/DIALTONE-1081

## :pencil: Checklist

- [ ] I have added tests


iropolo:
hmm... seems like its a bug in the core-scroller ... to be specific, is in the _updateVisibleItems function in line code 448

```
const key = keyField ? item?.[keyField] : item;

    if (key == null) {
      throw new Error(`Key is ${key} on item (keyField is '${keyField}')`);
    }
```

I did a check on the vue-virtual-scroller library and I set it up locally to replicate this, and it seems its a issue in they side.
This is not an issue on our modifications or some change we did...

So I guess this is a valid fix.

However I suggest to do it in this way:

 ```
result.push({
          item,
          [keyField]: id,
          size,
        });
```



image

image

File: components/scroller/modules/dynamic_scroller.vue [hunk 1]
```diff
@@ -178,11 +177,12 @@ export default {
...
         }
-        result.push({
+        const data = {
           item,
-          id,
           size,
-        });
+        };
+        data[keyField] = id;
+        result.push(data);
```
iropolo: Suggested change:

```
 result.push({
          item,
          [keyField]: id,
          size,
        });
```

iropolo (changes requested):

iropolo:
Look good, thanks! 👍🏼

iropolo (approved):

PR #1002, opened 2023-05-30
feat: state changes for feed item row

sanjeevdialpad:
# State changes for Feed item Row

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Feed item row should define state for normal, searched and error. Further have a boolean property for fade that transitions the background color back to normal.

## :pencil: Checklist

- [ ] I have added tests

## :crystal_ball: Next Steps

Unit tests in a separate PR

## :link: Sources

Figma: https://www.figma.com/file/GJiRmWORmDi0zxLJfwn8yp/DP-2.0---Conversation-View?type=design&node-id=3-20503&t=gpfjBAhBBdIXICDB-0


recipes/conversation_view/feed_item_row/feed_item_row.vue
```diff
@@ -157,11 +157,21 @@ export default {
...
-    // state: {
-    //   type: String,
-    //   default: 'normal',
-    // },
+    /**
+     * state for the feed item row. Can be default, searched & error
+     */
+    state: {
+      type: String,
+      default: 'NORMAL',
```
braddialpad: Need to add a constant for the allowed state values, and constrain this prop to only accept the allowed values.

recipes/conversation_view/feed_item_row/feed_item_row.vue
```diff
@@ -157,11 +157,21 @@ export default {
...
+    },
+
+    /**
+     * Whether to fade the background color to default
+     */
+    fade: {
+      type: Boolean,
+      default: false,
+    },
```
braddialpad: I am curious how this would actually be triggered on the product side, since this just triggers on mount is there a case when we might not want to fade this immediately, but rather when the specific feed item row becomes visible? After it has initially faded, there is no possible way to fade it again right? is there any case now or possibly in the future where we would need to do this?

The other possible way which might allow for more control is to just create a function that can be called via a component ref. We use this method sometimes for things like `setFocus()` and such. I think it makes a bit more sense for a "single trigger" type thing such as this.

recipes/conversation_view/feed_item_row/feed_item_row.vue
```diff
@@ -198,6 +209,29 @@ export default {
...
+  },
+
+  mounted () {
+    if (!this.fade) {
+      return;
+    }
+    setTimeout(() => {
+      this.$refs.FeedItemRef.$el.classList.remove(FEED_ROW_STATE_BACKGROUND_COLOR[this.state]);
+    }, 1500);
```
braddialpad: I'd recommend storing the return value in a data property and calling clearTimeout on it when the component unmounts otherwise it may be possible to get an error if the timer hasn't triggered yet when the component unmounts.

braddialpad (changes requested):

recipes/conversation_view/feed_item_row/feed_item_row.vue
iropolo: Hmm
I put a true value on fade prop and then 1500ms later it says:
`Failed to execute 'remove' on 'DOMTokenList': The token provided must not be empty.`
 
How I can test this? did I miss sth?

sanjeevdialpad: i didnt handle the case where state is normal, then no background is applied and hence no fade can occur. Will change sorry!

iropolo (approved):
Thanks for the update 🚀 ! LGTM

recipes/conversation_view/feed_item_row/feed_item_row.vue
sanjeevdialpad: so i removed setTimeout since now it uses method to do the fade.

braddialpad (approved):
Thanks!

PR #1004, opened 2023-06-01
fix: add placeholder text for richtexteditor

sanjeevdialpad:
# Placeholder Text for RichTextEditor

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :pencil: Checklist

- [ ] I have added tests

### DIFF
--- a/components/rich_text_editor/rich_text_editor.vue
+++ b/components/rich_text_editor/rich_text_editor.vue
@@ -9,6 +9,7 @@
 import { Editor, EditorContent } from '@tiptap/vue-2';
 import Document from '@tiptap/extension-document';
 import Paragraph from '@tiptap/extension-paragraph';
+import Placeholder from '@tiptap/extension-placeholder';
 import Text from '@tiptap/extension-text';
 import Link from './extensions/link';
 import {
@@ -102,6 +103,14 @@ export default {
       type: [Boolean, Object],
       default: false,
     },
+
+    /**
+     * Placeholder text
+     */
+    placeholderText: {
+      type: String,
+      default: 'Type here...',
+    },
   },
 
   emits: [
@@ -147,6 +156,13 @@ export default {
       if (this.link) {
         extensions.push(this.getExtension(Link, this.link));
       }
+      // Enable placeholderText
+      extensions.push(
+        this.getExtension(
+          Placeholder,
+          { placeholder: this.placeholderText },
+        ),
+      );
       return extensions;
     },
 
@@ -283,3 +299,13 @@ export default {
   },
 };
 </script>
+
+<style lang="less">
+.ProseMirror p.is-editor-empty:first-child::before {
+    content: attr(data-placeholder);
+    float: left;
+    color: #adb5bd;
+    pointer-events: none;
+    height: 0;
+  }
+</style>
--- a/components/rich_text_editor/rich_text_editor.vue
+++ b/components/rich_text_editor/rich_text_editor.vue
@@ -109,7 +109,7 @@ export default {
      */
     placeholder: {
       type: String,
-      default: 'Type here...',
+      default: '',
     },
   },
 
--- a/components/rich_text_editor/rich_text_editor.vue
+++ b/components/rich_text_editor/rich_text_editor.vue
@@ -107,7 +107,7 @@ export default {
     /**
      * Placeholder text
      */
-    placeholderText: {
+    placeholder: {
       type: String,
       default: 'Type here...',
     },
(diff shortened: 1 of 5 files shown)


components/rich_text_editor/rich_text_editor.vue
```diff
@@ -102,6 +103,14 @@ export default {
       type: [Boolean, Object],
       default: false,
     },
+
+    /**
+     * Placeholder text
+     */
+    placeholderText: {
```
MikaDialpad: I would call this just `placeholder` to align with the native input's attribute.

components/rich_text_editor/rich_text_editor.vue
```diff
@@ -283,3 +299,13 @@ export default {
   },
 };
 </script>
+
+<style lang="less">
+.ProseMirror p.is-editor-empty:first-child::before {
+    content: attr(data-placeholder);
+    float: left;
+    color: #adb5bd;
```
MikaDialpad: This should most likely use `color: var(--fc-placeholder);` that DtInput uses

components/rich_text_editor/rich_text_editor.vue
```diff
@@ -283,3 +299,13 @@ export default {
   },
 };
 </script>
+
+<style lang="less">
```
MikaDialpad: We can remove `lang="less"` since this is pure css

components/rich_text_editor/rich_text_editor.vue
```diff
@@ -147,6 +156,13 @@ export default {
       if (this.link) {
         extensions.push(this.getExtension(Link, this.link));
       }
+      // Enable placeholderText
+      extensions.push(
+        this.getExtension(
+          Placeholder,
+          { placeholder: this.placeholderText },
+        ),
+      );
```
MikaDialpad: `this.getExtension` was a helper to handle both booleans and options passed for the extensions, but since we always want to pass options to this one we don't need to use that
```
extensions.push(
  Placeholder.configure({ placeholder: this.placeholder }),
);
```

components/rich_text_editor/rich_text_editor.vue
```diff
@@ -283,3 +299,13 @@ export default {
   },
 };
 </script>
+
+<style lang="less">
+.ProseMirror p.is-editor-empty:first-child::before {
```
MikaDialpad: Hmm, I see that this is what the extension creators are suggesting, so I guess it's fine... The indentation looks off though.

MikaDialpad (changes requested):
Oh geez, TipTap really loves their extensions for EVERYTHING lol. Thanks for spotting this Sanjeev, I missed it while building this

components/rich_text_editor/rich_text_editor.vue
sanjeevdialpad: ty good catch!

components/rich_text_editor/rich_text_editor.vue
sanjeevdialpad: ohk, i just thought its a wrapper method made for this purpose. changed

components/rich_text_editor/rich_text_editor_default.story.vue
```diff
@@ -7,6 +7,7 @@
     :output-format="outputFormat"
     :auto-focus="autoFocus"
     :link="link"
+    :placeholder-text="placeholderText"
```
iropolo: Should be `placeholder`? 🤔

sanjeevdialpad: changed.

iropolo (approved):
LGTM 🚀

components/rich_text_editor/rich_text_editor.vue
```diff
@@ -102,6 +103,14 @@ export default {
       type: [Boolean, Object],
       default: false,
     },
+
+    /**
+     * Placeholder text
+     */
+    placeholder: {
+      type: String,
+      default: 'Type here...',
```
MikaDialpad: Hmm, now I'm thinking that probably this should be left empty by default as maybe we don't always want to show it and also since the default is in English only.

MikaDialpad (approved):
Thanks Sanjeev!

components/rich_text_editor/rich_text_editor.vue
sanjeevdialpad: i agree. changed

PR #1007, opened 2023-06-07
feat(datepicker): datepicker component

iropolo:
# Datepicker component

- [x] Feature

## :book: Description

Create stubs for datepicker component based on the details we have groomed in the epic https://dialpad.atlassian.net/browse/DT-1098

On this stub I will provide basic structure and components on how we plan to take the Datepicker.
Styles and details will be tackle on the following tickets of the epic.


components/datepicker/datepicker.vue
```diff
@@ -0,0 +1,95 @@
...
+    },
+
+    prevYearLabel: {
+      type: String,
+      required: true,
+    },
+
+    nextYearLabel: {
+      type: String,
+      required: true,
```
juliodialpad:
```suggestion
    prevMonthLabel: {
      type: String,
      default: 'Previous month',
    },

    nextMonthLabel: {
      type: String,
      default: 'Next month',
    },

    prevYearLabel: {
      type: String,
      default: 'Previous year',
    },

    nextYearLabel: {
      type: String,
      default: 'Next year',
```

juliodialpad: Try this solution, this will pass the tests and still enable the user to change the labels if needed.

components/datepicker/modules/month-year-picker.vue
```diff
@@ -0,0 +1,150 @@
...
+    <div>
+      <button
+        type="button"
+        :aria-label="prevYearLabel"
+        @click="handleYear(false)"
+      >
+        <dt-icon
+          name="chevrons-left"
+          size="300"
+          :description="prevYearLabel"
```
juliodialpad: This isn't a `dt-icon` prop, so this wouldn't do anything, you can remove it.

components/datepicker/datepicker._test.js
juliodialpad, line 1: Change this file name to `datepicker.test.js`

components/datepicker/datepicker.stories.js
```diff
@@ -0,0 +1,104 @@
...
+);
+
+// Stories
+export const Default = {
+  render: DefaultTemplate,
+  args: {
+    prevMonthLabel: 'Previous month',
+    nextMonthLabel: 'Next month',
+    prevYearLabel: 'Previous year',
+    nextYearLabel: 'Next year',
```
juliodialpad: You can remove this as they'll have default values

components/datepicker/datepicker._test.js
iropolo: I renamed the file to avoid trigger the test in the actions, because it is not done yet.
It should be tackled in the according ticket

components/datepicker/modules/month-year-picker.vue
iropolo: `test a11y fix` leftover

components/datepicker/datepicker.vue
iropolo: Already fixed, it was necessary to declare the props in the storybook

```
// Props
  prevMonthLabel: {
    control: 'text',
    table: {
      type: {
        summary: 'VNode',
      },
    },
  },
  nextMonthLabel: {
    control: 'text',
    table: {
      type: {
        summary: 'VNode',
      },
    },
  },

  prevYearLabel: {
    control: 'text',
    table: {
      type: {
        summary: 'VNode',
      },
    },
  },

  nextYearLabel: {
    control: 'text',
    table: {
      type: {
        summary: 'VNode',
      },
    },
  },
```

components/datepicker/datepicker.stories.js
iropolo: test a11y fix leftover

components/datepicker/modules/month-year-picker.vue
```diff
@@ -0,0 +1,146 @@
+<template>
+  <div class="d-datepicker__month-year-picker">
```
juliodialpad, line 2: make sure to use [BEM](https://getbem.com/introduction/) class naming conventions.

You could rename it to something like `d-date-picker--header`

components/datepicker/modules/month-year-picker.vue
juliodialpad, line 1: This should be named: `date-picker-header` according to what we agreed.

It's more generic as it could include stuff other than just month and year picker in the future.

components/datepicker/modules/month-year-picker.vue
```diff
@@ -0,0 +1,146 @@
...
+  },
+
+  methods: {
+    handleMonth (isNext = false) {
+      const initialDate = set(this.selectedDate, { month: this.month, year: this.year });
+      const date = isNext ? addMonths(initialDate, 1) : subMonths(initialDate, 1);
+
+      this.month = getMonth(date);
+      this.$emit('calendar-days', this.calendarDays);
+    },
```
juliodialpad: if you change the param to and integer instead, you could do something like `addMonths(initialDate, param)` if the param is negative date-fns is perfectly capable of handling it and the code is more redeable.

components/datepicker/modules/month-year-picker.vue
```diff
@@ -0,0 +1,146 @@
...
+  },
+
+  mounted () {
+    this.$emit('calendar-days', this.calendarDays);
+  },
+
+  methods: {
+    handleMonth (isNext = false) {
+      const initialDate = set(this.selectedDate, { month: this.month, year: this.year });
```
juliodialpad: Why are you setting the initialDate here? It's not too transparent.

If this doesn't need to be reactive, you could declare it on `mounted`.

components/datepicker/modules/month-year-picker.vue
```diff
@@ -0,0 +1,146 @@
...
+      return getCalendarDays(this.month, this.year);
+    },
+
+    getMonth () {
+      return format(new Date(2000, this.month, 1), 'MMMM');
+    },
+  },
+
+  mounted () {
+    this.$emit('calendar-days', this.calendarDays);
```
juliodialpad: Maybe we can rewrite this into a watcher and emit it when changing the `year` and the `month`

components/datepicker/modules/month-year-picker.vue
```diff
@@ -0,0 +1,146 @@
...
+  },
+
+  computed: {
+    // Get days for the currently selected month and year
+    calendarDays () {
+      return getCalendarDays(this.month, this.year);
+    },
+
+    getMonth () {
+      return format(new Date(2000, this.month, 1), 'MMMM');
```
juliodialpad: Why are you setting the year to '2000' you could use the `initialDate` variable once properly initialized.

components/datepicker/modules/month-year-picker.vue
juliodialpad, line 1: I don't like `handleYear` and `handleMonth` functions, they're not too transparent about what the param means, I'd prefer to have an integer value or change the function naming to something more self-descriptive.

components/datepicker/datepicker.vue
```diff
@@ -0,0 +1,125 @@
+<template>
+  <div class="d-datepicker">
+    <div class="d-datepicker--header">
```
juliodialpad, line 3: this `<div>` is not necessary as the `month-year-picker` component should be already wrapped with it.

components/datepicker/datepicker.vue
```diff
@@ -0,0 +1,125 @@
...
+      <month-year-picker
+        :prev-month-label="prevMonthLabel"
+        :next-month-label="nextMonthLabel"
+        :prev-year-label="prevYearLabel"
+        :next-year-label="nextYearLabel"
+        :selected-date="selectedDate"
+        @calendar-days="updateCalendarDays"
+      />
+    </div>
+    <div class="d-datepicker--body">
```
juliodialpad, line 13: Same here, the `calendar` should be already wrapped this `<div>` inside the component.

components/datepicker/modules/calendar.vue
```diff
@@ -0,0 +1,73 @@
+<template>
+  <div class="d-datepicker__calendar">
```
juliodialpad, line 2:
```suggestion
  <div class="d-date-picker--calendar">
```

components/datepicker/modules/calendar.vue
```diff
@@ -0,0 +1,73 @@
...
+     *
+     * @event select-date
+     * @type {Date}
+     */
+    'select-date',
+  ],
+
+  data () {
+    return {
+      thisCalendarDays: this.calendarDays,
```
juliodialpad: Is this necessary?

components/datepicker/modules/calendar.vue
```diff
@@ -0,0 +1,73 @@
...
+    };
+  },
+
+  computed: {
+    weekDays () {
+      return getDayNames(this.locale, WEEK_START);
```
juliodialpad: I think this is a better name to that function

```suggestion
      return getWeekDayNames(this.locale, WEEK_START);
```

components/datepicker/modules/calendar.vue
```diff
@@ -0,0 +1,73 @@
...
+    },
+  },
+
+  watch: {},
+
+  methods: {
+    selectDay (day) {
+      if (!day.current) { return; }
```
juliodialpad: What's the problem by emitting the current day?

components/datepicker/modules/calendar.vue
```diff
@@ -0,0 +1,73 @@
...
+      :key="indexWeek"
+    >
+      {{ indexWeek }}
+      <button
+        v-for="(day, indexDays) in week.days"
+        :key="indexWeek + indexDays"
+        type="button"
+        :aria-label="day.text"
+        @click="selectDay(day)"
+        @keydown="$emit('select-date', day.value)"
```
juliodialpad: This is not right, both methods should do the same.

But, we should only trigger on `click` and `keydown.space`, right?

components/datepicker/datepicker.stories.js
```diff
@@ -0,0 +1,89 @@
+import { createTemplateFromVueFile } from '@/common/storybook_utils';
+import { action } from '@storybook/addon-actions';
+import DtDatepicker from './datepicker.vue';
+import DtDatepickerTemplate from './datepicker_default.story.vue';
```
juliodialpad:
```suggestion
import DtDatepickerDefaultTemplate from './datepicker_default.story.vue';
```

Just in case we need to import another templates for variants or something like so.

components/datepicker/datepicker.stories.js
```diff
@@ -0,0 +1,89 @@
...
+// Stories
+export const Default = {
+  render: Template,
+
+  args: {
+    prevMonthLabel: 'Previous month',
+    nextMonthLabel: 'Next month',
+    prevYearLabel: 'Previous year',
+    nextYearLabel: 'Next year',
+  },
```
juliodialpad: You can move this to `argsData` as we'll need them on every date-picker story.

components/datepicker/datepicker_default.story.vue
```diff
@@ -0,0 +1,27 @@
+<template>
+  <div>
```
juliodialpad: Not necessary to wrap this on a `<div>`

components/datepicker/utils.js
```diff
@@ -0,0 +1,95 @@
+import {
+  startOfWeek, addDays, getMonth, isEqual, setHours,
+  setMinutes,
+  setSeconds,
+  setMilliseconds,
+} from 'date-fns';
+import { WEEK_START } from '@/components/datepicker/datepicker_constants.js';
+
+const resetDateTime = (value) => {
+  let dateParse = getDate(JSON.parse(JSON.stringify(value)));
```
juliodialpad: Why are you doing this `JSON.parse(JSON.stringify(date))`?

components/datepicker/utils.js
```diff
@@ -0,0 +1,95 @@
...
+
+  addDaysToWeek(firstDateInCalendar);
+
+  return weeks;
+};
+
+/**
+ * Generate week day names based on locale and in order specified in week start
+ */
+export const getDayNames = (locale, weekStart) => {
```
juliodialpad:
```suggestion
export const getWeekDayNames = (locale, weekStart) => {
```

components/datepicker/utils.js
juliodialpad, line 1: We should remove all the timezones information for now, we are getting into that later when we make the `time-picker` now it's unnecessary and complicate the process.

juliodialpad (changes requested):
Nice work here @iropolo!

This is working as expected, requesting changes mainly on code style to improve readability.

Let's rename the parent folder and files to `date_picker` instead of `datepicker`

components/datepicker/modules/month-year-picker.vue
iropolo: Not sure to renaming datepicker to date-picker 
it is `datepicker` in material ui, buefy, ant design, flowbite(tailwind and react), bootstrap-vue.

And I remember developing it like this 🤔 

About the BEM convention here, I replicate the structure of emoji-picker.

The parent component has `--header` and `--body`, the childs will have `__child-component`.

I already have this discussion with Brad in the previous component and for me its works as well and I dont see a relevant change here.

components/datepicker/modules/month-year-picker.vue
iropolo: Same explanation above, its an already discussed structure and I don't see a relevant change here.

components/datepicker/modules/month-year-picker.vue
iropolo: Not sure about this one, how it would be? I didn't understand

components/datepicker/modules/month-year-picker.vue
iropolo: It is necessary to have an updated `const date`, I don't see it necessary to have it on mounted because it is not shared anywhere.

components/datepicker/modules/month-year-picker.vue
iropolo: I don't see necessary to have a watcher looking around... the benefit is reduce the number of times its beeing called.
I will have it in mind later

components/datepicker/modules/month-year-picker.vue
iropolo: Is it just a fixed year and day to extract the month info. I don't see it necessary to be updated with `initialDate`

components/datepicker/modules/month-year-picker.vue
iropolo: I didn't understand the suggestion previously, could you describe it how it would be?

components/datepicker/datepicker.vue
iropolo: This is related the emoji-picker structure, I just did it in the same way.
Anyway the reason of this is give flexibility and control on the css styles.

components/datepicker/datepicker.vue
iropolo: same response as above...

```
This is related the emoji-picker structure, I just did it in the same way.
Anyway the reason of this is give flexibility and control on the css styles.
```
This structure response the idea of header-body-footer later if we want to add time or another update you already have the structure in place.

components/datepicker/modules/calendar.vue
iropolo: Not sure about this, it can be possible but on emoji-picker the only component which has -- are the `--header` `--body` and `--footer`.

components/datepicker/modules/calendar.vue
iropolo: Certainly not... seems to be a leftover.

components/datepicker/modules/calendar.vue
iropolo: I can take it 👍🏼

components/datepicker/modules/calendar.vue
iropolo: This is because `current` means the current month.
We are placing all days in the `calendar` components.

Screenshot 2023-06-13 at 12 02 37


Since we are not handle the automatic month change if you click those days, then it should not be clickable.
These days will be disabled with a gray color (I'm assuming this because the design did not handle this and we didnt groom this behaviour). Something like this:

image

components/datepicker/modules/calendar.vue
iropolo: Yes this is right, this should be improved in the respective accessibility ticket since this is a stub only but I can do it now.

And yes it should be triggered on `click` and `keydown.space` or even `.enter`

components/datepicker/datepicker.stories.js
iropolo: Not necessary to do it now but I will added it

components/datepicker/utils.js
iropolo: I removed the resetDateTime function 👍🏼

components/datepicker/utils.js
iropolo: Ye I did it initially to have a sanitized dates because Im not sure but seems to works well without it so.. I removed it

components/datepicker/modules/calendar.vue
iropolo: I updated `current` to `currentMonth` to be more descriptive

components/datepicker/modules/month-year-picker.vue
iropolo: I updated it to a watcher for month and year.

components/datepicker/modules/calendar.vue
juliodialpad: Exactly, the `calendar` part is like the `body` of the calendar component, that's why I suggested this change

components/datepicker/modules/calendar.vue
```diff
@@ -1,18 +1,33 @@
...
+        :class="{
+          'd-datepicker__day--disabled': !day.currentMonth,
+          'd-datepicker__day--selected': highlightDay ? ((day.text === highlightDay) && day.currentMonth) : day.selected,
+        }"
         type="button"
         :aria-label="day.text"
         @click="selectDay(day)"
-        @keydown="$emit('select-date', day.value)"
+        @keydown.enter="selectDay(day)"
+        @keydown.space="selectDay(day)"
```
juliodialpad: Maybe we can extract this into listeners? to avoid repeating the code

components/datepicker/modules/month-year-picker.vue
juliodialpad: It'd be better if you handle it with something like `changeYear(1 | -1)`, `incrementYear()`/`decreaseYear()` or even better directly just increase the current year value on the click event.

The `handleYear` function is not good because 'handle' could have a lot of interpretations and you need to actually go to the function to know what it is doing instead of being naturally read and understood at a glance.

components/datepicker/modules/month-year-picker.vue
juliodialpad: We're starting to change the conventions here, but if @braddialpad it's ok with that I have no problem with it. Was just pointing out that this and `emoji_picker` components are not following the same folder structures that we were following previously.

juliodialpad (approved):
Really nice improvement on the design side!

I'm good with the progress here, this far than good for a stub, thanks!

components/datepicker/modules/calendar.vue
iropolo: I will comment this in the code and will be improved in the respective ticket

components/datepicker/modules/month-year-picker.vue
iropolo: Gotcha, lemme take a look on this

components/datepicker/modules/month-year-picker.vue
braddialpad: Couple comments on the BEM structure. datepicker as a single word for component name is fine as long as it is consistent (for example the component is named "Datepicker" instead of "Date Picker" in storybook). 

I think we are using `--` incorrectly here for header and body. header and body are children of `d-datepicker` so it should be `d-datepicker__header` `d-datepicker__body`. `--` (modifier) is for specific state changes only. Ex: `d-datepicker__header--large`

for folder conventions, I don't think we need the modules folder. We only had the modules folder on scroller to separate 3rd party library code from component code. In this case it is all component code. More importantly I would ask that any subcomponents are prefixed with the component name to be more searchable.

ex: instead of `calendar.vue` -> `datepicker_calendar.vue`
instead of `month-year-picker.vue` -> `datepicker_month_year_picker.vue`

And yeah filenames should always use underscores as is our convention so `month-year-picker` should use `_` instead of `-`.

PR #1009, opened 2023-06-08
feat(modal): show event and initial-focus-element

juliodialpad:
# Modal - Emit Show event when opened and add initialFocusElement prop.

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

- Added the emit of update:show event after modal opened (previously were only emitted on close)
- Added initialFocusElement prop to enable the control of the initially focused element.

## :bulb: Context

This were missing features we had reported long time ago, adding them.

## :pencil: Checklist

- [ ] I have added tests

## :crystal_ball: Next Steps

Migrate to Vue 3 and dialtone8 branches

### DIFF
--- a/common/mixins/modal.js
+++ b/common/mixins/modal.js
@@ -35,9 +35,23 @@ export default {
      * @param {object} el - optional - ref of dom element to trap focus on.
      *  will default to the root node of the vue component
      */
-    async focusFirstElement (el) {
+    async focusFirstElement (el = this.$el) {
       const elToFocus = await this.getFirstFocusableElement(el);
       elToFocus?.focus({ preventScroll: true });
+    },
+
+    async focusElementById (elementId) {
+      await this.$nextTick();
+      const result = this.$el.querySelector(elementId);
+      if (result) {
+        result.focus();
+        return;
+      }
+
+      // eslint-disable-next-line no-console
+      console.warn('Could not find the element specified in dt-modal prop "initialFocusElement". ' +
+          'Defaulting to focusing the first element.');
+      await this.focusFirstElement();
     },
 
     /**
--- a/common/mixins/modal.js
+++ b/common/mixins/modal.js
@@ -42,7 +42,7 @@ export default {
 
     async focusElementById (elementId) {
       await this.$nextTick();
-      const result = this.$el.querySelector(elementId);
+      const result = this.$el?.querySelector(elementId);
       if (result) {
         result.focus();
         return;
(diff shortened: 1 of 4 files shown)


common/mixins/modal.js
```diff
@@ -35,11 +35,25 @@ export default {
...
+        result.focus();
+        return;
+      }
+
+      // eslint-disable-next-line no-console
+      console.warn('Could not find the element specified in dt-modal prop "initialFocusElement". ' +
+          'Defaulting to focusing the first element.');
+      await this.focusFirstElement();
+    },
+
```
iropolo: Code suggestion:

```
async focusElementById (elementId) {
      await this.$nextTick();

      try {
        const result = this.$el.querySelector(elementId);

        if (result) {
          result.focus();
          return;
        }
      } catch (error) {
        console.error(`Error occurred while executing querySelector: ${error}`);
      }

      // eslint-disable-next-line no-console
      console.warn('Could not find the element specified in dt-modal prop "initialFocusElement". ' +
          'Defaulting to focusing the first element.');
      await this.focusFirstElement();
    },
```

iropolo: I modify the `initialFocusElement` prop with a "#" in the controls to trigger the `else if (this.initialFocusElement.startsWith('#')) {` 
then the component crashes.... Just in case maybe a try/catch and a console.error handle this properly

iropolo (approved):
As far as I checked seems to be good and works properly :), LGTM 🚀

common/mixins/modal.js
juliodialpad: I don't think it's necessary to handle that error, it's simplier to do this.

`const result = this.$el?.querySelector(elementId);`

That way if for some reason the $el doesn't exist it will be undefined and focus the first element by default.

What was the case you were thinking that error could be triggered?

juliodialpad: > I modify the `initialFocusElement` prop with a "#" in the controls to trigger the `else if (this.initialFocusElement.startsWith('#')) {` then the component crashes.... Just in case maybe a try/catch and a console.error handle this properly

Did you set the property to just #button or "#button" it should be wrapped by quotes in order to properly work for storybook.

iropolo: Ye, right, that can be also a possible solution , it is working well with that.

Y set the property to only "#" and its crash, but now its works (cos the optional channing)

iropolo (approved):

PR #1019, opened 2023-06-13
feat(DT-1094): handle character limit logic in Message Input Recipe

sanjeevdialpad:
# Character Limit Logic in message Input Recipe

- [x] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

MessageInput Recipe has the ability to enable character limit for use cases like SMS. It has the following props: CharacterLimitCount which is the max number of characters allowed and the warningCount which is the counter after which the warning appears.

Submit button would be disabled after the limit is reached.

## :pencil: Checklist

- [ ] I have added tests


iropolo (changes requested):
Looks goods and it is working as expected, but the controls of storybook are not sync with the component.

Are these:

- characterLimitCount
- characterLimitWarning
- hasCharacterLimit

sanjeevdialpad:
@iropolo My bad, i didnt put the changes on default story. fixed it.

iropolo (approved):
Great!, Looks good 🚀

PR #1022, opened 2023-06-15
fix(general-row): nested a tags

juliodialpad:
# Fix (General row) Remove nested <a> tags

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Moved `dt-leftbar-row__omega` out of the `<a class="dt-leftbar-row__primary" ...>` parent to avoid nesting <a> within <a> tags.

## :bulb: Context

Is not correct to nest <a> tags within <button> or <button> within <button> tags, this will prevent such situations.

## :pencil: Checklist

- [ ] I have added tests

image
image
image

### DIFF
--- a/recipes/leftbar/general_row/general_row.vue
+++ b/recipes/leftbar/general_row/general_row.vue
@@ -106,8 +106,6 @@
               :aria-label="callButtonTooltip"
               @focus="actionFocused = true"
               @blur="actionFocused = false"
-              @mouseover="actionFocused = true"
-              @mouseleave="actionFocused = false"
               @click.stop="$emit('call', $event)"
             >
               <template #icon>
--- a/recipes/leftbar/general_row/general_row.vue
+++ b/recipes/leftbar/general_row/general_row.vue
@@ -46,6 +46,7 @@
       </div>
     </a>
     <div
+      v-if="hasActions"
       class="dt-leftbar-row__omega"
     >
       <dt-tooltip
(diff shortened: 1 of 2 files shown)


recipes/leftbar/general_row/general_row.vue
```diff
@@ -44,79 +44,82 @@
...
+          >
+            {{ unreadCount }}
+          </dt-badge>
+        </template>
+      </dt-tooltip>
+      <div
+        v-if="hasCallButton"
+        class="dt-leftbar-row__action"
+        data-qa="dt-leftbar-row-action"
+      >
```
braddialpad, line 94: tabbing to the call button and then mousing over the button causes the badge to show up behind
Screenshot 2023-06-15 at 3 04 31 PM

braddialpad: actually just realized this was a problem before this change as well... might just be the storybook example?

recipes/leftbar/style/leftbar_row.less
```diff
@@ -33,6 +33,14 @@
   //  ----------------------------------------------------------------------------
   position: relative;
   opacity: var(--leftbar-row-opacity);
+  display: flex;
+  background-color: var(--leftbar-row-color-background);
+  border-radius: var(--leftbar-row-radius);
```
braddialpad, line 38: We seem to have lost row focus styling.

before:
Screenshot 2023-06-15 at 3 08 23 PM

after:
Screenshot 2023-06-15 at 3 08 29 PM

recipes/leftbar/general_row/general_row.vue
```diff
@@ -44,79 +44,82 @@
...
+        v-if="activeVoiceChat"
+        class="dt-leftbar-row__active-voice"
+      >
+        <dt-icon
+          size="300"
+          name="activity"
+        />
+      </div>
+      <dt-tooltip
+        v-else-if="!!unreadCount && hasUnreads"
```
iropolo: Could we create a computed property for this?

Sth like showUnreadCount

recipes/leftbar/general_row/general_row.vue
iropolo: I couldn't replicate

iropolo: It is the tooltip-container, seems like is offside


image

iropolo: What I can see 

`dt-leftbat-row--action-focused` is removed when you hover with the mouse.

I searched for it in the code and I found this handling:

```
@focus="actionFocused = true"
@blur="actionFocused = false"
@mouseleave="actionFocused = false"
```

I suspect it is an issue doing this way

juliodialpad: > actually just realized this was a problem before this change as well... might just be the storybook example?

This should be fixed now, removed `@mouseover` and `@mouseleave` events to prevent the mouse to remove the `dt-leftbar-row--action-focused`

juliodialpad: > It is the tooltip-container, seems like is offside
> 
> image

You're watching the tooltip container of the button behind it, should be fixed now.

juliodialpad: > What I can see
> 
> `dt-leftbat-row--action-focused` is removed when you hover with the mouse.
> 
> I searched for it in the code and I found this handling:
> 
> ```
> @focus="actionFocused = true"
> @blur="actionFocused = false"
> @mouseleave="actionFocused = false"
> ```
> 
> I suspect it is an issue doing this way

Thanks! this helped me to fix the issue faster

recipes/leftbar/style/leftbar_row.less
juliodialpad: Fixed

iropolo (approved):
Thanks for the update 🚀

braddialpad (approved):
Very nice thanks!

PR #1023, opened 2023-06-16
fix(message-input): image picker

sanjeevdialpad:
# Image picker in message input

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

changed Dt input to include file type. Used hidden prop to visually hide the input element and also include the click handler on the image button.

## :pencil: Checklist

- [ ] I have added tests

### DIFF
--- a/components/input/input.vue
+++ b/components/input/input.vue
@@ -1,8 +1,7 @@
 <template>
   <div
     ref="container"
-    class="base-input"
-    :class="{ 'd-vi-hidden': hidden }"
+    :class="['base-input', { 'd-vi-hidden': hidden }]"
     data-qa="dt-input"
   >
     <label
@@ -410,9 +409,14 @@ export default {
 
     inputListeners () {
       return {
-        input: event => {
-          this.$emit('input', event.target.value);
-          this.$emit('update:modelValue', event.target.value);
+        input: async event => {
+          let val = event.target.value;
+          if (this.type === INPUT_TYPES.FILE) {
+            const files = Array.from(event.target.files);
+            val = files.map(file => file.name);
+          }
+          this.$emit('input', val);
+          this.$emit('update:modelValue', val);
         },
 
         blur: event => {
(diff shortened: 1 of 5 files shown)


components/input/input.vue
```diff
@@ -2,6 +2,7 @@
   <div
     ref="container"
     class="base-input"
+    :class="{ 'd-vi-hidden': hidden }"
```
braddialpad: Generally not great practice to have multiple class attributes. Preferable to be like this:

```
:class="['base-input', { 'd-vi-hidden': hidden }]"
```

components/input/input.vue
```diff
@@ -556,6 +565,9 @@ export default {
 
   methods: {
     inputClasses () {
+      if (this.hidden) {
+        return [];
+      }
```
braddialpad: Why is this necessary, does it break something to set these classes when hidden? the dom elements would still exist right?

recipes/conversation_view/message_input/message_input.vue
```diff
@@ -32,14 +32,24 @@
...
               <dt-icon
                 name="image"
                 size="300"
               />
             </template>
+            <dt-input
+              ref="messageInputImageUpload"
+              type="file"
+              class="d-ps-absolute"
+              multiple="multiple"
```
braddialpad: If I select multiple files it only fires an event for the first one.

braddialpad (changes requested):
cool!

components/input/input.vue
sanjeevdialpad: yea actually i could remove both the return statements.

recipes/conversation_view/message_input/message_input.vue
sanjeevdialpad: yea i just noticed, have no clue why, let me check..

components/input/input.vue
```diff
@@ -570,6 +583,9 @@ export default {
     },
 
     inputWrapperClasses () {
+      if (this.hidden) {
```
sanjeevdialpad, line 586: this is needed. otherwise some padding is coming up.

sanjeevdialpad:
Added event to fire for drag and drop of media as well as fixed review comments. For now using the names of files in event payload (to be replaced later with binary)

braddialpad (approved):
Looks good thanks!

PR #1025, opened 2023-06-16
feat(datepicker): keyboard navigation

iropolo:
# Datepicker keyboard navigation

- [x] Feature

## :book: Description

https://dialpad.atlassian.net/browse/DIALTONE-1121

Keyboard navigation with arrow and tab keys for datepicker component.

It will switch between `month-year-picker` and `calendar` with `TAB`.
It will move around all calendar days with `arrow-keys`.
It will jump from `month-year-picker` to `calendar` with `DOWN` arrow key.
It will jump to `month-year-picker` when there is no more days at the bottom or top of the calendar.
It will change year or month with `ENTER` or `SPACE` (native event)
It will select day with `ENTER` or `SPACE` (native event)
It will start with focus on previous-year on mounted

Screen reader now will say 

In `month-year-picker`:

- Change to previous year, 2022
- Change to previous month, july
- Change to next month, september
- Change to next year, 2024

In `calendar`:
- Select day {date}


iropolo:
I wanted to create it as 'DRAFT' because there is work to be done in the screen-reader but I didn't change at time.

components/datepicker/modules/calendar.vue
```diff
@@ -20,16 +20,16 @@
...
+        :ref="el => { if (el) setDayRef(el, day.currentMonth) }"
         class="d-datepicker__day"
         :class="{
           'd-datepicker__day--disabled': !day.currentMonth,
-          'd-datepicker__day--selected': highlightDay ? ((day.text === highlightDay) && day.currentMonth) : day.selected,
+          'd-datepicker__day--selected': selectedDay ? ((day.text === selectedDay) && day.currentMonth) : day.selected,
         }"
         type="button"
-        :aria-label="day.text"
+        :aria-label="`${selectDayLabel} ${day.text}`"
```
juliodialpad, line 30: This should read the whole date, including month and year, or at least the month and the day

juliodialpad: This should read the whole date, including month and year, or at least the month and the day

components/datepicker/modules/calendar.vue
```diff
@@ -81,17 +88,79 @@ export default {
...
+
+        case 'ArrowDown':
+          event.preventDefault();
+          this.focusDay += 7;
+          try {
+            this.daysRef[this.focusDay].focus();
+          } catch (error) {
+            this.$emit('focus-month-year-picker');
+          }
+          break;
```
juliodialpad, line 130: I don't like this behavior of changing to `month-year-picker` on `arrow-up/down`, I'll prefer to change to next, previous month with `arrow-left/right`.

Here's an example https://material.angular.io/components/datepicker/overview

components/datepicker/modules/month-year-picker.vue
```diff
@@ -140,7 +156,64 @@ export default {
...
+          } else {
+            this.focusPicker++;
+            this.focusRefs[this.focusPicker].focus();
+          }
+          break;
+
+        case 'ArrowDown':
+          event.preventDefault();
+          this.$emit('focus-day');
+          break;
```
juliodialpad, line 205: Also this shouldn't change the focus to the calendar, It should only change with `tab`

juliodialpad (changes requested):
This is getting a good shape, just a few issues found on `tab` vs `arrow` navigation

components/datepicker/modules/calendar.vue
iropolo: Goddamnit I didnt push this update lol, I describe it in the PR description but didn't pushed.

Updated now.

components/datepicker/modules/calendar.vue
iropolo: It is changing next/previous month with arrow-left/right, arrow-up/down is just an addition.
We have it in the emoji picker, you can move trought the tabset and search with arrow-up/down.

I don't see the problem to have this additional navigation. I see it useful.

components/datepicker/modules/month-year-picker.vue
iropolo: same answer as the above, and also it is doing it with `tab`, maybe @braddialpad could bring us his opinion, it is not a problem to me to update this behaviour, but I see it useful as an addition, it is just sth extra.

components/datepicker/modules/calendar.vue
braddialpad: This is one of those opinionated accessibility things that could be argued for or against. I'm okay leaving it the way it is now unless we have customers / product developers complaining about it.

juliodialpad:
image

Let's remove the time and timezone information here, it just needs to read the year, month and day.
Maybe we can format that output into a more human readable, so screen reader can read it more naturally.

juliodialpad (changes requested):
Remove the time from aria-label please

iropolo:
> Remove the time from aria-label please

Good catch, updated!!

image

juliodialpad (approved):
image

Just remove the remaining time data from `selected date` and we're good to go. Thanks!

iropolo:
> image
> 
> Just remove the remaining time data from `selected date` and we're good to go. Thanks!

I thought this but since we will return the default JS Date object I would like to leave this as it is and then they can handle whatever they want

braddialpad:
> I thought this but since we will return the default JS Date object I would like to leave this as it is and then they can handle whatever they want

Yeah if that's how the standard JS date is we should just leave it as is.

braddialpad (approved):

juliodialpad:
> > image
> > Just remove the remaining time data from `selected date` and we're good to go. Thanks!
> 
> I thought this but since we will return the default JS Date object I would like to leave this as it is and then they can handle whatever they want

Ah right, I missed that, it's all good here now, thanks! good work.

PR #1029, opened 2023-06-20
fix(rich-text-editor): add a check and workaround for phone regex

MikaDialpad:
# UC-18072: add a check and workaround for phone regex

So this regex I added fails on some older browser versions in a spectacular way that renders the whole app unusable. I didn't realize that the support for [lookbehind is so bad](https://caniuse.com/js-regexp-lookbehind) in JavaScript, so we need some workaround for it. I couldn't find a completely working alternative, but just omitting that part works for all except one edge case (+1234567890123456 will match the last 15 characters when it should discard the whole match), so I _think_ this is good enough.

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

### DIFF
--- a/common/utils.js
+++ b/common/utils.js
@@ -262,8 +262,19 @@ const emailAddressRegex = new RegExp(
  * @returns {RegExp}
  */
 export function getPhoneNumberRegex (minLength = 7, maxLength = 15) {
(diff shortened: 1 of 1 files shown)


braddialpad (approved):
Thanks, Safari just started supporting this so shouldn't really be a problem in the future.

PR #1041, opened 2023-06-26
fix(message-input): error notice message

sanjeevdialpad:
# fix(message-input): error notice message

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

noticeKind and noticeMessage props for message input to display the error banner. Using dtNotice for the error notice.

## :pencil: Checklist

- [ ] I have added tests


sanjeevdialpad:
pls ignore, creating a deployed version for testing.

sanjeevdialpad:
@braddialpad ready for review. Waiting on design to approve.

recipes/conversation_view/message_input/message_input.vue
```diff
@@ -1,11 +1,26 @@
 <template>
+  <dt-notice
+    v-if="errorNoticeOpen"
```
braddialpad: I'm not entirely sure what the real world use cases are for this notice, but I'm not sure "setting the string" to show the message is the best pattern.

1. It is not obvious it works like this
2. What if you show a message, the user closes it, and then you need to show the same message again later? If you update the prop to the same message it already is, it will not trigger any change and errorNoticeOpen will not get set and the message will not display. You'd have to do something really awkward like manually set the string to nothing do a nextTick and then set the string to the message again.

Probably better to have a noticeShow prop. I know it's yet another prop and you were probably trying to limit how many there were, but overall I think this will be easier to understand.

braddialpad (changes requested):

recipes/conversation_view/message_input/message_input.vue
sanjeevdialpad: but if i use a prop noticeShow and then assign it to a data value, when the user manually closes it using the x, changing the prop would not help again right? Should i let it be on the product to listen to `notice-close` event and change prop of showNotice to false?

sanjeevdialpad: but i agree we should another prop to show the notice. Have made the changes.

sanjeevdialpad: i added a watcher for showNotice to solve the problem you mentioned.

braddialpad: > but if i use a prop noticeShow and then assign it to a data value, when the user manually closes it using the x, changing the prop would not help again right? Should i let it be on the product to listen to notice-close event and change prop of showNotice to false?

You're right, this is where we need to use [sync (vue2)](https://v2.vuejs.org/v2/guide/components-custom-events.html#sync-Modifier) and [v-model:property (vue3)](https://vuejs.org/guide/components/v-model.html#component-v-model) in this case the prop would change to false when the notice was closed by the user.

sanjeevdialpad: done removed watcher and used v-model instead.  

Unrelated: Learned something new more importantly, something that bugged me about previous implementation of message input which was leading to a bug lol.

components/rich_text_editor/rich_text_editor.vue
```diff
@@ -293,6 +293,10 @@ export default {
     updateEditorAttributes (attributes) {
       this.editor.setOptions({ editorProps: { attributes } });
     },
+
+    focusEditor () {
```
sanjeevdialpad, line 297: @braddialpad Fixed it bit differently, tiptap had a command to focus to use, so i used ref to invoke this method directly.

The previous one had issue where the cursor was jumping from start to focus position on click.

braddialpad: oh nice yeah that sounds like a better way to do it.

braddialpad (approved):

braddialpad:
Hide these and display the actual output event name instead (commonly done in our other components)

Screenshot 2023-06-29 at 11 47 05 AM


ex stories.js
```
onClick: {
  table: {
    disable: true,
  },
},

click: {
  description: 'Native chip click event',
  table: {
    type: { summary: 'event' },
  },
},
```

PR #1042, opened 2023-06-27
feat(datepicker): datepicker vue3 component

iropolo:
# Datepicker vue 3 component

## :hammer_and_wrench: Type Of Change

- [x] Feature

## :book: Description

Datepicker vue 3 component migration.

https://dialpad.atlassian.net/browse/DLT-1127

PR #1043, opened 2023-06-27
feat(datepicker): added missing events

iropolo:
# feat(datepicker): added missing events

- [x] Feature

## :book: Description

Added missing events.
It works without them thats why I didn't catch it previously, but there is a warning because of this in the version 3 of the component, thats how I catch it.


braddialpad (approved):

iropolo:
Added missing locale prop requested in [Datepicker vue3]( https://github.com/dialpad/dialtone-vue/pull/1042#discussion_r1244194721)

PR #1055, opened 2023-06-30
fix(message-input): input value mutation error on usage of DtMessageInput fix

sanjeevdialpad:
# fix(message-input): input value mutation error on usage of DtMessageInput fix

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

When we try to clear input value through two way binding, it results in error on some versions of vue3 and hence generally a bad practice. Instead we expose the editor setContent from richtextEditor and only clear the value if needed. This way we can avoid this.

Initial value for messageInput can also be set using this setContent method exposed by MessageInput recipe.

## :pencil: Checklist

- [ ] I have added tests


sanjeevdialpad:
@braddialpad @MikaDialpad this was the bug on messageInput which fred helped to fix. But might need your recommendation if theres any better way/ your thoughts/opinion on the fix.

iropolo:
> @braddialpad @MikaDialpad this was the bug on messageInput which fred helped to fix. But might need your recommendation if theres any better way/ your thoughts/opinion on the fix.

what was the bug issue to replicate it?

sanjeevdialpad:
When the recipe is used in a vue 3 app, on submit results in error when setting input to null. This isn't replicable on the storybook currently. An example would be to use this on appcore.

MikaDialpad:
> When the recipe is used in a vue 3 app, on submit results in error when setting input to null. This isn't replicable on the storybook currently. An example would be to use this on appcore.

Sorry I wasn't available to debug this last week, but I'm curious what the error message was? Did we try to set `null` value on the input? Or do we call the method at the "wrong" time? I'm fine with exposing a method that can clear the contents, but this takes away the two way value binding that we might want to keep.

sanjeevdialpad:
i agree we might have to give up on the two way bindin this way but the only utility for product to use messageInput is to clear it on send or retain the text on server error. So i think this should be ok.

Am trying to reproduce the error which consitently came up for more than 2 weeks and magically it isnt coming up now. I wonder if its due to dep cache of dialtone vue in app-core. I will paste the error once i reproduce it again.

iropolo:
TBH `clearnOnSend` should be a default behaviour of the recipe, why you will want to have it in `false`?

I will directly do the `this.setContent('');` in `onSend` method... or even better, just clear the input there if you are not going to use `setContent` method anywhere.

I cannot see it necessary to expose a prop, or it is...?

As we cannot replicate the issue but for you this way is right to handle it, I will be good with this then..

sanjeevdialpad:
yea clearOnSend should probably true as you said. but we do have a case where we might not want to clear it as mentioned due to server error. Are you sggesting we just set it to true on default and not change anything else? 

Btw am still not able to reproduce it but the error was basically thrown on Message_input recipe when i set input.value = ''.
Below is the error
Screenshot 2023-07-04 at 12 00 49 PM

MikaDialpad:
> yea clearOnSend should probably true as you said. but we do have a case where we might not want to clear it as mentioned due to server error. Are you sggesting we just set it to true on default and not change anything else?
> 
> Btw am still not able to reproduce it but the error was basically thrown on Message_input recipe when i set input.value = ''. Below is the error Screenshot 2023-07-04 at 12 00 49 PM

Hmm, this looks like it might again be thrown by the Link extension. Let me know if you're able to create a repro environment and I can pull the code and take a look.

iropolo:
> yea clearOnSend should probably true as you said. but we do have a case where we might not want to clear it as mentioned due to server error. Are you sggesting we just set it to true on default and not change anything else?
> 
> Btw am still not able to reproduce it but the error was basically thrown on Message_input recipe when i set input.value = ''. Below is the error Screenshot 2023-07-04 at 12 00 49 PM

hmmm... ye I mean it is strange to do this to hotfix another bug... so.. the final result is... strange...

If this 'clearOnSend' is the regular behaviour of the recipe Im okay with it.
And also seems like it fixs the issue.

But if there is the case where it is not necessary and we should create a prop to handle it, we are moving forward to a non-correct solution, because IMO thats not the way to do it.

What Im thinking is 
1. you are binding the input value in the product side.
2. when you sent the message, the product side should clear the input value, because it is in charge of the correct behaviour of the input component.

And thats it... if it throw an error we should fix the error first, I dont like this way of bypassing it.

If the server threw an error, product side should handle it (in this case do not clear the input).

Also there is a case where someone use the component and face the same issue, we don't know if 'this solution' helps because this is attached on the product side behaviour.

Before move on I would like to deep in the error and do a research there on whats going on....

braddialpad:
> What Im thinking is
> 
> 1. you are binding the input value in the product side.
> 2. when you sent the message, the product side should clear the input value, because it is in charge of the correct behaviour of the input component.
> 
> And thats it... if it throw an error we should fix the error first, I dont like this way of bypassing it.
> 
> If the server threw an error, product side should handle it (in this case do not clear the input).

I would agree with all these statements. We need to find the root cause of the error.

sanjeevdialpad:
I tried the whole day to reproduce that exact error and i couldnt, havent updated vue version too. but am sure it was happening as both @ryandialpad and @superfreddialpad spent some time debugging this exact thing.

For now i will take the feedback that product should clear the input value may it be with two way binding or by exposing a method, so i will close this PR and reopen if that bug surfaces again to continue the discussion.

Have used this recipe on app-core here: https://github.com/dialpad/web-clients/pull/518

braddialpad:
> I tried the whole day to reproduce that exact error and i couldnt, havent updated vue version too. but am sure it was happening as both @ryandialpad and @superfreddialpad spent some time debugging this exact thing.
> 
> For now i will take the feedback that product should clear the input value may it be with two way binding or by exposing a method, so i will close this PR and reopen if that bug surfaces again to continue the discussion.
> 
> Have used this recipe on app-core here: [dialpad/web-clients#518](https://github.com/dialpad/web-clients/pull/518)

Hmm one of those weird things I guess.. Yeah close for now and let us know if it becomes a problem again.

sanjeevdialpad:
@braddialpad @MikaDialpad @iropolo opened new PR: https://github.com/dialpad/dialtone-vue/pull/1058

PR #1058, opened 2023-07-05
fix(message-input): two way binding for message input value

sanjeevdialpad:
# fix(message-input): Two way binding for message input value

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

- Two way binding using modelValue on message-input
- Remove redundant watch on rich text editor.
- message input story changes

## :pencil: Checklist

- [ ] I have added tests

### DIFF
--- a/components/rich_text_editor/rich_text_editor_default.story.vue
+++ b/components/rich_text_editor/rich_text_editor_default.story.vue
@@ -1,6 +1,6 @@
 <template>
   <dt-rich-text-editor
-    v-model:value="inputValue"
+    v-model="inputValue"
     :editable="editable"
     :input-aria-label="inputAriaLabel"
     :input-class="inputClass"
@@ -22,18 +22,6 @@ export default {
 
   components: {
     DtRichTextEditor,
-  },
-
-  data () {
-    return {
-      inputValue: this.modelValue,
-    };
-  },
-
-  watch: {
-    value (val) {
-      this.inputValue = val;
-    },
   },
 };
 </script>
--- a/recipes/conversation_view/message_input/message_input.vue
+++ b/recipes/conversation_view/message_input/message_input.vue
@@ -32,7 +32,7 @@
       <div class="d-of-auto d-mx16 d-mt8 d-mb4 d-hmx40p">
         <dt-rich-text-editor
           ref="richTextEditor"
-          v-model="inputValue"
+          v-model="internalInputValue"
           :editable="editable"
           :input-aria-label="inputAriaLabel"
           :input-class="inputClass"
@@ -504,7 +504,7 @@ export default {
   data () {
     return {
       skinTone: 'Default',
-      inputValue: this.modelValue,
+      internalInputValue: this.modelValue, // internal input content
       hasFocus: false,
       imagePickerFocus: false,
       emojiPickerFocus: false,
@@ -515,7 +515,7 @@ export default {
 
   computed: {
     inputLength () {
-      return this.inputValue.length;
+      return this.internalInputValue.length;
     },
 
     displayCharacterLimitWarning () {
@@ -558,7 +558,7 @@ export default {
 
   watch: {
     modelValue (newValue) {
-      this.inputValue = newValue;
+      this.internalInputValue = newValue;
     },
   },
 
@@ -584,7 +584,7 @@ export default {
         return;
       }
 
-      this.inputValue = this.inputValue + emoji.shortname;
+      this.internalInputValue = this.internalInputValue + emoji.shortname;
       this.emojiPickerOpened = false;
     },
 
@@ -604,7 +604,7 @@ export default {
       if (this.isSendDisabled) {
         return;
       }
-      this.$emit('submit', this.inputValue);
+      this.$emit('submit', this.internalInputValue);
     },
 
     noticeClose () {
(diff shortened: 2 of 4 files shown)


components/rich_text_editor/rich_text_editor_default.story.vue
```diff
@@ -1,6 +1,6 @@
 <template>
   <dt-rich-text-editor
-    v-model:value="inputValue"
```
sanjeevdialpad, line 3: wasnt required to have value attribute as it uses modelValue on prop.

components/rich_text_editor/rich_text_editor_default.story.vue
```diff
@@ -23,17 +23,5 @@ export default {
   components: {
     DtRichTextEditor,
   },
-
-  data () {
```
sanjeevdialpad, line 27: two way binding was working without these

recipes/conversation_view/message_input/message_input.vue
```diff
@@ -554,6 +555,12 @@ export default {
     },
   },
 
+  watch: {
+    modelValue (newValue) {
```
sanjeevdialpad, line 560: am not sure if we should be checking current value and compare to newValue here.

braddialpad: No don't think this is necessary. Vue itself is smart enough to know if the value was changed or not.

recipes/conversation_view/message_input/message_input.vue
braddialpad, line 1: Just happened to notice this..

```
.d-notice__icon {
  margin-right: 8px;
}
```

Please improve the scoping on this. Currently this will change every single `d-notice__icon` across dialtone and any consuming applications.

recipes/conversation_view/message_input/message_input.vue
```diff
@@ -554,6 +555,12 @@ export default {
...
+      this.inputValue = newValue;
```
braddialpad: When we change `this.inputValue` internally such as in onSelectEmoji, does that need to fire an `update:modelValue` event to keep modelValue in sync or is this done some other way?

sanjeevdialpad: since we use v-model on richTextEditor: (https://github.com/dialpad/dialtone-vue/pull/1058/files#diff-62664268bb3876ed01b9494168e2b23396f4748e4a4f36c1ed22a5ca075b57b1R5), changing inputValue should do both i guess, firing the event and also changing the prop value. it works without sending that event. Or are you suggesting to propagate the value upwards? That anyway happens with input event on input change.

recipes/conversation_view/message_input/message_input.vue
sanjeevdialpad: ah damn, thanks, changed it.

components/rich_text_editor/rich_text_editor_default.story.vue
braddialpad, line 1: Hmm wait a sec.. don't all these need to have $attrs since this is vue 3? I don't understand how this could even work

recipes/conversation_view/message_input/message_input.vue
braddialpad: This is sort of off the top of my head so let me know if I have not recognized some issue with this, but could we get rid of this internal inputValue completely?

here's what I'm thinking:

1. pass modelValue directly into dt-rich-text-editor
2. remove this watch
3. in `onSelectEmoji` instead of setting `this.inputValue = this.inputValue + emoji.shortname;` do `emit('update:modelValue', this.modelValue + emoji.shortname)`

Would this work or am I crazy?

components/rich_text_editor/rich_text_editor_default.story.vue
sanjeevdialpad: am having the same doubt, without attrs, message input stuff didnt work. but if you look at ivr_node, and feed_items, attrs isnt used and still works.

recipes/conversation_view/message_input/message_input.vue
sanjeevdialpad: @superfreddialpad tagging fred here since he wanted to think about nuking the entire prop and have a ref exposed to set content. 
Meanwhile i will try what you mentioned brad!

components/rich_text_editor/rich_text_editor_default.story.vue
braddialpad: Hmm. I wonder if they changed this in a storybook update

recipes/conversation_view/message_input/message_input.vue
sanjeevdialpad: @braddialpad vue doesnt allow it with unexpected mutation of prop error. In rich text editor its not directly used as a prop so its fine. but in this component, we need to send it to rich text editor which poses a problem.

braddialpad: Hmm where are we still mutating the prop after making this change?

superfreddialpad: I think the watch should be fine? @sanjeevdialpad lets meet tomorrow and figure this out. Hopefully we can just make it work with `v-model` (modelValue)

braddialpad (approved):

PR #1060, opened 2023-07-06
feat(datepicker): date format utils

iropolo:
# Date format utils

- [x] Feature

## :book: Description

We expose date utils format following https://dialpad.design/guides/writing-guidelines/#dates

https://dialpad.atlassian.net/browse/DLT-721


iropolo:
Hmm... why in the preview the styles are not working correctly?
but in https://vue.dialpad.design/?path=/story/components-datepicker--default it does....

Local works properly

image

braddialpad:
> Hmm... why in the preview the styles are not working correctly?

Hmm.. you are certainly getting some weird issues lately.. Maybe it doesn't like the scoped styles? Try unscoped styles

iropolo:
> > Hmm... why in the preview the styles are not working correctly?
> 
> Hmm.. you are certainly getting some weird issues lately.. Maybe it doesn't like the scoped styles? Try unscoped styles

The styles are just for the table and they are in the default template, I dont think it is related or maybe you are talking about something else.

I would like to merge this and I will deploy it. Will see if this is sth internal of the preview or.. sth else...

... But ye lemme check it... strange solutions for strange issues, lets see if it is works

braddialpad:
> The styles are just for the table and they are in the default template, I dont think it is related or maybe you are talking about something else.

Yeah I know but sometimes weird stuff happens so it might be worth trying. I only mentioned because we don't used scoped CSS often so it's something that is different from normal.

iropolo:
> > The styles are just for the table and they are in the default template, I dont think it is related or maybe you are talking about something else.
> 
> Yeah I know but sometimes weird stuff happens so it might be worth trying. I only mentioned because we don't used scoped CSS often so it's something that is different from normal.

image

Could be this related to the issue? seems like it is not installing dialtone vue dependencies idk..

braddialpad:
> Could be this related to the issue? seems like it is not installing dialtone vue dependencies idk..

It doesn't need to install them when it makes a cache hit

braddialpad:
We ever figure out what is going on here?

iropolo:
> We ever figure out what is going on here?

Same as #1061 

1. Dialtone wasn't updated with emoji-picker and datepicker styles (now is updated).
2. Emojipicker and datepicker are using `dt-` (dialtone 8 styles) they will not work.

Local datepicker in staging branch

image

Local emojipicker in staging-vue3 branch

image

braddialpad:
> Same as #1061
> 
> 1. Dialtone wasn't updated with emoji-picker and datepicker styles (now is updated).
> 2. Emojipicker and datepicker are using `dt-` (dialtone 8 styles) they will not work.
> 
> Local datepicker in staging branch
> 
> image
> Local emojipicker in staging-vue3 branch
> 
> image

I think we'll need to change the variables to work on the dialtone 7 branch. The changes to dt- variables should only be on the "dialtone-8" branch

braddialpad (approved):

iropolo:
Styles migrations will be tackled in another PR. Because this one is not related to that.

PR #1068, opened 2023-07-11
feat(rich-text-editor): add hardbreak and codeblocks

sanjeevdialpad:
# feat(rich-text-editor): add hardbreak and codeblocks

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Add hardbreak to enable linebreaks with keyboard shortcuts.
Add Enter on send on message input.
Enable ability to add codeblocks by typing triple ` or ~ with the codeblocks extension.


iropolo:
Is there a way to check the codeblock in the preview ? cos Im ussing ` or ~ but I cannot replicate the expected behaviour of the codeblock, what Im missing?

I can see the `input` event in the `shift+enter` witouth `submit` and thats correct cos the `.exact` event modifier, but also Im seeing the `input` event in the `enter` keydown, is that an expected behaviour? or should be only `submit`.?

sanjeevdialpad:
You have to add a space to trigger the codeblock after the triple ~. But it's a bit wonky. If you started a paragraph text and type ~~~ it does not start a code block. Idk why but it works sometimes. 

You are right. The input on enter shouldn't happen. It's a good catch. Let me check.

iropolo:
Okay cool I can play with the codeblock, yes it works weirdly.

But what I can see the input value is the same as if there is no codeblock inside. How you will handle it later?

iropolo:
We should add this hardbreak and codeblock functionality in the rich-text-editor documentation.

sanjeevdialpad:
> Okay cool I can play with the codeblock, yes it works weirdly.
> 
> But what I can see the input value is the same as if there is no codeblock inside. How you will handle it later?

If you change the output format to json or html, you will get the required format tags etc.

sanjeevdialpad:
> We should add this hardbreak and codeblock functionality in the rich-text-editor documentation.

We auto add/enable these extensions but maybe you are right, we can make these optional ? But am not certain if we wanna do it that way.
But i can mention in short what each functinality uses as extension so the developer can refer it when they use it.

iropolo:
> > Okay cool I can play with the codeblock, yes it works weirdly.
> > But what I can see the input value is the same as if there is no codeblock inside. How you will handle it later?
> 
> If you change the output format to json or html, you will get the required format tags etc.

Okay cool, yes it worked with html and json

image

image



We should add this also in the documentation to explain how to get the expected input.

The expected return on HTML format will be:
`<pre><code> </code></pre>`

The expected return on JSON format will be 
 type: 'codeblock'  on content array

There is no codeblock exposed in TEXT format

iropolo:
> > We should add this hardbreak and codeblock functionality in the rich-text-editor documentation.
> 
> We auto add/enable these extensions but maybe you are right, we can make these optional ? But am not certain if we wanna do it that way. But i can mention in short what each functinality uses as extension so the developer can refer it when they use it.

hmm ye, we should have this shortcuts or behaviour wrote somewhere, you can explain this in the component documentation because those are additions to the component behaviour so it will be okay.

iropolo:
> > > Okay cool I can play with the codeblock, yes it works weirdly.
> > > But what I can see the input value is the same as if there is no codeblock inside. How you will handle it later?
> > 
> > 
> > If you change the output format to json or html, you will get the required format tags etc.
> 
> Okay cool, yes it worked with html, but no with json
> 
> image image
> We should add this also in the documentation to explain how to get the expected input.

Oh sorry I can see the `type` codeblock now, we can detail this in the docuemntation 👍🏼

sanjeevdialpad:
@superfreddialpad I verified that the CMD+Enter only exits out of a codeblock but doesnt add a newline generally. is that ok or should i override that too?

sanjeevdialpad:
> > > Okay cool I can play with the codeblock, yes it works weirdly.
> > > But what I can see the input value is the same as if there is no codeblock inside. How you will handle it later?
> > 
> > 
> > If you change the output format to json or html, you will get the required format tags etc.
> 
> Okay cool, yes it worked with html and json
> 
> image image
> We should add this also in the documentation to explain how to get the expected input.
> 
> The expected return on HTML format will be: `\<pre\><code> </code>\</pre\>`
> 
> The expected return on JSON format will be type: 'codeblock' on content array
> 
> There is no codeblock exposed in TEXT format

i think we should mention/make this change in rich text editor documentation. and then refer that here in message input.

sanjeevdialpad:
> I can see the input event in the shift+enter witouth submit and thats correct cos the .exact event modifier, but also Im seeing the input event in the enter keydown, is that an expected behaviour? or should be only submit.?

@iropolo I fixed this. ty for the catch.

sanjeevdialpad:
Screenshot 2023-07-13 at 2 21 29 PM

docs updated as well. will add some examples in there.

iropolo (approved):
Looks better now, thanks!

PR #1069, opened 2023-07-13
fix(general-row): overridden display property

juliodialpad:
# General Row: Fix overridden display property

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Added `scoped` to style tag, to increase specificity, this should fix any override coming from dialtone-styles in the future.

## :bulb: Context

Were having issues on product because display property is being overridden probably due to the order CSS files imported by WebPack.

## :pencil: Checklist

- [ ] I have added tests

## :crystal_ball: Next Steps

- Cherry-picking this commit into dialtone8 branch
- Publish a new `dialtone-vue@next` version
- Update dialtone-vue on product and cherry-pick the commit.

https://vue-loader.vuejs.org/guide/scoped-css.html#child-component-root-elements

### DIFF
--- a/recipes/leftbar/style/leftbar_row.less
+++ b/recipes/leftbar/style/leftbar_row.less
@@ -219,7 +219,7 @@
   }
 
   &__unread-badge {
-    display: var(--leftbar-row-omega-display);
+    display: var(--leftbar-row-omega-display) !important;
   }
 
   &__active-voice {
--- a/recipes/leftbar/style/leftbar_row.less
+++ b/recipes/leftbar/style/leftbar_row.less
@@ -14,7 +14,6 @@
   --leftbar-row-alpha-color-foreground: var(--theme-sidebar-icon-color);
   --leftbar-row-alpha-width: calc(var(--size-300) * 10);
   --leftbar-row-alpha-height: calc(var(--size-300) * 9);
-  --leftbar-row-omega-min-width: 0;
   --leftbar-row-omega-height: var(--leftbar-row-alpha-height);
   --leftbar-row-omega-display: inline-flex;
   --leftbar-row-description-font-weight: var(--fw-normal);
@@ -37,7 +36,7 @@
   background-color: var(--theme-sidebar-row-color-background);
   border-radius: var(--leftbar-row-radius);
   transition-duration: var(--td200);
-  transition-property: background-color,border,box-shadow;
+  transition-property: background-color, border, box-shadow;
   transition-timing-function: var(--ttf-out-quint);
   cursor: pointer;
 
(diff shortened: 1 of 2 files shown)


recipes/leftbar/style/leftbar_row.less
```diff
@@ -219,7 +219,7 @@
   }
 
   &__unread-badge {
-    display: var(--leftbar-row-omega-display);
+    display: var(--leftbar-row-omega-display) !important;
```
braddialpad: What exactly is overriding this that we have to add important?

juliodialpad: image

image

image

Styles from `.d-badge` coming from `dialtone.css` and `badge.less`.

Does `display: inline-flex` have priority over `display: var(--some-var)`?

juliodialpad: https://github.com/dialpad/dialtone-vue/assets/87546543/ceb944a8-d025-4da8-a334-b03ad61fb9c3

It doesn't make a lot of sense, seems that `.dt-leftbar-row__unread-badge` is actually being called after those other classes and they haven't more specificity, so the last display property should be applied, right?

braddialpad: Hmm I wonder if it's the order it's imported since one is coming from dialtone and the other from dialtone vue. Maybe just hack more specificity into it rather than use important. Something like.  .d-badge.dt-leftbar-row...

iropolo: Since is CSS Im pretty sure the import order will affect this.
But why we have dialtone-vue styles in this?, This should be in dialtone and imported from there like all the components, Im missing sth here...

Im agree on try to find a proper fix for this, if we want to go with this hotfix we will probably miss a bunch of other issues we didn't find yet.

If we want to go with this kind of hotfixes we should document this approach somewhere...

braddialpad: > But why we have dialtone-vue styles in this?

We don't put recipe components in dialtone, only core components

iropolo: hmm, okay so... recipe styles are overridden by core component styles.

Then we can confirm using !import in specific styles will not fix this issue, because this is a 'how we work styles' issue.

So Im guessing MAYBE if we import dialtone-vue styles after dialtone styles these would be overwritten.
(It depends if the css access is more specific than the other)

But could this bring us more issues?

I'm not convinced with this anyway...

iropolo: Ok I tested it locally in vue-3 branch and yes, the library override the component style

image

braddialpad: What did you change here? the import order of the files?

iropolo: I deleted the comment cos after realising what I did it was dumb, I just used the content-class prop of the component to fix it, but that is how it supposed to work.

Im doing test now with the current bug

iropolo:
Now we know (by tests) , `scoped` styles will fix this issue.

Should we create the ticket to migrate recipes styles to `scoped` approach and verify the enforcement?

iropolo (approved):

juliodialpad:
> Now we know (by tests) , `scoped` styles will fix this issue.
> 
> Should we create the ticket to migrate recipes styles to `scoped` approach and verify the enforcement?

I agree, please create a ticket to migrate our recipe components to scoped styling

PR #1077, opened 2023-07-18
fix: add customized css to badge label and contact info (vue3)

tonydialpad:
## :hammer_and_wrench: Type Of Change

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

This pull request adds a property to customize the CSS for badge label and adds fix css for Contact Info conponent.

## :pencil: Checklist

- [x] I have reviewed my changes
- [ ] I have added tests
- [ ] I have added all relevant documentation
- [x] I have validated components with a screen reader
- [x] I have validated components keyboard navigation
- [x] I have considered the performance impact of my change
- [x] I have checked that my change did not significantly increase bundle size
- [ ] I am exporting any new components or constants in the index.js in the component directory
- [ ] I am exporting any new components or constants in the index.js in the root

## :crystal_ball: Next Steps

Make a new release for Dialtone-Vue

## :camera: Screenshots / GIFs
Screenshot 2023-07-17 at 6 47 17 PM

## :link: Sources

JIRA: https://dialpad.atlassian.net/browse/DP-74536
vue2 PR: https://github.com/dialpad/dialtone-vue/pull/1076

PR #1078, opened 2023-07-19
fix: `contact_info` style override issue

tonydialpad:

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Per comment: https://github.com/dialpad/dialtone-vue/pull/1077#discussion_r1267416074
~With `scoped` tag, the `min-width` override does not work. It looks like after `scoped` tag is added, `<style>` can only override contact_info (and if `vue` find this is not `contact_info` component element, it will just ignore this override).~ We should use :deep() to access to
```
.dt-item-layout--content
.dt-item-layout--left
.dt-item-layout--right
```
Because these classes are on `item_layout.vue` component, that is a different component.

## :pencil: Checklist

- [ ] I have added tests

## :crystal_ball: Next Steps

Make a new release for Dialtone-Vue

## :link: Sources

Original pull request: https://github.com/dialpad/dialtone-vue/pull/1077#discussion_r1267416074
Vue3 pull request: https://github.com/dialpad/dialtone-vue/pull/1079

### DIFF
--- a/recipes/list_items/contact_info/contact_info.vue
+++ b/recipes/list_items/contact_info/contact_info.vue
@@ -221,21 +221,21 @@ export default {
 };
 </script>
 
-<style>
-.dt-contact-info .dt-item-layout--content {
+<style scoped>
+.dt-contact-info :deep(.dt-item-layout--content) {
   /*
   DP-74536: Add `min-width` to make the width of "contact info" adjustable.
   */
   min-width: var(--space-825);
 }
-.dt-contact-info .dt-item-layout--left {
+.dt-contact-info :deep(.dt-item-layout--left) {
   /*
   DP-74536: To make 'Avatar' in fixed position when resizing the window.
   */
   min-width: var(--space-650);
   justify-content: flex-start;
 }
-.dt-contact-info .dt-item-layout--right {
+.dt-contact-info :deep(.dt-item-layout--right) {
   /*
   DP-74536: Remove `min-width` which cause extra unused empty space on the right of "contact info".
   */
--- a/recipes/list_items/contact_info/contact_info.vue
+++ b/recipes/list_items/contact_info/contact_info.vue
@@ -221,7 +221,7 @@ export default {
 };
 </script>
 
-<style scoped>
+<style>
 .dt-contact-info .dt-item-layout--content {
   /*
   DP-74536: Add `min-width` to make the width of "contact info" adjustable.


iropolo:
Using `:deep()` pseudo-class fixes the issue.

`.dt-contact-info :deep(.dt-item-layout--content) {`
`.dt-contact-info :deep(.dt-item-layout--left) {`
`.dt-contact-info :deep(.dt-item-layout--right) {`

But my interesting concern is why we should access with `:deep` with `scoped` attribute on style tag

iropolo:
> Using `:deep()` pseudo-class fixes the issue.
> 
> `.dt-contact-info :deep(.dt-item-layout--content) {` `.dt-contact-info :deep(.dt-item-layout--left) {` `.dt-contact-info :deep(.dt-item-layout--right) {`
> 
> But my interesting concern is why we should access with `:deep` with `scoped` attribute on style tag

If you structure like this:

```
<div class="dt-contact-info">
    <dt-list-item
```

The styles will work as expected with `scoped` and without any `:deep`... so its probably because some off-sync load component

iropolo:
> Using `:deep()` pseudo-class fixes the issue.
> 
> `.dt-contact-info :deep(.dt-item-layout--content) {` `.dt-contact-info :deep(.dt-item-layout--left) {` `.dt-contact-info :deep(.dt-item-layout--right) {`
> 
> But my interesting concern is why we should access with `:deep` with `scoped` attribute on style tag

Huh... I didn't check the entire file... so.. yes you MUST use `:deep()` to access to

```
.dt-item-layout--content
.dt-item-layout--left
.dt-item-layout--right
```

Because these classes are on `item_layout.vue` component, that is a different component.

When you add `scoped` you are generating a unque id to this component to add the style.
This id is just for `contact_info.vue` component, not for `item_layout.vue` component so it won't work in this last one.

Without `scoped` will work because its a global style, and it will access to child component due CSS specificity.

So.. yes, you need to write the styles using `:deep()` as I put before.

iropolo (changes requested):
Add `scoped` attribute to style tag

Update classes using `:deep()` pseudo-class.

```
.dt-contact-info :deep(.dt-item-layout--content) {
.dt-contact-info :deep(.dt-item-layout--left) {
.dt-contact-info :deep(.dt-item-layout--right) {
```

iropolo (approved):
Thanks! 🚀

PR #1080, opened 2023-07-19
feat(message-input): emoji extension

sanjeevdialpad:
# feat(message-input): emoji extension

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

tiptap allows custom nodes which render a specific vue component using [node views](https://tiptap.dev/guide/node-views) . This hooks up with the custom node extension to give the emoji extension.

We need to define two sets of rules for input and paste.
1. Shortcode (:code:)
2. unicode (usually pasted but windows apparently has unicode typing support too).

The json has the node defined in it as type emoji and the shortcode as the text. Please provide further feedback if more attributes are needed. (parsed skin tones etc? )

The documentation does not give a broad view hence you can take a look at already present extensions for more info like this:  [code block](https://github.com/ueberdosis/tiptap/blob/develop/packages/extension-code-block/src/code-block.ts)

## :pencil: Checklist

- [ ] I have added tests


components/rich_text_editor/extensions/emoji/emoji.test.js
```diff
@@ -0,0 +1,94 @@
...
+
+// Helpers
+const _setChildWrappers = () => {
+  editorEl = document.getElementsByClassName('qa-editor')[0];
+};
+
+const _setValue = async (value) => {
+  // await wrapper.setProps({ modelValue: value });
+  editorEl.innerHTML = value;
+  await wrapper.vm.$nextTick();
```
iropolo, line 32: Here I still have doubt on the way to do it.

I don't know tiptap but in the documentation it says to uset [set-content](https://tiptap.dev/api/commands/set-content). 

When I tried to use it, It didn't catch even :cat:, thats a suspicious for me to know if we are doing it in the correct way in the first place.

components/rich_text_editor/extensions/emoji/emoji.test.js
```diff
@@ -0,0 +1,94 @@
...
+
+// Wrappers
+let wrapper;
+let editorEl;
+
+// Test Environment
+let propsData;
+let attrs;
+let slots;
+let listeners;
```
iropolo: If we are not using it, pleace remove this implementation.

iropolo:
Could you explain to us a little bit more deep whats the idea of this implementation and the usage of it please?.

Maybe create a different story to see this in action if is necessary.

components/rich_text_editor/extensions/emoji/emoji.test.js
sanjeevdialpad: So the weird thing is, the actual implementation uses setContent whenever v-model changes and that does trigger inputrules and mark the node. Am not sure why doing this in test environment is not working as expected.

sanjeevdialpad:
@iropolo different story as in JIRA story?

The idea of the implementation is to identify emoji as a separate node in the editor and also to render emoji text when shortcode is used. Parallely, when we paste in the unicode, it should also identify it as a emoji. right now it just puts it as a text node. thats everything this extension does for now. but we can improve it later further by including dropdown for emoji selection as user types etc

iropolo:
> @iropolo different story as in JIRA story?
> 
> The idea of the implementation is to identify emoji as a separate node in the editor and also to render emoji text when shortcode is used. Parallely, when we paste in the unicode, it should also identify it as a emoji. right now it just puts it as a text node. thats everything this extension does for now. but we can improve it later further by including dropdown for emoji selection as user types etc

Sorry, maybe it is not necessary to create a different story, thanks for the explanation (with story I mean the storybook story, for example you have emoji_picker and also you have emoji_picker_with_popover for example).

How I can test this extension in the actual story?, What are the steps to reproduce the intent behaviour?

sanjeevdialpad:
oh gotcha. go to https://vue.dialpad.design/deploy-previews/pr-1080/ and rich test editor , type `:cat:` or any shortcode

iropolo:
> oh gotcha. go to https://vue.dialpad.design/deploy-previews/pr-1080/ and rich test editor , type `:cat:` or any shortcode

Okay cool, seems to work fine, lemme just review code now

components/rich_text_editor/extensions/emoji/emoji.js
```diff
@@ -0,0 +1,159 @@
...
+  }
+
+  // needs to be a dict returned
+  // ref type InputRuleMatch: https://github.com/ueberdosis/tiptap/blob/main/packages/core/src/InputRule.ts#L16
+  return {
+    index: match.index,
+    text: match[0],
+    match,
+  };
+};
```
iropolo: This could be simplify

```
const shortCodeInputMatch = (text) => {
  const match = text.match(inputShortCodeRegex);

  if (match && shortcodeToEmojiData(match[0])) {
    return {
      index: match.index,
      text: match[0],
      match,
    };
  };
};
```

components/rich_text_editor/extensions/emoji/emoji.js
```diff
@@ -0,0 +1,159 @@
...
+          if (!codeToEmojiData(code)) {
+            return;
+          }
+
+          // needs to be a dict returned
+          // ref type InputRuleMatch: https://github.com/ueberdosis/tiptap/blob/main/packages/core/src/InputRule.ts#L16
+          return {
+            index: match.index,
+            text: match[0],
+          };
```
iropolo: this is very similar as the above code, could we extract it maybe into a function?

components/rich_text_editor/extensions/emoji/emoji.js
```diff
@@ -0,0 +1,159 @@
+import { mergeAttributes, Node, nodeInputRule, nodePasteRule } from '@tiptap/core';
+import { VueNodeViewRenderer } from '@tiptap/vue-3';
+
```
iropolo: Non necessary space

sanjeevdialpad:
we can come back to the test i feel. For now what i will do is go through proseMirror documentation which is what tiptap uses to see what they use to match the input regex. Theres no way we can figure this out without knowing what it does under the hood. I thought i wouldnt need to know the abstracted details but 😭

components/rich_text_editor/extensions/emoji/emoji.test.js
```diff
@@ -0,0 +1,94 @@
...
+    attachTo: document.body,
+  });
+};
+
+describe('DtRichTextEditor Emoji Extension tests', () => {
+  // Test Setup
+  beforeAll(() => {
+    global.Range.prototype.getClientRects = vi.fn(() => [{}]);
+    global.Range.prototype.getBoundingClientRect = vi.fn(() => [{}]);
+    global.scrollBy = vi.fn();
```
iropolo:
```
const getClientRectsMock = vi.fn(() => [{}]);
const getBoundingClientRect = vi.fn(() => [{}]);
const scrollByMock = vi.fn();
```

components/rich_text_editor/extensions/emoji/emoji.js
```diff
@@ -0,0 +1,159 @@
...
+  });
+
+  return validMatches.map(match => {
+    return {
+      index: match.index,
+      text: match[0],
+      match,
+    };
+  });
+};
```
iropolo: Could be simplified to

```
const shortCodePasteMatch = (text) => {
  const matches = [...text.matchAll(pasteShortCodeRegex)];

  return matches
    .filter(match => codeToEmojiData(match[0]))
    .map(match => ({
      index: match.index,
      text: match[0],
      match,
    }));
};
```

iropolo (changes requested):
Simplify code suggestion are not necessary.
But all the other comments I will request changes.

components/rich_text_editor/extensions/emoji/EmojiComponent.vue
```diff
@@ -0,0 +1,24 @@
+<template>
+  <node-view-wrapper class="d-d-inline-block">
+    <dt-emoji
+      size="200"
+      :code="node.attrs.code"
+    />
+  </node-view-wrapper>
```
braddialpad: Hmm so when the screenreader reads this emoji in the input it doesn't read it's proper description ex "grinning face with smiling eyes" instead it just says "object replacement character". The description is correctly set on the emoji but it's also wrapped with `data-node-view-wrapper` perhaps this is what's preventing it from being read?

components/rich_text_editor/extensions/emoji/emoji.js
braddialpad, line 1: Kind of a weird thing.. but when you highlight the text with mouse and the last character is an emoji it doesn't highlight.

braddialpad (changes requested):

components/rich_text_editor/extensions/emoji/EmojiComponent.vue
```diff
@@ -0,0 +1,24 @@
+<template>
+  <node-view-wrapper class="d-d-inline-block">
+    <dt-emoji
+      size="200"
```
braddialpad: The emoji is quite small, if you look in slack for example, it seems it's quite typical to render the emoji at a size that is actually larger than the text. Maybe experiment with size 300 and 400.

Screenshot 2023-07-21 at 11 49 34 AM

sanjeevdialpad: yea i thought 300 was looking too big but now you mentioned slack , its a good idea to change it to that size.

components/rich_text_editor/extensions/emoji/emoji.js
sanjeevdialpad: good catch, i did make it selectable:true, but there might be something else.

components/rich_text_editor/extensions/emoji/EmojiComponent.vue
sanjeevdialpad: that node viewer wrapper is needed for custom node viewer. let me check docs on how to go about this. thanks!

components/rich_text_editor/extensions/emoji/emoji.js
sanjeevdialpad: seems good, we follow the if not pattern in DPM code for readability, hence i did that lol

components/rich_text_editor/extensions/emoji/emoji.js
sanjeevdialpad: I could not override the function as its defined on the core package with typescript typedef. I abstracted the match logic which is common to a common function for now.

components/rich_text_editor/extensions/emoji/emoji.js
sanjeevdialpad: I checked and went through proseMirror docs as well and the only option i see is the `selectable: true` which i have done already. It actually does highlight it but you have to begin dragging a bit before or after the point where emoji is. Double clicking to select also doesnt seem to work. I will keep looking but for now i think we have to do with this and log a bug.

braddialpad: k cool... I don't think this is that big of a deal it's kind of an edge case.

components/rich_text_editor/extensions/emoji/EmojiComponent.vue
braddialpad: This issue is too complex to fix atm. We can log and try to fix later. Even slacks message input does not do this correctly.

sanjeevdialpad:
made the changes requested but i could not fix two things:
1. Screen reader to read out emoji description. 
2. Emoji can be highlighted and selected but theres an edge case where sometimes it does not highlight properly on double click. or dragging cursor over emoji.

braddialpad (approved):

iropolo:
> made the changes requested but i could not fix two things:
> 
> 1. Screen reader to read out emoji description.
> 2. Emoji can be highlighted and selected but theres an edge case where sometimes it does not highlight properly on double click. or dragging cursor over emoji.

It says `object replacement character` on slack too...

iropolo (approved):
Thanks! 🚀

PR #1081, opened 2023-07-19
test(percy): fix visual tests

juliodialpad:
# Fix visual tests

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Percy and storybook v7 aren't fully compatible currently, so I changed the way we are taking snapshots to take a snapshot of every story instead of packing them on a single story.

I was having a lot of issues with weird errors coming from Storybook but those were gone after changing `discovery.concurrency` to 1 which makes the process of getting and uploading assets lineal instead of concurrent.

**note**: Had to remove `DtDatepicker` for now as percy is having trouble taking snapshots of it, I'll create a ticket to fix that later.

The rest of the test should work perfectly now.


components/toast/toast.stories.js
```diff
@@ -130,6 +130,14 @@ export const Default = {
...
     kind: 'base',
   },
+
+  parameters: {
+    percy: {
+      args: {
+        show: true,
+      },
+    },
+  },
```
braddialpad, line 140: Should we just make this the default for all stories and then set it to false on the ones we don't want? You can set parameters globally in `preview.js`

juliodialpad: It's possible, but we are setting different variables on different stories so it's not completely generic.

braddialpad: Oh I see it's the actual "show" prop. I thought this was a built in Percy attribute that enables whether to do visual testing or not

braddialpad (approved):

PR #1087, opened 2023-07-20
fix: fix emoji skin tone not updating when skinTone prop updates

cjmollitor-dialpad:
# [DLT-1152](https://dialpad.atlassian.net/browse/DLT-1152): Fix Emoji Skin Selector Prop

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

There's an issue currently where the emoji skin selector only shows `props.skinTone` for the initial value set. If the value ever changes, it doesn't seem to reflect. To see the issue in action, go to the [storybook](http://localhost:9010/?path=/story/components-emoji-picker--default&args=skinTone:Default) and try changing the skin tone from the props dropdown.

This change modifies the logic so that it'll initially take the value of `props.skinTone`. If a new value is selected from the selector, it'll display that value but if at any point `props.skinTone` changes, that value will be overwritten.

## :bulb: Context

N/A

## :pencil: Checklist

- [x] I have added tests

## :link: Sources

[DLT-1152]: https://dialpad.atlassian.net/browse/DLT-1152?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


cjmollitor-dialpad:
@iropolo I'll add tests after the initial round of review just in case this isn't a change we want to make or in case there's another way we should go about it

iropolo:
Hello!, thanks for taking your time reviewing it :).

What is the case were it will be used for?. 
In which case you will update a skin tone by prop with the component opened?

I can reproduce it if you have two emoji picker opened (or with v-show) and you update the skin tone in one.
But if you have another emoji-picker it will probably be in a completely other component or section or at least probably using v-if.
It seems like this new 'behaviour' handles a `wrong usability` of the component.

Initially the emoji-picker take the given by props just in case the user already has a skin tone previously saved. 
If this is not the case it will go with the `Default` one. If you want to update the skin tone you will use the skin-tone selector, thats why it also emits an event when it changes.

I barely remember this behaviour and I guess this is non harmful side effect.

Storybook shows it more explicit because the component is always open and also it has two handle ways, by storybook and by the component.

However this is a suitable solution for this :) just update `skinPassedIn` to `skinPassedIn.value` in the `watchEffect` and it will be ready to go.

braddialpad:
Yeah I believe this was actually intended behaviour to set the initial skin tone when it is opened, and from there if you wish to set the skin tone you do it from the emoji picker internally. Is there a case where you would actually need to set the skin tone from somewhere else when the emoji picker is actively open?

cjmollitor-dialpad:
> What is the case were it will be used for?. In which case you will update a skin tone by prop with the component opened?

We'll be using this in DP2 for the emoji pickers in feed item actions and the message input. Each feed item and the message input will have its own emoji picker. [This is the PR](https://github.com/dialpad/web-clients/pull/517/files#diff-b04de9fe32a82ad4c6928306595e67b18c6aa4983c13a5ec982caa997be00566) we currently have open for adding feed item reactions via the emoji picker. I'm currently based off this branch working on adding the ability to switch your default skin tone which isn't out for review yet but I can share it with you if you want to take a look.

> I can reproduce it if you have two emoji picker opened (or with v-show) and you update the skin tone in one. But if you have another emoji-picker it will probably be in a completely other component or section or at least probably using v-if. It seems like this new 'behaviour' handles a `wrong usability` of the component.

Is this in the app that you're reproducing this behavior? I noticed that behavior in DP2 when developing.

> Initially the emoji-picker take the given by props just in case the user already has a skin tone previously saved. If this is not the case it will go with the `Default` one. If you want to update the skin tone you will use the skin-tone selector, thats why it also emits an event when it changes.

I see I see, so just to clarify: was the intention of the original design that the emoji picker would only render skinTone prop once on the first render if it was present?

> Storybook shows it more explicit because the component is always open and also it has two handle ways, by storybook and by the component.
> 
> However this is a suitable solution for this :) just update `skinPassedIn` to `skinPassedIn.value` in the `watchEffect` and it will be ready to go.

Sounds good! I'll make those changes and write some tests. Thanks Nacho!

cjmollitor-dialpad:
> Yeah I believe this was actually intended behaviour to set the initial skin tone when it is opened, and from there if you wish to set the skin tone you do it from the emoji picker internally. Is there a case where you would actually need to set the skin tone from somewhere else when the emoji picker is actively open?

Ah darn I didn't see this until I submitted my other comment 😅

> Is there a case where you would actually need to set the skin tone from somewhere else when the emoji picker is actively open?

I think there is. The issue I'm noticing in DP2 is that, say you have feed items A and B and you can open the emoji picker for each of them via a popover in feed item actions and you've already opened each of them once. If you then go to feed item A's emoji picker and select a new skin tone, we're going to save this in the backend on the `UserSettings` model and we want this skintone to be present in all other emoji pickers in the app afterwards. The issue though is that since we've already opened the emoji picker once for feed item B, when the skin tone prop changes to the newly saved default skintone, it doesn't update the default skin tone in that emoji picker since currently the prop only sets it once on render (i.e. any subsequent changes to the prop don't change the displayed skin tone in the picker). We might be able to fix this behavior by `v-if`ing the emoji picker or popover instead but I actually wasn't able to render the emoji picker popover at all when I tried that out. I could take another crack at that though if need be.

Now that I'm stepping back and reading this, that feels really wordy and potentially confusing lol. Let me know if you want to hop on a call or something and I can share my screen to show what I'm talking about.

braddialpad:
I see what you're saying, I'm a bit surprised it doesn't reinitialize the props when you open it because I believe the popover is supposed to fully unmount when closed. That said.. I don't really know what sort of optimizations vue is doing itself in cases like this.

I think it's probably safer to just make this change and have it work more predictably

braddialpad (approved):

iropolo:
As I said it is not necessary to over engineer this, there is a case where the component could be off-sync, whatever is the case it should not.

It is okay to move on with this addition.

iropolo:
> I see what you're saying, I'm a bit surprised it doesn't reinitialize the props when you open it because I believe the popover is supposed to fully unmount when closed. That said.. I don't really know what sort of optimizations vue is doing itself in cases like this.
> 
> I think it's probably safer to just make this change and have it work more predictably

I tried to replicate it using a emoji_picker_with_popover and a emoji_picker together and they are out of sync. 
Seems like the popover does not destroy and mount the component again so that generates the issue.

iropolo:
> > I see what you're saying, I'm a bit surprised it doesn't reinitialize the props when you open it because I believe the popover is supposed to fully unmount when closed. That said.. I don't really know what sort of optimizations vue is doing itself in cases like this.
> > I think it's probably safer to just make this change and have it work more predictably
> 
> I tried to replicate it using a emoji_picker_with_popover and a emoji_picker together and they are out of sync. Seems like the popover does not destroy and mount the component again so that generates the issue.

...having said this .. that could be an issue later... heads up with this @cjmollitor-dialpad Im not sure about the performance if you have a ton of this emoji_picker running around...

braddialpad:
> Seems like the popover does not destroy and mount the component again so that generates the issue.

Do you mean it remains in the DOM after closing? Because that should not happen

iropolo:
> > Seems like the popover does not destroy and mount the component again so that generates the issue.
> 
> Do you mean it remains in the DOM after closing? Because that should not happen


TBH I dont know at this point...

components/emoji_picker/emoji_picker.test.js
```diff
@@ -171,6 +157,58 @@ describe('DtEmojiPicker Tests', () => {
...
+      it('Should render provided skin tone in skin tone selector', () => {
+        const skinToneSelector = wrapper.find('.d-emoji-picker__skin-selected button img');
+
+        expect(skinToneSelector.exists()).toBe(true);
+        expect(skinToneSelector.attributes('aria-label')).toBe(':wave_tone1:');
+      });
+
+      it('Should display skin tone selected from selector', async () => {
+        const skinToneSelectorButton = wrapper.find('.d-emoji-picker__skin-selected button');
+        const skinToneSelectorImg = wrapper.find('.d-emoji-picker__skin-selected button img');
```
iropolo: Create and move all `wrapper.find` inside

```
const _setChildWrappers = () => {
     const skinToneSelectorButton = wrapper.find('.d-emoji-picker__skin-selected button');
     const skinToneSelectorImg = wrapper.find('.d-emoji-picker__skin-selected button img');
     const skinToneMediumLightButton = wrapper.find('.d-emoji-picker__skin-list').findAll('button').at(1);
     const skinToneSelectorButton = wrapper.find('.d-emoji-picker__skin-selected button');
     const skinToneSelectorImg = wrapper.find('.d-emoji-picker__skin-selected button img');
     const skinToneMediumLightButton = wrapper.find('.d-emoji-picker__skin-list').findAll('button').at(1);
  };
```

then call it after inside the `setWrapper` function

iropolo (changes requested):

iropolo (approved):
Great, Thanks!

iropolo:
@juliodialpad percy here is rejecting this PR due a difference in screenshot.

Checking it you will see there is a difference in the scroller component, and that's okay because that content is generated dynamically and does not impact this PR.

There is a chance where pull reviews will be affected by this. Could we ignore it somehow?

juliodialpad:
> @juliodialpad percy here is rejecting this PR due a difference in screenshot.
> 
> Checking it you will see there is a difference in the scroller component, and that's okay because that content is generated dynamically and does not impact this PR.
> 
> There is a chance where pull reviews will be affected by this. Could we ignore it somehow?

Approved the visual-test as it doesn't have anything to do with this PR. 

@iropolo is fixing this later.

PR #1088, opened 2023-07-21
fix(tooltip): change anchor tag to span

braddialpad:
# fix(tooltip): change anchor tag to span

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Changed anchor tag on tooltip to a span. I think the real issue was just that it needed to be a `display: inline` element rather than a `display: block` element, so no reason for the `<a>` tag

## :bulb: Context

Users were sometimes having a tooltip navigate to the root page when clicked because it was wrapped in an `<a>` tag. Additionally it is invalid html to have buttons nested within `<a>` tags which is a very common way to use a tooltip.

## :pencil: Checklist

### DIFF
--- a/components/tooltip/tooltip.vue
+++ b/components/tooltip/tooltip.vue
@@ -1,9 +1,10 @@
 <template>
-  <div data-qa="dt-tooltip-container">
-    <a
+  <span data-qa="dt-tooltip-container">
+    <!-- disabling as the below events are for capturing events from interactive
+         elements within the span rather than on the span itself -->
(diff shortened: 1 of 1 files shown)


iropolo (approved):
If I update the the span element with display block I cannot replicate the original issue of corners. 🤔 

But if it fixs it as the `<a>` its okay for me to go with this

braddialpad:
> If I update the the span element with display block I cannot replicate the original issue of corners. 🤔
> 
> But if it fixs it as the `<a>` its okay for me to go with this

I had to test on a circle button. A regular button it is very hard to tell

PR #1101, opened 2023-07-26
fix(tooltip): anchor display property issue

braddialpad:
# fix(tooltip): anchor display property issue

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Previous PR was meant to fix the issue of the `<a>` tag but I changed both the container tag and anchor tag to spans. Changing the container tag caused issues with popovers dialogs showing higher than they should and overlapping the anchor in an awkward way. I should have only changed the anchor and not the outer container to a span. This fixes the issue.

Also added a tooltip variant for circle button tooltip so we can easily test if the "corners" are cut off

## :pencil: Checklist

### DIFF
--- a/components/tooltip/tooltip.vue
+++ b/components/tooltip/tooltip.vue
@@ -1,5 +1,5 @@
 <template>
-  <span data-qa="dt-tooltip-container">
+  <div data-qa="dt-tooltip-container">
     <!-- disabling as the below events are for capturing events from interactive
          elements within the span rather than on the span itself -->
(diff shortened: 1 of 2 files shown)


iropolo (approved):

iropolo:
Should we have vue3 branch like https://github.com/dialpad/dialtone-vue/pull/1089 ?

braddialpad:
> Should we have vue3 branch like #1089 ?

yep incoming

PR #1098, opened 2023-07-26
fix(avatar): thin border behind image

juliodialpad:
# Fix (Avatar): Thin border behind image

## :hammer_and_wrench: Type Of Change

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Removed almost invisible thin border behind DtAvatar's image

## :bulb: Context

Found a thin line that was almost no visible, but that was making our visual tests to flaky sometimes due to that thin line being random color.

## :pencil: Checklist

- [x] I have reviewed my changes
- [ ] I have added tests
- [ ] I have added all relevant documentation
- [ ] I have validated components with a screen reader
- [ ] I have validated components keyboard navigation
- [ ] I have considered the performance impact of my change
- [ ] I have checked that my change did not significantly increase bundle size
- [ ] I am exporting any new components or constants in the index.js in the component directory
- [ ] I am exporting any new components or constants in the index.js in the root

## :camera: Screenshots / GIFs

image
image

PR #1097, opened 2023-07-26
fix(contact-info): avatar's border color

juliodialpad:
# Fix (Contact info): Avatar's border color

## :hammer_and_wrench: Type Of Change

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Removed utility classes from `DtRecipeContactInfo` and added custom CSS to scoped style tag

## :bulb: Context

Border color for contact info's avatar was not ok, making it look off on dark-mode.

## :pencil: Checklist

- [x] I have reviewed my changes
- [ ] I have added tests
- [ ] I have added all relevant documentation
- [ ] I have validated components with a screen reader
- [ ] I have validated components keyboard navigation
- [ ] I have considered the performance impact of my change
- [ ] I have checked that my change did not significantly increase bundle size
- [ ] I am exporting any new components or constants in the index.js in the component directory
- [ ] I am exporting any new components or constants in the index.js in the root

## :camera: Screenshots / GIFs

Before
image

After
image

PR #1102, opened 2023-07-26
feat(avatar): extract initials from full name

juliodialpad:
# Feat (Avatar): Extract initials from full name

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

- Refactored Avatar to remove slot based usage in favor of prop based
- Added `iconSize` property to be able to resize the avatar's icon
- Removed slot related documentation
- Migrated avatar's usage on components that were using `DtAvatar`
- Changed imported images to public path to improve Percy visual tests stability with images
- Included changes from #1097 and #1098 

## :bulb: Context

We we're having a lot of issues maintaining avatar's as it was created slot based initially to make it more customizable.

## :pencil: Checklist

- [x] I have added tests

## :crystal_ball: Next Steps

Migrate usages of `DtAvatar`, `DtRecipeFeedItemRow`, `DtRecipeContactRow` and `DtRecipeContactInfo` components on product.

image
image
image
image
image
image
image
image
image
image
image


components/avatar/avatar.stories.js
```diff
@@ -82,12 +73,25 @@ export const argTypesData = {
...
+    control: {
+      type: 'select',
+      labels: {
+        undefined: '(empty)',
+      },
+    },
+  },
+  iconSize: {
+    defaultValue: null,
+  },
```
iropolo: there are props in the component were are not documented here, also could we sort alphabetically props here and in the component please? so many props its kind of messy

components/avatar/avatar.stories.js
```diff
@@ -12,27 +11,19 @@ const iconsList = getIconNames();
 export const argsData = {
   size: 'md',
   presence: null,
+  fullName: 'Jaqueline Nackos',
+  seed: null,
```
iropolo: this `seed: null,`

cause this in `controls`

image

components/avatar/avatar.vue
```diff
@@ -252,105 +286,82 @@ export default {
...
-      return element?.tagName?.toUpperCase() === 'IMG';
-    },
-
     getColor () {
       return this.color ?? getRandomElement(AVATAR_COLORS, this.seed);
     },
 
-    validateImageAttrsPresence () {
-      const isSrcMissing = !this.$refs.canvas.firstChild.getAttribute('src');
+    _extractInitials (string) {
```
iropolo: Could we move this `_extractInitials (string)` to an util function inside avatar folder?

iropolo: Code suggestion:
change `string` to `fullName`

do the trim direclty on the names constant and get the initials in a for loop

```
extractInitials (fullName) {
      if (typeof fullName !== 'string' || !fullName.trim()) return '';

      const names = fullName.trim().split(' ');
      let initials = '';

      for (const name of names) {
        initials += name.substring(0, 1).toUpperCase();
      }

      return initials;
    },
```

components/avatar/avatar.mdx
```diff
@@ -15,22 +15,10 @@ import * as AvatarStories from './avatar.stories.js';
   <Story of={AvatarStories.Default} />
 </Canvas>
 
-## Icon
+## Variants
```
iropolo, line 30: Could we update the doc to explain all these new behaviours and how it will works?

components/avatar/avatar.test.js
```diff
@@ -102,93 +98,79 @@ describe('DtAvatar Tests', () => {
...
-
-      // Test Setup
+    describe('With no imageSrc or iconName is provided', () => {
       beforeEach(async () => {
-        slots = { default: initials };
         await _setWrappers();
       });
 
       it('should display initials', () => {
+        console.log(wrapper);
```
iropolo: remove console.log

recipes/leftbar/contact_row/contact_row.vue
```diff
@@ -14,28 +14,14 @@
   >
     <template #left>
       <dt-avatar
+        :full-name="name"
+        :image-src="avatarSrc"
+        :icon-name="noInitials ? 'user' : null"
```
iropolo: could we move this to a computed property so we will not use logic in template

iropolo: Also I see in some places we are using
`:icon-name="avatarIcon"` or `:icon-name="iconName"` it is probably because in some cases is taken by stories.js, could we take a look on this please?

iropolo (changes requested):
Thanks! a lot of updates and improvements on this PR, great job!!

components/avatar/avatar.vue
braddialpad, line 1: Since we're already working on this, can we remove all utility classes in this file and replace with regular css / tokens?

Had a request for this this morning: https://dialpad.slack.com/archives/CEDUKPKC1/p1690465549087039

components/avatar/avatar.stories.js
juliodialpad: We shouldn't include all the properties there, only the ones that need set default values, hiding them from the panel and so forth, using JSDoc on the component itself it's enough for storybook and for our documentation extractor to work properly. I'll alphabetically sort them.

components/avatar/avatar.vue
braddialpad: Actually never mind I will do this in a separate commit to release faster

components/avatar/avatar.vue
juliodialpad: Sure, I'll extract the method to an `utils.js` file inside avatar folder.

Thanks for the suggestion, improved the code.

components/avatar/avatar.mdx
juliodialpad: All of that is self-explained in the properties description, I don't think it's necessary to add more documentation here. what do you think @braddialpad?

components/avatar/avatar.test.js
juliodialpad: oops 😅

components/avatar/avatar.vue
juliodialpad: Sure!

recipes/leftbar/contact_row/contact_row.vue
juliodialpad: > Also I see in some places we are using `:icon-name="avatarIcon"` or `:icon-name="iconName"` it is probably because in some cases is taken by stories.js, could we take a look on this please?

What should I do here? I don't see any problem really

juliodialpad:
Seems like visual tests are running even the PR is not approved yet, I'll take a look into that.

recipes/leftbar/contact_row/contact_row.vue
iropolo: k cool, nvm I thought maybe it's some like inconsistency 👍🏼

components/avatar/avatar.mdx
iropolo: IMO It would be cool to explain which behaviours bring Avatar beyond the props explanation.

This facilitates the quick understanding of what avatar performs.
For example explain the herarchy of what Avatar will show.

```
The Avatar component is designed to prioritize different sources for content display.
It will sequentially check for the availability of 
an image source `imageSrc`,
an icon name `iconName`
or the full name `fullName` to create initials. 

For example `Jaqueline Nackos` will be `JN`, some edge cases here are:
`Jaqueleine` = ??
`Jaqueline Nackos Lorem Ipsum` = ??

In the absence of any of these sources, the component will throw an error, as the fullName prop is required.

```

juliodialpad: Sounds good, adding more docs

components/avatar/avatar.vue
```diff
@@ -252,107 +296,77 @@ export default {
...
+    imageSrc: {
+      immediate: true,
+      async handler (newSrc) {
+        if (!newSrc) return;
+
+        this.validateProps();
+
+        this.imageLoadedSuccessfully = null;
+        await this.$nextTick();
+        this.setImageListeners();
```
braddialpad: It feels a bit flaky to do `nextTick` in an immediate watcher to wait for the UI to update. I think it would be much safer to change this to a non-immediate watcher and put the code within it into a method and call this method here and in `mounted()`

components/avatar/avatar.vue
```diff
@@ -252,107 +296,77 @@ export default {
...
-      // If alt set to empty string consider it valid, as this is a valid case if the
-      // image is already described by something else (ex: visible description)
-      // eslint-disable-next-line no-unneeded-ternary
-      const isAltMissing = this.$refs.canvas.firstChild.getAttribute('alt') === null;
-
-      if (isSrcMissing || isAltMissing) {
-        Vue.util.warn('src and alt attributes are required for image avatars', this);
-      }
-    },
-
```
braddialpad, line 344: Beautiful to get rid of all this code 😀

components/avatar/utils.js
```diff
@@ -0,0 +1,9 @@
+export const extractInitialsFromName = (fullName) => {
+  if (typeof fullName !== 'string' || !fullName.trim()) return '';
+
+  const names = fullName.trim().split(' ');
+
+  return names.length === 1
+    ? names.join('').substring(0, 2)
+    : names.slice(0, 2).map(name => name.substring(0, 1).toUpperCase()).join('');
```
braddialpad: Why do we `slice(0, 2)` what if the user had like 6 names? We should always use the first and the last.

components/avatar/avatar.vue
```diff
@@ -194,31 +203,76 @@ export default {
...
+
     avatarClasses () {
       return [
         'd-avatar',
         AVATAR_SIZE_MODIFIERS[this.validatedSize],
         this.avatarClass,
         {
           'd-avatar--group': this.showGroup,
-          [`d-avatar--color-${this.getColor()}`]: this.kind !== 'icon',
+          [`d-avatar--color-${this.getColor()}`]: !this.isIconType,
```
iropolo: Since `isIconType` seems is a computed only for this 
```[`d-avatar--color-${this.getColor()}`]: !this.isIconType,```

II suggest the computed name is `isNotIconType` and returns `!this.iconName;`

iropolo (changes requested):

iropolo:
Thanks for the update!!

components/avatar/utils.js
juliodialpad: That makes sense, I was thinking on just a name and a last name, fixing

iropolo (approved):
Thanks! 🚀

PR #1105, opened 2023-07-27
[NO-MERGE] [DISCUSSION] Test workflow reseearch

iropolo:
# Test workflow research 

## :book: Description

Important files:

- **old test template generated:** 
components/test_component/test_component.test.js 

- **new test template generated:**
components/test_component/test_component4.test.js

- **checkbox test migrated to new template:**
components/test_component/checkbox.test_new.test.js

- **icon test migrated to new template:**
components/test_component/icon.test_new.test.js

-----------------------
-----------------------

Notes:

- Removed functions like `itBehavesLikeAppliesChildProp` or `itBehavesLikeAppliesClassToChild` from `tests/shared_examples/extendability'`, It also removes `// eslint-disable-next-line vitest/expect-expect `
Not agree on the use of them.

I could agree on the use of sth like `expectClassToExist` so we can replace this:
```expect(wrapper.find(`.${CHECKBOX_INPUT_VALIDATION_CLASSES[VALIDATION_MESSAGE_TYPES.ERROR]}`).exists()).toBe(true);```
to this
```expectClassToExist(wrapper, CHECKBOX_INPUT_VALIDATION_CLASSES[VALIDATION_MESSAGE_TYPES.ERROR]);```
but still not sure.

- Removed beforeEach on test with a single test.

- Added initial test which will assert for the exist of the component. 
This is helpful because guaranteed an initial working test.

- Fixed some ways of write test.

Previously:
image

Now:
image

- Updated BeforeEach in one line, should have a break line.

- Updated test in one line, should have a break line.

- Updated test with a break line in the description, description should be in one line.

- Removed unnecessary comments.

- Tests now should follow certain rules while writing. (Its shown on the new migrated tets)
This will improve consistency between test and readability.
These are for example: 
All following test description will not have break line.
Non related sentences will have a break line.
Screenshot 2023-07-27 at 22 04 15

-  There is a possibility where you can update props using `setProps` method isntead the `mockProps` proposed by the test template. IMO we should follow `mockProps` for consitency.

- Outside test, will propose 3 `variable structuring`

image

- This template is easy to replicate in existing tests to do a quick migration.

### DIFF
--- a/components/test_component/checkbox.test_new.test.js
+++ b/components/test_component/checkbox.test_new.test.js
@@ -178,7 +178,6 @@ describe('DtTestComponent Tests', () => {
     });
 
     describe('When disabled', () => {
-      // Test Setup
       beforeEach(() => {
         mockProps = { disabled: true };
 
@@ -315,6 +314,7 @@ describe('DtTestComponent Tests', () => {
     describe('When indeterminate', () => {
       beforeAll(() => {
         mockProps = { indeterminate: true };
+
         updateWrapper();
       });
 
@@ -397,7 +397,6 @@ describe('DtTestComponent Tests', () => {
       });
 
       describe('When the checkbox group has a validation state', () => {
-        // Test Setup
         beforeEach(() => {
           _setGroupContext([], false, VALIDATION_MESSAGE_TYPES.SUCCESS);
         });
--- a/components/test_component/checkbox.test_new.test.js
+++ b/components/test_component/checkbox.test_new.test.js
@@ -1,0 +1,501 @@
+import { mount } from '@vue/test-utils';
+import DtCheckbox from '../checkbox/checkbox.vue';
+import { VALIDATION_MESSAGE_TYPES } from '@/common/constants.js';
+import { CHECKBOX_INPUT_VALIDATION_CLASSES } from '@/components/checkbox/checkbox_constants.js';
+
+const MOCK_SLOT_LABEL = 'My Slotted Label';
+const MOCK_SLOT_DESCRIPTION = 'My Slotted Description';
+const MOCK_INPUT_LISTENER_SPY = vi.fn();
+const MOCK_GROUP_NAME = 'checkboxGroup';
+
+let MOCK_ELEMENT = null;
+const MOCK_CUSTOM_CLASS = 'my-custom-class';
+const MOCK_PROP_NAME = 'some';
+const MOCK_PROP_VALUE = 'prop';
+const MOCK_CHILD_PROPS = {};
+
+const baseProps = {
+  label: 'My Checkbox Label',
+  value: 'Value',
+};
+const baseAttrs = {};
+const baseSlots = {};
+const baseProvide = {};
+
+let mockProps = {};
+let mockAttrs = {};
+let mockSlots = {};
+let mockProvide = {};
+
+describe('DtTestComponent Tests', () => {
+  let wrapper;
+  let checkbox;
+  let input;
+  let label;
+  let description;
+  let labelDescriptionContainer;
+
+  const updateWrapper = () => {
+    wrapper = mount(DtCheckbox, {
+      props: { ...baseProps, ...mockProps },
+      attrs: { ...baseAttrs, ...mockAttrs },
+      slots: { ...baseSlots, ...mockSlots },
+      global: {
+        provide: { ...baseProvide, ...mockProvide },
+      },
+    });
+
+    checkbox = wrapper.findComponent(DtCheckbox);
+    input = wrapper.find('input');
+    label = wrapper.find('[data-qa="checkbox-label"]');
+    description = wrapper.find('[data-qa="checkbox-description"]');
+    labelDescriptionContainer = wrapper.find('[data-qa="checkbox-label-description-container"]');
+  };
+
+  beforeEach(function () {
+    updateWrapper();
+  });
+
+  afterEach(function () {
+    mockProps = {};
+    mockAttrs = {};
+    mockSlots = {};
+    mockProvide = {};
+  });
+
+  describe('Presentation Tests', () => {
+    it('Should render the component', () => {
+      expect(wrapper).toBeDefined();
+      expect(checkbox).toBeDefined();
+    });
+
+    it('should have checkbox group class', () => {
+      expect(wrapper.find('.d-checkbox-group').exists()).toBe(true);
+    });
+
+    describe('Checkbox Input Tests', () => {
+      it('should exist', () => {
+        expect(input.exists()).toBe(true);
+      });
+
+      it('should have type checkbox', () => {
+        expect(input.attributes('type')).toBe('checkbox');
+      });
+
+      it('should not be checked', () => {
+        expect(input.element.checked).toBe(false);
+      });
+    });
+
+    describe('Checkbox Label Tests', () => {
+      it('should exist', () => {
+        expect(label.exists()).toBe(true);
+      });
+
+      it('should match provided label prop', () => {
+        expect(label.text()).toBe(baseProps.label);
+      });
+    });
+
+    describe('When a label prop is provided', () => {
+      it('should exist', () => {
+        mockProps = { label: 'My Label' };
+
+        updateWrapper();
+
+        expect(label.exists()).toBe(true);
+      });
+    });
+
+    describe('When a description prop is provided', () => {
+      it('should exist', () => {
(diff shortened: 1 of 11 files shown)


braddialpad:
Expanding on some of the questions asked in our meeting:

> Removed functions like itBehavesLikeAppliesChildProp or itBehavesLikeAppliesClassToChild

Agree with this, however in some cases custom expects may be useful: https://vitest.dev/api/expect.html#expect

Perhaps on highly reused assertions, or very complex assertions.

> There is a possibility where you can update props using setProps method isntead the mockProps proposed by the test template. IMO we should follow mockProps for consitency.

I agree this will probably work for most cases, however there will be some cases we want to test changing a prop after the component has mounted. This is necessary for testing reactivity of props outside of the initial render. In these cases we'll have to use `setProps`.

> Outside test, will propose 3 variable structuring

hmm not sure field names and expected values are really considered mocks, should they have this prefix?

> Previously:
image

> Now:
image

This certainly makes our tests have less boilerplate.. My only concern is that developers that aren't as experienced in dialtone vue as we are will just copy paste the existing it statement repeating the setup of state in every test when really it should be moved to a beforeEach. Also the state change itself is not attached to what describes it. Ex: "when the group is disabled" does not actually disable the group. "should disable input" does. And I'd say the assertion being named "should disable input" is probably not a good description of the assert? It should be more like "input should be disabled".

> Screenshot 2023-07-27 at 22 04 15

A very common way to write tests is "AAA" Arrange (setup), Act (trigger), Assert (validate).  https://medium.com/@pjbgf/title-testing-code-ocd-and-the-aaa-pattern-df453975ab80

Maybe we should space by this method? If we did this we would put the mockProps and updateWrapper together as they are both part of Arrange

> styling changes

For styling changes lets enforce linter rules as much as possible, most of those are probably already rules that exist that we just need to configure differently.

braddialpad:
Some thoughts on additional things to add:

- Coverage reports
- Is there anything that we should set and or mock for every single test? perhaps disable for specific tests if needed. Mostly looking to make it easier to work with more complex setup here.
- This is perfect for running on lint-staged! https://vitest.dev/guide/cli.html#vitest-related

iropolo:
> Expanding on some of the questions asked in our meeting:
> 
> > Removed functions like itBehavesLikeAppliesChildProp or itBehavesLikeAppliesClassToChild
> 
> Agree with this, however in some cases custom expects may be useful: https://vitest.dev/api/expect.html#expect
> 
> Perhaps on highly reused assertions, or very complex assertions.
> 

On `very complex assertions` there is a chance were we are not doing the correct assertion.
But in this case I can agree on use custom expects.
This custom expect should be in the same test file or in a test-utils.js file in the same folder as the test.

Im still not convinced to use it in `highly reused assertions`. 
If the assertion is simple, just repeat it, thats what you expect when you check a test.

Im not convinced on the idea to have a shared folder where it has multiple kind of assertions. because thats mean you will probably need to check that folder to see if you can find one for you before you are writing your test, if thats not the case you will probably want to create it and add it on the folder.

The result of this is a folder which contains a bunch of assertion were they are not shared (or most of them) and you may need to look at it when you create a new test.
This is another reason on why I would prefer to have this `custom expect` in the same test file or in a test-utils.js file in the same folder as the test.

Im agree on have functions to set environment and move this functions to a `test-utils.js` or have it in the same test.
It would be little bit less noisy.

Maybe Im not understading this point or how you want to implement it, but I cannot see it in the way we are doing it now.

It would be cool if we can move forward with this template and migrate our current test to this.
Then if it is necessary, we can make another round of workflow research.

> > There is a possibility where you can update props using setProps method isntead the mockProps proposed by the test template. IMO we should follow mockProps for consitency.
> 
> I agree this will probably work for most cases, however there will be some cases we want to test changing a prop after the component has mounted. This is necessary for testing reactivity of props outside of the initial render. In these cases we'll have to use `setProps`.
> 

Agree, if you expect a reactivity on prop change it will be needed to use `.setProps`. 

There might be a confustion when you use `mockProp` (to set environment) and `.setProps` for a reactivity update on the component.
This `.setProps` works with an `await`, like `.trigger` or `.setValue` or all this methods to manipulate the mounted component.

**I would have this in mind to create the contribution guide for test.**

> > Outside test, will propose 3 variable structuring
> 
> hmm not sure field names and expected values are really considered mocks, should they have this prefix?
> 

The idea to have a `mock` prefix whatever is `mock` or `MOCK_` is to indicate an specfic distinction of variables outside the component.
When you are mocking, you explicit indicate a `value` or an `expected value`.
The `base` props are also mock, but these would't change and will persist in the code.

If you have for example variables names without prefix or without something that indicates the nature of it, its kind of difficult to quickly recognize them. 

Im guessing this would help on our code health and readability in tests.


> > Previously:
> 
> image
> > Now:
> 
> image
> This certainly makes our tests have less boilerplate.. My only concern is that developers that aren't as experienced in dialtone vue as we are will just copy paste the existing it statement repeating the setup of state in every test when really it should be moved to a beforeEach. Also the state change itself is not attached to what describes it. Ex: "when the group is disabled" does not actually disable the group. "should disable input" does. And I'd say the assertion being named "should disable input" is probably not a good description of the assert? It should be more like "input should be disabled".
> 

I would like to have a better description  instead of add dummy code.
Environment configuration inside test is not wrong.
If you are using the same environment configuration in multiple tests you should do it in a beaforEach previously to avoid DRY.

I guess this is beyond dialtone-vue or dialtone or wathever repo, this is how you write code.
**In addition, I can document this in the contribution guide for test.**

> > Screenshot 2023-07-27 at 22 04 15
> 
> A very common way to write tests is "AAA" Arrange (setup), Act (trigger), Assert (validate). https://medium.com/@pjbgf/title-testing-code-ocd-and-the-aaa-pattern-df453975ab80
> 
> Maybe we should space by this method? If we did this we would put the mockProps and updateWrapper together as they are both part of Arrange
> 

Initially I started to diggin in some fancy ways to write code, but I decided to go with something "common" and quickly understandable for us and external contributors.

We can check it out this discussion after implementing this template. Because there is a lot of things we need to be in agree before lol.

> > styling changes
> 
> For styling changes lets enforce linter rules as much as possible, most of those are probably already rules that exist that we just need to configure differently.

Ofc, completely agree.
I would probably need to see how to implement them but initially **this would be explained in the contribution guide for test**

iropolo:
> Some thoughts on additional things to add:
> 
> * Coverage reports

We can move forward with this in a different ticket, I can build it later.

> * Is there anything that we should set and or mock for every single test? perhaps disable for specific tests if needed. Mostly looking to make it easier to work with more complex setup here.

The only thing Im guessing it would persist on every single test is the initial test provided by this new test template.

> * This is perfect for running on lint-staged! https://vitest.dev/guide/cli.html#vitest-related

Definetly to research in a different ticket and see how is the performance

iropolo:
I would start to create the contributing guide for test.

braddialpad:
Yep you can move ahead with this.

iropolo:
[Contributing Guideline - Writing Tests for Design System Components
](https://dialpad.atlassian.net/wiki/spaces/ENG/pages/726859796/Contributing+Guideline+-+Writing+Tests+for+Design+System+Components)

WIP

iropolo:
https://dialpad.atlassian.net/wiki/spaces/ENG/pages/726859796/Contributing+Guideline+-+Writing+Tests+for+Design+System+Components

Confluence doc approved and ticket to implement new template created.

https://dialpad.atlassian.net/jira/software/c/projects/DLT/boards/548/backlog?view=detail&selectedIssue=DLT-1170&issueLimit=100

We are ready to go with this 👍🏼

iropolo:
[Contributing Guideline - Writing Tests for Design System Components
](https://dialpad.atlassian.net/wiki/spaces/ENG/pages/726859796/Contributing+Guideline+-+Writing+Tests+for+Design+System+Components)


Confluence doc approved and ticket to implement new template created.

[DLT-1170](https://dialpad.atlassian.net/jira/software/c/projects/DLT/boards/548/backlog?view=detail&selectedIssue=DLT-1170&issueLimit=100)

We are ready to go with this 👍🏼 

[DLT-1170]: https://dialpad.atlassian.net/browse/DLT-1170?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #1107, opened 2023-07-27
fix(avatar): remove util classes, set tokens

braddialpad:
Vue3 pr https://github.com/dialpad/dialtone-vue/pull/1106


components/tooltip/tooltip_variants.vue
juliodialpad, line 1: This should be included on this PR?

juliodialpad (approved):
If tooltip changes need to be included here, this looks good, thanks!

components/tooltip/tooltip_variants.vue
braddialpad: hmm I am not sure what is going on with these.. it should have been only the single commit

braddialpad:
fixed

PR #1113, opened 2023-08-07
fix(contact-info): overlay icon alignment

fabiangarga:
# (contact_info): fix overlay icon alignment

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Update `dt-avatar` overlay class to fix the alignment when using the new callbar 

## :pencil: Checklist

- [ ] I have added tests

### DIFF
--- a/recipes/list_items/contact_info/contact_info.vue
+++ b/recipes/list_items/contact_info/contact_info.vue
@@ -21,7 +21,7 @@
             :key="index"
           >
             <dt-avatar
-              overlay-class="d-ac-center"
+              overlay-class="d-jc-center"
               :size="avatarSize"
               :seed="avatar.seed"
               :full-name="avatar.fullName"
--- a/recipes/list_items/contact_info/contact_info.vue
+++ b/recipes/list_items/contact_info/contact_info.vue
@@ -21,6 +21,7 @@
             :key="index"
           >
             <dt-avatar
+              overlay-class="d-ac-center"
               :size="avatarSize"
               :seed="avatar.seed"
               :full-name="avatar.fullName"


iropolo:
Hello!, Im not sure this class is getting any effect... 🤔 

image

juliodialpad (changes requested):
Please provide screenshots before and after this change

fabiangarga:
Thanks @juliodialpad @iropolo, my bad, updated with the wrong DT class.

Here are some screenshots

Before:
Screenshot 2023-08-07 at 19 11 54

After:
Screenshot 2023-08-07 at 19 13 10

iropolo:
Hmm ok... it is strange why the svg with `width: 100%` didnt fix it, seems like there is some width issue some around...
But okay Im guessing we can double check this with the `justify-content: center` property.

Im wondering if we can add this in the `.d-avatar__overlay` class in `avatar.vue` component to ensure the svg will be always in the middle... what you think @juliodialpad ?

fabiangarga:
> Im wondering if we can add this in the `.d-avatar__overlay` class in `avatar.vue` component to ensure the svg will be always in the middle... what you think @juliodialpad ?

Just in case, to only affect the new callbar I added to this recipe.

iropolo:
There is a chance where this is a issue that need to be fixed in dialtone side, lets wait for @juliodialpad  confirmation then I can create the PR for it.

juliodialpad:
I'm not sure where the issues are coming from, the [example on documentation](https://vue.dialpad.design/?path=/story/recipes-list-items-contact-info--variants) works perfectly without that class. Give me a minute to check on product if they have some custom class that's breaking the implementation

iropolo:
> I'm not sure where the issues are coming from, the [example on documentation](https://vue.dialpad.design/?path=/story/recipes-list-items-contact-info--variants) works perfectly without that class. Give me a minute to check on product if they have some custom class that's breaking the implementation

Maybe this help to debug,

Take a look on https://vue.dialpad.design/?path=/story/recipes-list-items-contact-info--variants

Initially the svg has 100% of width, if you update it for example to 25px you can replicate the issue.

Screenshot 2023-08-09 at 13 28 30

So what Im guessing is the 100% of width in the svg is not enough to ensure svg in the middle.

>  it is strange why the svg with `width: 100%` didnt fix it, seems like there is some width issue some around..

🤔

juliodialpad:
That makes sense, I'll investigate on product why they're changing the size.

But this should be fixed on `DtAvatar`, it's ok to add `justify-content: center` to be sure the icon is always in the middle.

Please close this PR, I'll create another one with the proper fix.

fabiangarga:
> That makes sense, I'll investigate on product why they're changing the size.
> 
> But this should be fixed on `DtAvatar`, it's ok to add `justify-content: center` to be sure the icon is always in the middle.
> 
> Please close this PR, I'll create another one with the proper fix.

Got it, I will add you to the Jira ticket then. Thanks.

PR #1118, opened 2023-08-14
test(checkbox, icon): new test template

iropolo:
# Test migrations

- [x] Refactoring

## :book: Description


.github/CONTRIBUTING.md
```diff
@@ -227,6 +227,8 @@ export {
 
 Our unit tests are written with [Vitest](https://vitest.dev/) and can be found alongside any javascript/vue code files existing in Dialtone Vue. Tests are suffixed with `.test.js` of the corresponding javascript/vue file. ex: `link.vue` is tested by `link.test.js`. All code in Dialtone Vue should have tests written for it. You can see more details on writing tests in the following [documentation](https://vue.dialpad.design/?path=/story/docs-component-driven-development-unit-tests--page).
 
+Before start to write tests, please follow the [Contributing Guideline - Writing Tests for Design System Components](https://dialpad.atlassian.net/wiki/spaces/ENG/pages/726859796/Contributing+Guideline+-+Writing+Tests+for+Design+System+Components).
+
```
juliodialpad: This documentation should be public, because dialtone-vue is open source

components/checkbox/checkbox.test.js
```diff
@@ -1,520 +1,507 @@
...
-        _setWrappers();
+    describe('When a label prop is provided', () => {
+      it('should exist', () => {
+        mockProps = { label: 'My Label' };
+
+        updateWrapper();
+
+        expect(label.exists()).toBe(true);
       });
+    });
```
juliodialpad, line 115: 1. Shouldn't this be inside `describe('Checkbox Label Tests')` ?
2. This is not actually testing that the passed label prop is equal to the label text. The label always exists as it is part of `baseProps`

juliodialpad (changes requested):
The readability of the tests definitely improve, thanks Nacho! 

Just a couple comments, but everything looks a lot better.

juliodialpad:
Remember to put the `no-visual-test` tag if the PR doesn't include any visual change.

components/checkbox/checkbox.test.js
iropolo: Im not doing a review of the test, just rewriting it with the new template. 

And I will not do a review of the test, it is not part of the ticket. I can do some modifications or improvements if I catch it but definitely not check if it is well asserted.

This new way...

```
describe('When a label prop is provided', () => {
      it('should exist', () => {
        mockProps = { label: 'My Label' };

        updateWrapper();

        expect(label.exists()).toBe(true);
      });
    });
```

Is the same as this old way...

```
describe('When a label prop is provided', () => {
      // Test Setup
      beforeEach(() => {
        propsData = { ...basePropsData, label: 'My Label' };
        _setWrappers();
      });

      it('should exist', () => { expect(label.exists()).toBe(true); });
    });
```

And regarding what you concern, yes seems like this test it is not fine...
As I can see in the `describe('When a label prop is provided'`

Seems like this code `propsData = { ...basePropsData, label: 'My Label' };` is not necessary and you can assert if the label exist directly because its already provided by the baseProps

So the new updated code should be 

```
describe('When a label prop is provided', () => {
      it('should exist', () => {
        expect(label.exists()).toBe(true);
      });
    });
```

.github/CONTRIBUTING.md
iropolo: Cool, where should I update it?

braddialpad: Either make the confluence page public (if that is even possible). Or transfer to markdown in github

braddialpad (approved):
Looks good. Easier to read.

.github/CONTRIBUTING.md
juliodialpad: IDK if it's possible to make Dialpad confluence link public.

CC: @braddialpad

components/checkbox/checkbox.test.js
juliodialpad: Cool, if it's out of the scope of this PR, this is ok.

.github/CONTRIBUTING.md
juliodialpad: > IDK if it's possible to make Dialpad confluence link public.
> 
> CC: @braddialpad

I haven'r read [brad comment](https://github.com/dialpad/dialtone-vue/pull/1118#discussion_r1294992573) when posting this.

I'll say it'd be easier and more according to how we're working to migrate this to a markdown file in GitHub just as README.md, CONTRIBUTING.md, etc.

iropolo:
Test contributing guideline markdown pushed 👍🏼

.github/TEST_CONTRIBUTING/TEST_GUIDELINE.md
```diff
@@ -0,0 +1,145 @@
...
+
+You can also check these tests to get familiar with how it is working now in [checkbox-component-test](../../components/checkbox/checkbox.test.js) or [icon-component-test](../../components/icon/icon.test.js)
+
+This template provides an initial scaffolding and follows the structure we expect for our tests.
+
+This template is also generated by yeoman which creates a new component with `npx yo dialtone-vue` script.
+
+
+## Organizing Variables for Effective Testing:
+[structuring_variables](assets/structuring_variables.png)
```
juliodialpad: Add an ! (exclamation mark) at the beginning of the images to inline them instead of making them links
```suggestion
![structuring_variables](assets/structuring_variables.png)
```

juliodialpad (changes requested):
Change images from links to inline

.github/TEST_CONTRIBUTING/TEST_GUIDELINE.md
iropolo: I did it in the beginning but I didn't like how its shows due different sizes. 
I would prefer to keep it as it.

juliodialpad: You can use HTML in markdown to make it the appropriate size if needed.

It's not usable that way, If you are going to include images, I'd like to see the images there, not links that take me out of the current docs.

https://stackoverflow.com/a/14747656/16506300

iropolo: I tryed with html also... width of 200, still does not convince due different sizes of images.
Also you cannot zoom it.

As I tested previously when you open the reference, it will open you in a different window I didn't see anything wrong besides normal behaviour.

I still see it more properly with links instead of images.

But as I see it is a required demand, let me adjust them individually to look it ok.

iropolo: Since it impact the readability of the document I will ping @braddialpad here to check it out.

For compare you can check vue-3 version.

braddialpad: It looks much better when the images are embedded. However why are we even using screenshots for code examples? Shouldn't we use a markdown code block? This is much more maintainable.

iropolo: lol right, lets move it to code block

juliodialpad: Ah right, I haven't thought about code blocks instead of images. That'd be better, IDK if it's possible to import the lines of code from the file, that'd be even better.

iropolo: Updated using codeblocks

.github/TEST_CONTRIBUTING/TEST_GUIDELINE.md
```diff
@@ -0,0 +1,468 @@
...
+You can also check these tests to get familiar with how it is working now in [checkbox-component-test](../../components/checkbox/checkbox.test.js) or [icon-component-test](../../components/icon/icon.test.js)
+
+This template provides an initial scaffolding and follows the structure we expect for our tests.
+
+This template is also generated by yeoman which creates a new component with `npx yo dialtone-vue` script.
+
+
+## Organizing Variables for Effective Testing:
+
+```md
```
braddialpad: Change to `js` and I think it will code highlight it correctly.

.github/TEST_CONTRIBUTING/TEST_GUIDELINE.md
braddialpad: good just need to update code language as mentioned below

.github/CONTRIBUTING.md
```diff
@@ -227,6 +227,8 @@ export {
 
 Our unit tests are written with [Vitest](https://vitest.dev/) and can be found alongside any javascript/vue code files existing in Dialtone Vue. Tests are suffixed with `.test.js` of the corresponding javascript/vue file. ex: `link.vue` is tested by `link.test.js`. All code in Dialtone Vue should have tests written for it. You can see more details on writing tests in the following [documentation](https://vue.dialpad.design/?path=/story/docs-component-driven-development-unit-tests--page).
 
+Before start to write tests, please follow the [Contributing Guideline - Writing Tests for Design System Components](TEST_CONTRIBUTING/TEST_GUIDELINE.md).
```
braddialpad: "Before you start"

braddialpad (approved):

juliodialpad (approved):
Looks good now, thanks!

PR #1127, opened 2023-08-18
test(breadcrumb, breadcrumb-item, button): new test template

iropolo:
# Test migrations

## :hammer_and_wrench: Type Of Change

- [x] Refactoring

## :book: Description

Implementing new test template discussed following [Contributing Guideline - Writing Tests for Design System Components](https://dialpad.atlassian.net/wiki/spaces/ENG/pages/726859796/Contributing+Guideline+-+Writing+Tests+for+Design+System+Components)

PR #1135, opened 2023-08-24
feat: feed item pill recipe

AlexLamDialpad:
# [feat: feed item pill recipe](https://dialpad.atlassian.net/browse/DLT-638)

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Creates a feed item pill recipe for DP2 to use for feed call pills

## :pencil: Checklist

- [x] I have added tests

image

image

On hover
image

### DIFF
--- a/recipes/conversation_view/feed_pill/feed_item_pill.stories.js
+++ b/recipes/conversation_view/feed_pill/feed_item_pill.stories.js
@@ -1,0 +1,102 @@
+import { createTemplateFromVueFile } from '@/common/storybook_utils';
+import DtRecipeFeedItemPill from './feed_item_pill.vue';
+import DtRecipeFeedItemPillDefaultTemplate from './feed_item_pill_default.story.vue';
+
+// Default Prop Values
+export const argsData = {
+};
+
+export const argTypesData = {
+  // Slots
+  subtitle: {
+    control: 'text',
+    table: {
+      type: {
+        summary: 'VNode',
+      },
+    },
+  },
+  bottom: {
+    control: 'text',
+    table: {
+      type: {
+        summary: 'VNode',
+      },
+    },
+  },
+  right: {
+    control: 'text',
+    table: {
+      type: {
+        summary: 'VNode',
+      },
+    },
+  },
+  content: {
+    control: 'text',
+    table: {
+      type: {
+        summary: 'VNode',
+      },
+    },
+  },
+};
+
+// Story Collection
+export default {
+  title: 'Recipes/Conversation View/Feed Item Pill',
+  component: DtRecipeFeedItemPill,
+  args: argsData,
+  argTypes: argTypesData,
+  excludeStories: /.*Data$/,
+};
+
+// Templates
+const DefaultTemplate = (args, { argTypes }) => createTemplateFromVueFile(
+  args,
+  argTypes,
+  DtRecipeFeedItemPillDefaultTemplate,
+);
+
+// Stories
+export const Default = {
+  render: DefaultTemplate,
+  parameters: { a11y: { disable: true } },
+  args: {
+    iconName: 'video',
+    title: 'This meeting has ended',
+    ariaLabel: 'Click to expand',
+    buttonClass: '',
+    toggleable: true,
+    wrapperClass: 'd-w628',
+    borderColor: 'default',
+  },
+};
+
+export const AiBorderGradient = {
+  render: DefaultTemplate,
+  parameters: { options: { showPanel: false }, controls: { disable: true }, a11y: { disable: true } },
+  args: {
+    iconName: 'video',
+    title: 'This meeting has ended',
+    wrapperClass: 'd-w628',
+    buttonClass: '',
+    ariaLabel: 'Click to expand',
+    toggleable: true,
+    borderColor: 'ai',
+  },
+};
+
+export const NoToggleVariant = {
+  render: DefaultTemplate,
+  parameters: { options: { showPanel: false }, controls: { disable: true }, a11y: { disable: true } },
+  args: {
+    iconName: 'video',
+    title: 'This meeting has ended',
+    wrapperClass: 'd-w628',
+    expanded: false,
+    ariaLabel: 'Click to expand',
+    toggleable: false,
+    borderColor: 'critical',
+  },
+};
--- a/recipes/conversation_view/feed_pill/feed_item_pill.test.js
+++ b/recipes/conversation_view/feed_pill/feed_item_pill.test.js
@@ -1,0 +1,117 @@
+import { mount } from '@vue/test-utils';
+import DtRecipeFeedItemPill from './feed_item_pill.vue';
+import { beforeEach } from 'vitest';
+
+describe('DtRecipeFeedItemPill Tests', function () {
+  let wrapper, feedItemPill, icon;
+
+  const MOCK_ARIA_LABEL = 'Click to expand';
+  const MOCK_ICON_NAME = 'Video';
+
+  const baseProps = {
+    iconName: MOCK_ICON_NAME,
+    title: 'This meeting has ended',
+    ariaLabel: MOCK_ARIA_LABEL,
+    buttonClass: '',
+    toggleable: true,
+  };
+  const baseAttrs = {};
+  const baseSlots = {
+    content: '<div data-qa="content-element"> content </div>',
+  };
+  const baseProvide = {};
+
+  let mockProps = {};
+  let mockAttrs = {};
+  let mockSlots = {};
+  let mockProvide = {};
+
+  const updateWrapper = () => {
+    wrapper = mount(DtRecipeFeedItemPill, {
+      propsData: { ...baseProps, ...mockProps },
+      attrs: { ...baseAttrs, ...mockAttrs },
+      slots: { ...baseSlots, ...mockSlots },
+      global: {
+        provide: { ...baseProvide, ...mockProvide },
+      },
+    });
+    feedItemPill = wrapper.find('[data-qa="dt-feed-item-pill"]');
+    icon = wrapper.find('[data-qa="dt-feed-item-pill-icon"]');
+  };
+
+  beforeEach(function () {
+    updateWrapper();
+  });
+
+  afterEach(function () {
+    mockProps = {};
+    mockAttrs = {};
+    mockSlots = {};
+    mockProvide = {};
+  });
+
+  describe('Presentation Tests', function () {
+    describe('Default render', function () {
+      it('should exist', () => {
+        expect(wrapper.exists()).toBe(true);
+      });
+
+      it('should render a feed item pill', () => {
+        expect(feedItemPill.exists()).toBeTruthy();
+        expect(icon.exists()).toBe(true);
+        expect(icon.attributes('data-name')).toBe(MOCK_ICON_NAME);
+        expect(wrapper.find('[data-qa="content-element"]').exists()).toBe(false);
+      });
+    });
+  });
+
+  describe('Accessibility Tests', function () {
+    describe('Default Render', function () {
+      it('should render a reaction button', () => {
+        expect(feedItemPill.attributes('aria-label')).toBe(MOCK_ARIA_LABEL);
+      });
+    });
+  });
+
+  describe('Interactivity Tests', function () {
+    describe('Toggle feed item pill', function () {
+      it('Should emit a click event', async () => {
+        await feedItemPill.trigger('click');
+        await wrapper.vm.$nextTick();
+
+        expect(wrapper.find('[data-qa="content-element"]').exists()).toBe(true);
+      });
+
+      describe('toggleable false', function () {
+        beforeAll(() => {
+          mockProps = {
+            toggleable: false,
+          };
+        });
+
+        it('should exist', () => {
+          expect(wrapper.exists()).toBe(true);
+        });
+
+        it('should not respond to clicks', () => {
+          expect(feedItemPill.exists()).toBeTruthy();
+          expect(icon.exists()).toBe(true);
+          expect(icon.attributes('data-name')).toBe(MOCK_ICON_NAME);
+          expect(wrapper.find('[data-qa="content-element"]').exists()).toBe(false);
+        });
+      });
+    });
+
+    describe('Hover Feed Item Pill event', function () {
+      beforeEach(async () => {
+        await feedItemPill.trigger('focusin');
+        icon = wrapper.find('[data-qa="dt-feed-item-pill-icon"]');
+      });
+
+      it('should show a different icon', () => {
+        expect(icon.exists()).toBe(true);
+        expect(icon.attributes('data-name')).toBe('Chevron Right');
+      });
+    });
+  });
+});
--- a/recipes/conversation_view/feed_pill/feed_item_pill.vue
+++ b/recipes/conversation_view/feed_pill/feed_item_pill.vue
@@ -1,0 +1,184 @@
+<template>
+  <div :class="['dt-feed-item-pill--border', borderClass, wrapperClass]">
+    <div class="d-p8 d-bgc-secondary">
+      <dt-collapsible :open="expanded">
+        <template #anchor>
+          <button
+            :aria-label="ariaLabel"
+            data-qa="dt-feed-item-pill"
+            :class="['d-baw0 d-bgc-moderate d-bar-pill d-w100p d-ta-left d-btn--circle', toggleableClass, buttonClass]"
+            @focusin="hover = true"
+            @focusout="hover = false"
+            @mouseenter="hover = true"
+            @mouseleave="hover = false"
+            @click="onClick"
+          >
+            <dt-item-layout class="d-w100p d-p8">
+              {{ title }}
+              <template #left>
+                <dt-icon
+                  size="300"
+                  class="dt-feed-item-pill--icon d-pr8"
+                  data-qa="dt-feed-item-pill-icon"
+                  :name="computedIcon"
+                />
+              </template>
+              <template #subtitle>
+                <slot name="subtitle" />
+              </template>
+              <template #bottom>
+                <slot name="bottom" />
+              </template>
+              <template #right>
+                <slot name="right" />
+              </template>
+            </dt-item-layout>
+          </button>
+        </template>
+        <template #content>
+          <div class="d-jc-center d-d-flex">
+            <slot name="content" />
+          </div>
+        </template>
+      </dt-collapsible>
+    </div>
+  </div>
+</template>
+
+<script>
+import { FEED_ITEM_PILL_BORDER_COLORS } from './feed_item_pill_constants';
+import { DtIcon, DtItemLayout, DtCollapsible } from '@/index';
+
+export default {
+  name: 'DtRecipeFeedItemPill',
+
+  components: { DtItemLayout, DtIcon, DtCollapsible },
+
+  props: {
+    /**
+     * Accepts a DtIcon name to be shown in the left
+     */
+    iconName: {
+      type: String,
+      default: () => '',
+    },
+
+    /**
+     * Bolded primary text
+     */
+    title: {
+      type: String,
+      default: () => '',
+    },
+
+    /**
+     * Additional styling around the pill
+     */
+    wrapperClass: {
+      type: [String, Array, Object],
+      default: '',
+    },
+
+    /**
+     * Additional styling for the pill
+     */
+    buttonClass: {
+      type: [String, Array, Object],
+      default: '',
+    },
+
+    /**
+     * Aria label for feed pill
+     */
+    ariaLabel: {
+      type: String,
+      required: true,
+    },
+
+    /**
+     * Sets whether the pill can be toggled (Icon changing on hover, expanding and collapsing, pointer)
+     */
+    toggleable: {
+      type: Boolean,
+      default: () => true,
+    },
+
+    /**
+     * Callbox border color
+     * @values default, ai, critical
+     */
+    borderColor: {
+      type: String,
+      default: 'default',
+      validator: (color) => Object.keys(FEED_ITEM_PILL_BORDER_COLORS).includes(color),
+    },
+  },
+
+  data () {
+    return {
+      hover: false,
+      expanded: false,
+    };
+  },
+
+  computed: {
+    computedIcon () {
+      if (this.toggleable && this.hover) {
+        return this.expanded ? 'chevron-down' : 'chevron-right';
+      } else {
+        return this.iconName;
+      }
+    },
+
+    toggleableClass () {
+      return this.toggleable ? 'd-c-pointer' : '';
+    },
+
+    borderClass () {
+      return FEED_ITEM_PILL_BORDER_COLORS[this.borderColor];
+    },
+  },
+
+  methods: {
+    onClick () {
+      if (!this.toggleable) return;
+
+      this.expanded = !this.expanded;
+    },
+  },
+};
+</script>
+
+<style lang="less" scoped>
+  // Gradient radius solution taken from https://stackoverflow.com/a/53037637
+  .dt-feed-item-pill--border {
+    border: double 1px transparent;
+    border-radius: 4.8rem; // Special value determined by designer here where it works in both expanded and collapsed
+    background-origin: border-box;
+    background-clip: content-box, border-box;
+    overflow: hidden;
+  }
+
+  .dt-feed-item-pill--border-default {
+    background: var(--dt-color-border-default)
+  }
+
+  .dt-feed-item-pill--border-ai {
+    background-image:
+      linear-gradient(var(--dt-color-surface-primary), var(--dt-color-surface-primary)),
+      var(--dt-badge-color-background-ai);
+  }
+
+  .dt-feed-item-pill--border-critical {
+    background: var(--dt-color-foreground-critical);
+  }
+
+  .dt-feed-item-pill--icon {
+    animation: fade 0.15s ease-in;
+  }
+
+  @keyframes fade {
+    0%   {transform: scale(0);}
+    100% {transform: scale(1);}
+  }
+</style>
--- a/recipes/conversation_view/feed_pill/feed_item_pill_constants.js
+++ b/recipes/conversation_view/feed_pill/feed_item_pill_constants.js
@@ -1,0 +1,9 @@
+export const FEED_ITEM_PILL_BORDER_COLORS = {
+  default: 'dt-feed-item-pill--border-default',
+  ai: 'dt-feed-item-pill--border-ai',
+  critical: 'dt-feed-item-pill--border-critical',
+};
+
+export default {
+  FEED_ITEM_PILL_BORDER_COLORS,
+};
--- a/recipes/conversation_view/feed_pill/feed_item_pill_default.story.vue
+++ b/recipes/conversation_view/feed_pill/feed_item_pill_default.story.vue
@@ -1,0 +1,55 @@
+<template>
+  <dt-recipe-feed-item-pill
+    :icon-name="iconName"
+    :title="title"
+    :wrapper-class="wrapperClass"
+    :button-class="buttonClass"
+    :border-color="borderColor"
+    :aria-label="ariaLabel"
+    :toggleable="toggleable"
+  >
+    <template #subtitle>
+      Last 43 minutes - Ended at 5:04pm
+    </template>
+    <template #bottom>
+      <span class="d-fc-tertiary d-fs-100">8 people joined</span>
+    </template>
+    <template #right>
+      <div class="d-pr16">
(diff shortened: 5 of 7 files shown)


recipes/conversation_view/feed_pill/feed_item_pill.test.js
```diff
@@ -0,0 +1,137 @@
...
+     */
+
+    describe('Click Feed Item Pill event', async function () {
+      it('Should emit a click event', async () => {
+        await feedItemPill.trigger('click');
+        itBehavesLikeEmitsExpectedEvent(wrapper, 'click');
+      });
+    });
+
+    // Why this no work with trigger focus or trigger mouseenter
```
AlexLamDialpad: Does this work? Didn't see a comparable test in the repo

recipes/conversation_view/feed_pill/feed_item_pill.vue
```diff
@@ -0,0 +1,135 @@
...
+          @focusin="hover = true"
+          @focusout="hover = false"
+          @mouseenter="hover = true"
+          @mouseleave="hover = false"
+          @click="$emit('click')"
+        >
+          <dt-item-layout class="d-w100p d-p8">
+            {{ title }}
+            <template slot="left">
+              <!-- TODO: Add transition styling to the icon on hover -->
```
AlexLamDialpad: There's a click animation inside the prototype 
https://www.figma.com/proto/8ib3sY0ZVuBMGoxxceG6Ja/Ai-Tiger-Team?page-id=4186%3A130594&type=design&node-id=4186-134058&viewport=313%2C369%2C0.54&t=4myGHU3w75xvGK2d-1&scaling=min-zoom&starting-point-node-id=4186%3A134058&show-proto-sidebar=1

Might take a crack at this if there's time, but if not then I think we're ok with out it for now

AlexLamDialpad:
Do I need to do anything here for screen readers? I'm assuming the consumer should pass in screen reader compatible content via slots

AlexLamDialpad:
After adding the animation and on hover icon behaviour mentioned in the prototype, I realized we probably want a way to disable that for simpler pills (Fax / screenshare) where there is not a expandable/toggleable behaviour.

recipes/conversation_view/feed_pill/feed_item_pill.test.js
juliodialpad, line 1: Read the [new test guideline](https://github.com/dialpad/dialtone-vue/blob/staging/.github/TEST_CONTRIBUTING/TEST_GUIDELINE.md) and write your test with that new standards please.

recipes/conversation_view/feed_pill/feed_item_pill.vue
```diff
@@ -0,0 +1,147 @@
+<template>
+  <div :class="[wrapperClass, expanded ? 'dt-feed-item--expanded-pill' : 'd-bar-pill']">
+    <div :class="['d-bgc-secondary', expanded ? 'dt-feed-item--expanded-pill' : 'd-bar-pill']">
```
juliodialpad: Make `expanded ? 'dt-feed-item--expanded-pill' : 'd-bar-pill'` a computed prop please

recipes/conversation_view/feed_pill/feed_item_pill.vue
```diff
@@ -0,0 +1,147 @@
...
+          @click="$emit('click')"
+        >
+          <dt-item-layout class="d-w100p d-p8">
+            {{ title }}
+            <template slot="left">
+              <dt-icon
+                v-if="toggleable && hover"
+                class="dt-feed-icon"
+                data-qa="dt-feed-item-hover-icon"
+                :name="expanded ? 'chevron-down' : 'chevron-right'"
```
juliodialpad: a computed prop here would be cleaner

recipes/conversation_view/feed_pill/feed_item_pill.vue
```diff
@@ -0,0 +1,147 @@
...
+      default: () => '',
+    },
+
+    /**
+     * Aria label for feed pill
+     */
+    ariaLabel: {
+      type: String,
+      default: () => '',
+    },
```
juliodialpad: make this property required as for it'd be needed for accessibility

recipes/conversation_view/feed_pill/feed_item_pill.vue
```diff
@@ -0,0 +1,147 @@
...
+     */
+    toggleable: {
+      type: Boolean,
+      default: () => true,
+    },
+  },
+
+  emits: [
+    'click',
+  ],
```
juliodialpad: Add a comment for documentation on this click event, you can copy/paste from another components with native click event.

recipes/conversation_view/feed_pill/feed_item_pill.vue
```diff
@@ -0,0 +1,147 @@
...
+
+  .dt-feed-icon {
+    animation: fade 0.2s ease-in;
+  }
+
+  @keyframes fade {
+    0%   {transform: scale(0);}
+    100% {transform: scale(1);}
+  }
+</style>
```
juliodialpad: Styles for recipe components need to be scoped https://github.com/dialpad/dialtone-vue/blob/staging/.github/CONTRIBUTING.md#coding-guidelines

Also make sure to be using [BEM naming convention](https://getbem.com/introduction/)

recipes/conversation_view/feed_pill/feed_item_pill_constants.js
juliodialpad, line 1: If this component is not going to have any constants it's safe to remove this file

recipes/conversation_view/feed_pill/feed_item_pill.vue
```diff
@@ -0,0 +1,147 @@
...
+  },
+};
+</script>
+
+<style lang="less">
+  .dt-feed-item--expanded-pill {
+    border-radius: 5rem
+  }
+
+  .dt-feed-icon {
```
juliodialpad:
```suggestion
  .dt-feed-item--icon {
```

recipes/conversation_view/feed_pill/feed_item_pill.vue
```diff
@@ -0,0 +1,147 @@
...
+  data () {
+    return {
+      hover: false,
+    };
+  },
+};
+</script>
+
+<style lang="less">
+  .dt-feed-item--expanded-pill {
```
juliodialpad:
```suggestion
  .dt-feed-item--pill-expanded {
```

recipes/conversation_view/feed_pill/feed_item_pill.vue
juliodialpad, line 1: It's recommended to avoid using utility classes on components as it could potentially create a lot of problems while trying to override them.

You should create custom `.dt-feed-item*` classes and use custom CSS using dialtone CSS variables to make sure this recipe is compatible with both light and dark modes.

recipes/conversation_view/feed_pill/feed_item_pill_default.story.vue
```diff
@@ -0,0 +1,104 @@
...
+    ariaLabel: {
+      type: String,
+      default: 'Click to expand',
+    },
+
+    toggleable: {
+      type: Boolean,
+      default: false,
+    },
+  },
```
juliodialpad: You don't need to declare the props here, storybook passthrough the variables from the *.stories.js if needed.

Check other components *_default.story.vue files for references.

recipes/conversation_view/feed_pill/index.js
```diff
@@ -0,0 +1,2 @@
+export { default as DtRecipeFeedItemPill } from './feed_item_pill.vue';
+export { } from './feed_item_pill_constants';
```
juliodialpad: it's safe to remove the constants if none of them exists.

juliodialpad (changes requested):
Sorry for the late response.

Good work! the component looks amazing, a couple comments.

The pill is missing `cursor: pointer` and the animation feels a bit rushy.

recipes/conversation_view/feed_pill/feed_item_pill.test.js
juliodialpad: you're missing the await and maybe updating the childWrappers might be needed.

juliodialpad:
> Do I need to do anything here for screen readers? I'm assuming the consumer should pass in screen reader compatible content via slots

You just need to provide the prop so users are able to pass in the translated text and make sure the screen reader works as expected.

recipes/conversation_view/feed_pill/feed_item_pill.vue
```diff
@@ -0,0 +1,147 @@
...
+
+<script>
+import { DtIcon, DtItemLayout } from '@/index';
+
+export default {
+  name: 'DtRecipeFeedItemPill',
+
+  components: { DtItemLayout, DtIcon },
+
+  mixins: [],
```
braddialpad: remove

recipes/conversation_view/feed_pill/feed_item_pill.vue
```diff
@@ -0,0 +1,147 @@
+<template>
+  <div :class="[wrapperClass, expanded ? 'dt-feed-item--expanded-pill' : 'd-bar-pill']">
+    <div :class="['d-bgc-secondary', expanded ? 'dt-feed-item--expanded-pill' : 'd-bar-pill']">
+      <div class="d-p8">
+        <button
```
braddialpad: Is there really supposed to be no styling on hover? I notice there also isn't in the prototype you posted but it kinda feels wrong to me... verify with design?

recipes/conversation_view/feed_pill/feed_item_pill.stories.js
```diff
@@ -0,0 +1,114 @@
+import { action } from '@storybook/addon-actions';
+import { createTemplateFromVueFile } from '@/common/storybook_utils';
+import DtRecipeFeedItemPill from './feed_item_pill.vue';
+import DtRecipeFeedItemPillDefaultTemplate from './feed_item_pill_default.story.vue';
+
+// Default Prop Values
+export const argsData = {
+  onClicked: action('clicked'),
```
braddialpad: This is not being output from your default.story.vue

recipes/conversation_view/feed_pill/feed_item_pill.vue
```diff
@@ -0,0 +1,147 @@
...
+      default: () => '',
+    },
+
+    /**
+     * Sets whether the pill can be toggled
+     */
+    toggleable: {
+      type: Boolean,
+      default: () => true,
+    },
```
braddialpad: be a bit more clear in the documentation about what this does. It confused me a bit between this and expanded. Toggleable has no involvement with the actual expanding / collapsing functionality right? it simply changes the visual icon?

recipes/conversation_view/feed_pill/feed_item_pill.vue
```diff
@@ -0,0 +1,147 @@
...
+      default: () => '',
+    },
+
+    /**
+     * Shows/hides the expanded portion
+     */
+    expanded: {
+      type: Boolean,
+      default: () => false,
+    },
```
braddialpad: There is currently no "automatically expand on click" functionality right? It has to be controlled by the consumer capturing the click event and setting expanded? Should we have the default behaviour be to expand on click and then implement .sync functionality into expanded?

recipes/conversation_view/feed_pill/feed_item_pill.vue
```diff
@@ -0,0 +1,147 @@
...
+      default: () => '',
+    },
+
+    /**
+     * Additional styling for the pill
+     */
+    buttonClass: {
+      type: String,
+      default: () => '',
+    },
```
braddialpad: There are many ways to pass classes in vue and we try to support everything vue supports, so string, object, array. You can just copy the code from another component like this for example:

```
contentClass: {
  type: [String, Array, Object],
  default: '',
},
```

recipes/conversation_view/feed_pill/feed_item_pill.vue
braddialpad: or maybe even better, when toggleable is true it would automatically expand / collapse on click but when false it does nothing.

braddialpad (changes requested):

recipes/conversation_view/feed_pill/feed_item_pill.vue
AlexLamDialpad: confirmed lol

braddialpad: okay haha

recipes/conversation_view/feed_pill/feed_item_pill.test.js
```diff
@@ -0,0 +1,120 @@
...
+  const baseProvide = {};
+
+  let mockProps = {};
+  let mockAttrs = {};
+  let mockSlots = {};
+  let mockProvide = {};
+
+  const updateWrapper = () => {
+    wrapper = mount(DtRecipeFeedItemPill, {
+      propsData: { ...baseProps, ...mockProps },
```
AlexLamDialpad, line 31: props didn't work here, used propsData instead

juliodialpad: That's right, you should use `propsData` for Vue 2 and `props` for Vue 3 version.

recipes/conversation_view/feed_pill/feed_item_pill.test.js
```diff
@@ -1,133 +1,117 @@
...
-    /*
-     * Test(s) to ensure that the component is correctly rendering
-     */
-
     describe('Default render', function () {
       it('should exist', () => {
         expect(wrapper.exists()).toBe(true);
       });
       it('should render a feed item pill', () => {
+        console.log(wrapper.html());
```
juliodialpad: remove console log

recipes/conversation_view/feed_pill/feed_item_pill.test.js
```diff
@@ -1,133 +1,117 @@
...
   describe('Presentation Tests', function () {
-    /*
-     * Test(s) to ensure that the component is correctly rendering
-     */
-
     describe('Default render', function () {
       it('should exist', () => {
         expect(wrapper.exists()).toBe(true);
       });
       it('should render a feed item pill', () => {
```
juliodialpad, line 59: Add a break line between each `it` sentence please.

recipes/conversation_view/feed_pill/feed_item_pill.test.js
```diff
@@ -1,133 +1,117 @@
...
+        beforeAll(() => {
+          mockProps = {
+            toggleable: false,
+          };
+        });
+
+        it('should exist', () => {
+          expect(wrapper.exists()).toBe(true);
+        });
+        it('should not respond to clicks', () => {
```
juliodialpad, line 96: break line between `it` sentences

recipes/conversation_view/feed_pill/feed_item_pill.vue
```diff
@@ -1,29 +1,29 @@
...
-          @click="$emit('click')"
+          @click="onClick"
         >
           <dt-item-layout class="d-w100p d-p8">
             {{ title }}
             <template slot="left">
               <dt-icon
                 v-if="toggleable && hover"
-                class="dt-feed-icon"
+                class="dt-feed-pill--icon"
```
juliodialpad: I think 

```suggestion
                class="dt-feed-item--pill-icon"
```

would be the right naming

recipes/conversation_view/feed_pill/feed_item_pill.vue
```diff
@@ -1,29 +1,29 @@
...
-                :name="expanded ? 'chevron-down' : 'chevron-right'"
+                :name="hoverIcon"
               />
               <dt-icon
                 v-else
-                class="dt-feed-icon"
+                class="dt-feed-pill--icon"
                 data-qa="dt-feed-item-icon"
                 :name="iconName"
                 size="300"
```
juliodialpad: Can this be replaced with a single <dt-icon /> computed props to remove redundant code?

recipes/conversation_view/feed_pill/feed_item_pill.vue
```diff
@@ -78,65 +76,80 @@ export default {
...
     /**
      * Aria label for feed pill
      */
     ariaLabel: {
       type: String,
-      default: () => '',
+      default: null,
+      validator: (val) => {
+        return !!val;
+      },
```
juliodialpad:
```suggestion
      required: true,
```

would be simpler

recipes/conversation_view/feed_pill/feed_item_pill.vue
```diff
@@ -78,65 +76,80 @@ export default {
...
      */
     toggleable: {
       type: Boolean,
       default: () => true,
     },
   },
 
-  emits: [
-    'click',
-  ],
```
juliodialpad: Isn't this event required for the component? It wouldn't hurt to have this event emitted and it could potentially be used for something else by the consumer.

recipes/conversation_view/feed_pill/feed_item_pill.vue
```diff
@@ -78,65 +76,80 @@ export default {
...
+    toggleableClass () {
+      return this.toggleable ? 'd-c-pointer' : '';
+    },
+  },
+
+  methods: {
+    onClick () {
+      if (this.toggleable) {
+        this.expanded = !this.expanded;
+      }
```
juliodialpad:
```suggestion
    onClick () {
      if (!this.toggleable) return;
      
      this.expanded = !this.expanded;
```

recipes/conversation_view/feed_pill/feed_item_pill.vue
```diff
@@ -78,65 +76,80 @@ export default {
...
+    },
+  },
 };
 </script>
 
-<style lang="less">
-  .dt-feed-item--expanded-pill {
-    border-radius: 5rem
+<style scoped>
```
juliodialpad:
```suggestion
<style lang="less" scoped>
```

recipes/conversation_view/feed_pill/feed_item_pill.test.js
```diff
@@ -1,133 +1,117 @@
...
+    });
+
+    // Test not working
+    describe.skip('Hover Feed Item Pill event', function () {
       it('should show a different icon', async () => {
+        await feedItemPill.trigger('focus');
+        _setChildWrappers();
         expect(icon.exists()).toBe(false);
         expect(hoverIcon.exists()).toBe(true);
       });
```
juliodialpad: Maybe you can emit an event on focus and check if the event has been fired?

IDK what else to suggest here, any ideas? @braddialpad @iropolo

juliodialpad (changes requested):
This looks much cleaner, thanks! still having some minor issues.

recipes/conversation_view/feed_pill/feed_item_pill.vue
AlexLamDialpad: Should be no cases where the pill does anything aside from expand

recipes/conversation_view/feed_pill/feed_item_pill.test.js
braddialpad: I've always had big problems with jsdom and setting focus.. 1 sec taking a look

recipes/conversation_view/feed_pill/feed_item_pill.test.js
```diff
@@ -0,0 +1,121 @@
...
+          expect(icon.attributes('data-name')).toBe(MOCK_ICON_NAME);
+          expect(wrapper.find('[data-qa="content-element"]').exists()).toBe(false);
+        });
+      });
+    });
+
+    // Test not working
+    describe.skip('Hover Feed Item Pill event', function () {
+      it('should show a different icon', async () => {
+        await feedItemPill.trigger('focus');
```
braddialpad: This is not working because the event is `focusin` rather than `focus`

braddialpad (approved):

braddialpad:
you can ignore the broken accessibility test it was broken by a different commit

recipes/conversation_view/feed_pill/feed_item_pill.test.js
AlexLamDialpad: 🤦 yeah that works now

juliodialpad (approved):
Nice work, thank you!

juliodialpad:
Don't forgot to create Vue 3 version of this

juliodialpad:
The accessibility broken tests are actually coming from this component:
https://vue.dialpad.design/deploy-previews/pr-1135/?path=/story/recipes-conversation-view-feed-item-pill--default

image

dialpad-marshall:
Looking pretty good!

* Call Summary button
    * use the [round button style](https://dialpad.design/components/button.html#icon-only)
    * Icon size should be `d-icon—size-300`
* Border colors
    * See writing below
    * Default border should be `—dt-color-border-default`
* Are the responsive variations on the way?
* Icon hover transition should be quicker, maybe half the duration it is now
* Is it possible to add a transition to the expand/collapse action as demonstrated in the prototype?

**Regarding the gradient border.** When expanded there is some wobble in the border width. This is likely because we are using a background div attached to the main element to create the border. I suggest using multiple backgrounds on the main element. Check [this codepen](https://codepen.io/rappin_wizard/pen/PoXNwKe). [Browser support](https://caniuse.com/?search=multiple%20backgrounds) is decent. And we can fall back to a solid purple or magenta border if we need to.

iropolo (changes requested):
I will review `test` section, since Julio and Brad already taken a look on this.

Please read our [TEST CONTRIBUTION GUIDELINE](https://github.com/dialpad/dialtone-vue/blob/staging/.github/TEST_CONTRIBUTING/TEST_GUIDELINE.md).

Some notes:
There is no more `setChildWrapper`, now all wrappers are defined on our `updateWrapper` function.

We no longer  use unnecessary `//comments`, just if it is needed, so please remove `// Wrappers` `// Helpers` etc..

Now we have a style structure, please check [here](https://github.com/dialpad/dialtone-vue/blob/staging/.github/TEST_CONTRIBUTING/TEST_GUIDELINE.md#styling-and-a-few-notes).

Since this is a vue2 version tests. please check [here](https://github.com/dialpad/dialtone-vue/blob/staging/.github/TEST_CONTRIBUTING/TEST_GUIDELINE.md#styling-and-a-few-notes).

Here are some examples:
https://github.com/dialpad/dialtone-vue/pull/1145
https://github.com/dialpad/dialtone-vue/pull/1141
https://github.com/dialpad/dialtone-vue/pull/1127

We are currently running a `test migrations` to move to this new template.
Thanks!

dialpad-marshall:
Looking good! I assume the rest of the changes are coming (if not here, no worries) but I will record here just incase:

* Call Summary button
  - use the [round button style](https://dialpad.design/components/button.html#icon-only)
  - Icon size should be d-icon—size-300
* Border colors
  - Default border should be `--dt-color-border-default`
* Is it possible to add a transition to the expand/collapse action as demonstrated in the prototype?

Thanks!

AlexLamDialpad:
@juliodialpad @iropolo Hey I made a few changes as per design requested since the approval, biggest changes:
- border colors are now pre-set just like https://vue.dialpad.design/?path=/story/recipes-leftbar-callbox--default
- Using DtCollapsible instead of v-if because design wants there to be a collapse/expanding animation (speaking of which is the animation actually working?)
Thanks!

recipes/conversation_view/feed_pill/feed_item_pill_constants.js
```diff
@@ -0,0 +1,9 @@
+export const BORDER_COLORS = {
+  default: 'dt-feed-item-pill--border-default',
+  ai: 'dt-feed-item-pill--border-ai',
+  critical: 'dt-feed-item-pill--border-critical',
+};
+
+export default {
+  BORDER_COLORS,
+};
```
iropolo: I would rename this to be more specifc to something like
`PILL_BORDER_COLORS`
Im doing this suggestions because there is also a `BORDER_COLORS` constant from callbox.

Also it is not necessary to export it in two ways. I suggest to only export it as 
`export const PILL_BORDER_COLORS = {`
and remove the `export default {`

I will extend this suggestions to @juliodialpad since callbox it is a new recently recipe.
`recipes/leftbar/callbox/callbox_constants.js`

recipes/conversation_view/feed_pill/feed_item_pill_default.story.vue
```diff
@@ -0,0 +1,55 @@
...
+    <template #right>
+      <div class="d-pr16">
+        <dt-button
+          aria-label="Open external link"
+          kind="muted"
+          importance="clear"
+          :circle="true"
+          @click.stop=""
+        >
+          <template slot="icon">
```
iropolo: This syntax `slot="icon"` is deprecated.

The recommended way to do it in shorthand `#icon`

Doc: https://v2.vuejs.org/v2/guide/components-slots#Named-Slots-with-the-slot-Attribute

Please replace all `slot="...` to this new syntax. 
Found some references in this file and `feed_item_pill.vue`.

recipes/conversation_view/feed_pill/feed_item_pill.test.js
```diff
@@ -0,0 +1,116 @@
...
+        });
+      });
+    });
+
+    describe('Hover Feed Item Pill event', function () {
+      beforeEach(async () => {
+        await feedItemPill.trigger('focusin');
+        icon = wrapper.find('[data-qa="dt-feed-item-pill-icon"]');
+      });
+      it('should show a different icon', () => {
```
iropolo: Break line here

iropolo (changes requested):
Thanks for the update, looks clean!

Thanks for your update in the `.test`. There is the plan to create lint rules to make it easy for you guys to contribute. :)

Please answer/close requests with information on what you did or what you think lets avoid to have unanswered/open requests, then it will be easy for me or another reviewer to follow.

recipes/conversation_view/feed_pill/feed_item_pill_constants.js
juliodialpad: > I would rename this to be more specifc to something like `PILL_BORDER_COLORS` Im doing this suggestions because there is also a `BORDER_COLORS` constant from callbox.
> 

You're right, I'll rename my constants too, to avoid conflicts.

> Also it is not necessary to export it in two ways. I suggest to only export it as `export const PILL_BORDER_COLORS = {` and remove the `export default {`
> 

This is the standard we're following, it's a good practice so we can import the constants in two ways

## With default export
```js
import pillConstants from './feed_item_pill_constants' 
```

## Named export
```js
import { BORDER_COLORS } from './feed_item_pill_constants' 
```

This way we don't enforce a single way for clients to import our constants.

AlexLamDialpad: So no change here right?

juliodialpad: > So no change here right?

Sorry for not being clear on my comment.

Please rename the variables, prepend the component name

`BORDER_COLORS` -> `FEED_ITEM_PILL_BORDER_COLORS`

iropolo:
Seems like there is a failing test in a11y check

image

Could you track it is related with the PR please?

AlexLamDialpad:
> Details

@iropolo 
It's because we're nesting button on top of button, is there a way to acknowledge this and disable the failure? 
'Ensures interactive controls are not nested as they are not always announced by screen readers or can cause focus problems for assistive technologies'

iropolo:
> > Details
> 
> @iropolo It's because we're nesting button on top of button, is there a way to acknowledge this and disable the failure? 'Ensures interactive controls are not nested as they are not always announced by screen readers or can cause focus problems for assistive technologies'

image

This one from callbox I will ping @juliodialpad on this.
I removed the whole feed-item-pill folder and this callbox accesibility issue persist.

@AlexLamDialpad as you said, seems like is the `dt-button` of the `collapsible.vue` and the button provided by `feed_item_pill_default.story.vue` in `#right` slot cause the issue...

Let me see what I can do

juliodialpad:
> image
>
> This one from callbox I will ping @juliodialpad on this. 

Already fixed that, just need to merge master I think

iropolo:
@AlexLamDialpad please add this
`parameters: { a11y: { disable: true } },` on each of `feed_item_pill.stories.js`
`Default` `AiBorderGradient` and `NoToggleVariant` 
and push it.

This will skip the accessibility test. 
Update the branch, check `Accessibility Tests / a11y_tests (pull_request)` actions works and then merge.

I will aprove this PR because this is not a blocker.
The accesibility issue comes from how its being used in the default stories of the component, not from inside the component.

iropolo (approved):

PR #1139, opened 2023-08-24
refactor(callbox): visuals

juliodialpad:
# Refactor(Callbox): Visuals

- [ ] Fix
- [ ] Feature
- [x] Refactoring
- [ ] Documentation

## :book: Description

- Refactored visuals to match designs on: https://www.figma.com/file/oTY56LgQha8ljb4c8w4GtO/DT-Shared%3A-App-Components?node-id=9240%3A7131&t=qTyNqMCFoyQN41f4-0

## :bulb: Context

Component wasn't visually polished.

## :pencil: Checklist

- [ ] I have added tests

## :crystal_ball: Next Steps

- Write documentation and variants
- Write tests

image
image


recipes/leftbar/callbox/callbox_default.story.vue
```diff
@@ -66,39 +68,11 @@
...
-            </template>
-            Stop
-          </dt-button>
-        </template>
-      </dt-item-layout>
+    <template
+      v-if="bottom"
+      #bottom
+    >
+      <div v-html="bottom" />
```
braddialpad, line 47: I think it is a bit unclear just looking at the default story what is included as part of the component and what is not. For example it seems like the AI and hang up buttons are part of this component but they actually are not you can pass anything into the slot.

I think we should for the default example have a string that just says "right slot", and then under the variants story we can show more defined examples like showing the AI / hangup button and such. Maybe a few different variants from the design examples here. This might be relevant for the bottom slot too as the icon is not included.

braddialpad (commented):
Really nice work here. I just had one suggestion.

recipes/leftbar/callbox/callbox_default.story.vue
juliodialpad: That makes sense! I'll update the examples

iropolo (approved):
looks good!!

PR #1146, opened 2023-08-28
fix(avatar): dynamic icon sizes

juliodialpad:
# Fix (Avatar): Dynamic icon sizes

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Added back icon sizes according to avatar size

## :bulb: Context

We were missing the dynamic icon size according to the avatar's selected size, can't remember when we loosed this functionality.

## :pencil: Checklist

- [ ] I have added tests

image

### DIFF
--- a/components/avatar/avatar.vue
+++ b/components/avatar/avatar.vue
@@ -19,7 +19,7 @@
       <dt-icon
         v-else-if="iconName"
         :name="iconName"
-        :size="iconSize"
+        :size="iconSize || AVATAR_ICON_SIZES[size]"
         :class="[iconClass, AVATAR_KIND_MODIFIERS.icon]"
         data-qa="dt-avatar-icon"
       />
@@ -76,6 +76,7 @@ import {
   AVATAR_PRESENCE_STATES,
   AVATAR_COLORS,
   AVATAR_GROUP_VALIDATOR,
+  AVATAR_ICON_SIZES,
 } from './avatar_constants';
 import { getIconNames } from '@/common/storybook_utils.js';
 import { ICON_SIZE_MODIFIERS } from '@/components/icon/icon_constants.js';
@@ -243,8 +244,8 @@ export default {
      */
     iconSize: {
       type: String,
-      default: '500',
-      validator: (size) => Object.keys(ICON_SIZE_MODIFIERS).includes(size),
+      default: '',
+      validator: (size) => !size || Object.keys(ICON_SIZE_MODIFIERS).includes(size),
     },
 
     /**
@@ -261,6 +262,7 @@ export default {
       AVATAR_SIZE_MODIFIERS,
       AVATAR_KIND_MODIFIERS,
       AVATAR_PRESENCE_SIZE_MODIFIERS,
+      AVATAR_ICON_SIZES,
       imageLoadedSuccessfully: null,
       formattedInitials: '',
       initializing: false,
--- a/components/avatar/avatar_constants.js
+++ b/components/avatar/avatar_constants.js
@@ -26,7 +26,7 @@ export const AVATAR_PRESENCE_STATES = {
 };
 
 export const AVATAR_ICON_SIZES = {
-  xs: undefined,
+  xs: '100',
   sm: '200',
   md: '300',
   lg: '500',


braddialpad:
> can't remember when we loosed this functionality.

I'm guessing during the dialtone 8 migration. Hmm was there some intentional reason we did this? Maybe it was supposed to be all handled via CSS?

braddialpad:
~~Seems like the avatar is set to size to 100% of the canvas size. https://github.com/dialpad/dialtone/blob/3231f28b2cbe7375afc587ad08ab8a89194fdfec/lib/build/less/components/avatar.less#L49~~

~~Are there situations where we don't want that?~~

NM this is for images only

juliodialpad:
> > can't remember when we loosed this functionality.
> 
> I'm guessing during the dialtone 8 migration. Hmm was there some intentional reason we did this? Maybe it was supposed to be all handled via CSS?



> > can't remember when we loosed this functionality.
> 
> I'm guessing during the dialtone 8 migration. Hmm was there some intentional reason we did this? Maybe it was supposed to be all handled via CSS?

Seems like longer ago, can't find the commit where we removed the sizing from the icon, or maybe we never used that constant really?

juliodialpad:
> Seems like the avatar is set to size to 100% of the canvas size. https://github.com/dialpad/dialtone/blob/3231f28b2cbe7375afc587ad08ab8a89194fdfec/lib/build/less/components/avatar.less#L49
> 
> Are there situations where we don't want that?

Don't really know, just thought It'd look better with its corresponding icon sizes and realized that constant was unused

Before
image

After
image

braddialpad:
> Don't really know, just thought It'd look better with its corresponding icon sizes and realized that constant was unused

Yeah that definitely looks bad.

I think maybe we should set this on the dialtone (CSS/LESS) side though. perhaps we can still keep this prop here to override if necessary

juliodialpad:
> > Don't really know, just thought It'd look better with its corresponding icon sizes and realized that constant was unused
> 
> Yeah that definitely looks bad.
> 
> I think maybe we should set this on the dialtone (CSS/LESS) side though. perhaps we can still keep this prop here to override if necessary

I think this is the right way to do it, as the icon itself handles the sizing, we're just mapping the icon sizes to the avatar sizes.

If we do this with CSS, maybe we we'll have issues overriding the size with the `iconSize` prop, don't you think?

braddialpad:
> I think this is the right way to do it, as the icon itself handles the sizing, we're just mapping the icon sizes to the avatar sizes.
> 
> If we do this with CSS, maybe we we'll have issues overriding the size with the `iconSize` prop, don't you think?

Yeah true because we pass into the dt-icon component which always has a size... I think this could be done via css but might require some restructuring. Fine to do this for now, but we do need to update the examples on dialpad.design to have icon sizes applied then because right now it looks like this:

Screenshot 2023-08-28 at 10 38 55 AM

juliodialpad:
> > I think this is the right way to do it, as the icon itself handles the sizing, we're just mapping the icon sizes to the avatar sizes.
> > If we do this with CSS, maybe we we'll have issues overriding the size with the `iconSize` prop, don't you think?
> 
> Yeah true because we pass into the dt-icon component which always has a size... I think this could be done via css but might require some restructuring. Fine to do this for now, but we do need to update the examples on dialpad.design to have icon sizes applied then because right now it looks like this:
> 
> Screenshot 2023-08-28 at 10 38 55 AM

Yeah, I think once this is release and updated on dialtone, it should be updated automatically, if not, I'll fix it anyways while updating dialtone-vue on dialtone.

braddialpad:
> Yeah, I think once this is release and updated on dialtone, it should be updated automatically, if not, I'll fix it anyways while updating dialtone-vue on dialtone.

Oh is this actually displaying via the vue component? In that case we should still update the code example to apply an icon size, because if you were to copy the html that is there it would not match the component example.

though I guess it doesn't really show the icon code there.. only the wrapping avatar classes.

braddialpad (approved):

juliodialpad:
> > Yeah, I think once this is release and updated on dialtone, it should be updated automatically, if not, I'll fix it anyways while updating dialtone-vue on dialtone.
> 
> Oh is this actually displaying via the vue component? In that case we should still update the code example to apply an icon size, because if you were to copy the html that is there it would not match the component example.
> 
> though I guess it doesn't really show the icon code there.. only the wrapping avatar classes.

Will do the corresponding updates on dialtone if needed.

PR #1150, opened 2023-08-29
docs(callbox): variants and docs

juliodialpad:
# Docs (Callbox): Variants and docs

- [ ] Fix
- [ ] Feature
- [ ] Refactoring
- [x] Documentation

## :book: Description

Added variants and updated documentation

## :bulb: Context

Variants for more examples and documentation were missing.

## :pencil: Checklist

- [ ] I have added tests

## :crystal_ball: Next Steps

- Write tests.

image


fede-dp (approved):

iropolo (approved):
Looks good!

Would be cool if we can resize the image to match the same as the available in video.
Also use some compressor like [`tinypng`](https://tinypng.com/).

So we can decrease 338kb to maybe 40kb.

juliodialpad:
> Would be cool if we can resize the image to match the same as the available in video. 
I don't get this. The image is using 

```CSS
width: 100%;
height: 100%;
object-fit: cover;
```

So it is resized to match the available video space. What do you mean?

>Also use some compressor like [`tinypng`](https://tinypng.com/). So we can decrease 338kb to maybe 40kb.

I don't think this is necessary as it is only for documentation it's not going to affect build size.

I tried another compressed image from figma but it looked really awful, this was the one I found to be best mid-sized good quality image I could get from there.

juliodialpad:
Renamed the constants to avoid conflicts

PR #1151, opened 2023-08-29
docs(callbox): variants and docs - vue3

juliodialpad:
# Vue 3 version of #1150


recipes/leftbar/callbox/callbox.stories.js
```diff
@@ -44,6 +45,12 @@ const DefaultTemplate = (args) => createTemplateFromVueFile(
   DtRecipeCallboxDefaultTemplate,
 );
 
+const VariantsTemplate = (args, { argTypes }) => createTemplateFromVueFile(
+  args,
+  argTypes,
```
iropolo: Seems like it is not working cos this `argTypes`, remove it and it will works.

iropolo (changes requested):
Variants not working

recipes/leftbar/callbox/callbox.stories.js
juliodialpad: Thanks for catching that! forgot to remove it while migrating to vue3

juliodialpad:
Renamed the constants to avoid conflicts

iropolo (approved):
Looks goods, thanks for the update

PR #1156, opened 2023-09-01
test(combobox, datepicker, dropdown): new test template

iropolo:
# Test migrations

- [x] Refactoring

## :book: Description


.github/TEST_CONTRIBUTING/TEST_GUIDELINE.md
```diff
@@ -403,6 +403,30 @@ describe('Extendability Tests', () => {
...
+afterEach(() => {
+    mockAttrs = {};
+    mockProps = {};
+    ...
+    vi.restoreAllMocks();
+  });
+
+```
+
+
```
juliodialpad: Nice, thanks for adding this.

components/combobox/combobox.test.js
```diff
@@ -11,158 +15,141 @@ const basePropsData = {
...
-        _setChildWrappers();
+      beforeEach(() => {
+        mockScopedSlots = { input: '<input v-bind="props.inputProps" />' };
+
+        updateWrapper();
       });
 
       it('should provide proper label prop to input element', () => {
-        expect(input.attributes('label')).toEqual(basePropsData.label);
+        expect(input.attributes('label')).toEqual(baseProps.label);
```
juliodialpad, line 106: I'm confused, should we do this `baseProps.property` or create a `const MOCK_PROP = 'Some text';` and using that across the test?

components/combobox/combobox.test.js
```diff
@@ -175,55 +162,48 @@ describe('DtCombobox Tests', () => {
...
+      beforeEach(() => {
+        mockScopedSlots = { input: '<input v-bind="props.inputProps" />' };
+
+        updateWrapper();
       });
 
       describe('When list is not expanded', () => {
-        beforeEach(async () => {
+        it('aria-expanded should be "false"', async () => {
           await wrapper.setProps({ showList: false });
```
juliodialpad, line 186: This should be changed to:

```suggestion
          mockProps = { showList: false };
          
          updateWrapper();
```

and remove the async

components/combobox/combobox.test.js
```diff
@@ -175,55 +162,48 @@ describe('DtCombobox Tests', () => {
...
 
-        it('aria-expanded should be "false"', () => {
           expect(input.attributes('aria-expanded') === 'false').toBe(true);
         });
       });
 
       describe('When list is expanded', () => {
-        beforeEach(async () => {
+        it('aria-expanded should be "true"', async () => {
           await wrapper.setProps({ showList: true });
```
juliodialpad, line 194: same here, change to mockProps

components/combobox/combobox.test.js
```diff
@@ -175,55 +162,48 @@ describe('DtCombobox Tests', () => {
...
-        it('aria-expanded should be "true"', () => {
+          input = wrapper.find('input');
+
           expect(input.attributes('aria-expanded') === 'true').toBe(true);
         });
 
         describe('When list is loading', () => {
-          beforeEach(async () => {
+          it('aria-busy should be "true"', async () => {
             await wrapper.setProps({ loading: true });
```
juliodialpad, line 203: change to mockProps

components/combobox/combobox.test.js
```diff
@@ -175,55 +162,48 @@ describe('DtCombobox Tests', () => {
...
-            _setChildWrappers();
-          });
 
-          it('aria-busy should be "true"', () => {
+            listWrapper = wrapper.find('[data-qa="dt-combobox-list-wrapper"]');
```
juliodialpad, line 205: remove this and call `updateWrapper()` instead

components/dropdown/dropdown.test.js
```diff
@@ -81,148 +62,132 @@ describe('DtDropdown Tests', () => {
...
+    it('should render the component', () => {
+      expect(wrapper.exists()).toBe(true);
     });
 
-    it(
-      'should render the component',
-      () => { expect(wrapper.exists()).toBe(true); },
-    );
+    it('should not render the visually hidden close button', () => {
+      const buttonExists = wrapper.findComponent(SrOnlyCloseButton).exists();
```
juliodialpad: Something like this would be better, or if you're reusing the component, add it to the `updateWrapper` function.

```suggestion
      const button = wrapper.findComponent(SrOnlyCloseButton);
      
      expect(button.exists()).toBe(false);
```

components/dropdown/dropdown.test.js
```diff
@@ -81,148 +62,132 @@ describe('DtDropdown Tests', () => {
...
+        mockProps = { visuallyHiddenClose: true };
+
+        updateWrapper();
       });
 
       it('should contain a visually hidden close button', () => {
-        itBehavesLikeVisuallyHiddenCloseButtonExists(wrapper);
+        const buttonExists = wrapper.findComponent(SrOnlyCloseButton).exists();
+
+        expect(buttonExists).toBe(true);
```
juliodialpad: Reuse this by adding it to `updateWrapper`

components/dropdown/dropdown.test.js
```diff
@@ -81,148 +62,132 @@ describe('DtDropdown Tests', () => {
...
-        () => { expect(wrapper.vm.highlightIndex).toBe(-1); },
-      );
+        expect(wrapper.vm.highlightIndex).toBe(-1);
+      });
     });
 
     describe('When sr-only close button is enabled and activated', () => {
-      beforeEach(async () => {
+      it('should close the dropdown', async () => {
         await wrapper.setProps({ visuallyHiddenClose: true });
```
juliodialpad, line 185: change to mockProps

juliodialpad (changes requested):
Good work here! just a couple minor improvements that can be done quickly.

components/combobox/combobox.test.js
iropolo: IMO Is the same.

You can use `MOCK_PROP` and use it inside `baseProps` and then match `MOCK_PROP` in assertion or directly use `baseProps`.

I would preffer to use directly `baseProps`, because there is no need to use `MOCK_`.

But it depends on the test, maybe in some case you will prefer to `MOCK_` because you have a bunch of assertions.

The important thing is avoid use the same direct value on multiple assertions.

If you will use direct values, make sure it will be the only one. If you need to use it in multiple tests, it will be better to have a `MOCK_` or some inmutable variable.

Let me show an example:

It is okay to do:
```
it(...){
expect(...).toBe(25);
}
it(...){
expect(...).toBe(26);
}
it(...){
expect(...).toBe(27);
}
```

It is NOT okay to do:
```
it(...){
expect(...).toBe(25);
}
it(...){
expect(...).toBe(25);
}
it(...){
expect(...).toBe(25);
}
```
In this case you should use `MOCK_` or some `base` prop.
It is not necessary to be strict on this, to avoid multiple non-reusable `MOCK_` variables.

components/combobox/combobox.test.js
iropolo: This way of do it is valid and is specified on [TEST_GUIDELINE](https://github.com/dialpad/dialtone-vue/blob/staging/.github/TEST_CONTRIBUTING/TEST_GUIDELINE.md).

`You can utilize mockProps to set environment configuration to the mounted component, and use .setProps method if you want to trigger some component behaviour.`

The difference on doing it using `updateWrapper();` or `.setProps()` is in the first one you will re-mount the component, on the second one it will react to a change.

Your suggestion on this is valid.

As you will see on the change, I removed the `beforeEach` (because it is only one test) and just reuse the `setProps` because there is no need to update it.

components/combobox/combobox.test.js
iropolo: Same as in https://github.com/dialpad/dialtone-vue/pull/1156#discussion_r1315199383

`If you are testing some conditional render of the component, you can create a wrapper inside the test to work on it.`


This way of doing it is also valid.

With `.setProps()` you will trigger a wrapper change, then you will update the wrapper to assert it.

components/dropdown/dropdown.test.js
iropolo: Definetly, but Im just doing refactor test with the new test template.

iropolo:
I don't see it necessary to update the PR. 

@juliodialpad if you think it is necessary to update doc to be more clear please do your suggestion and I will update it.

Also if it is ncessary to discuss some point we can do it and if you want to have another POV on this we can bring it to discuss with @braddialpad later when he is back.

Your sugestions on this PR are valid ones. 

The difference on why Im writing it in one way or another it depends on the original test code. For example if the original test code use `.setProps` I will probably keep it.

The same if the user create a `const wrapper` or update it in the test.

To do a quick refactor Im doing this hacks. Also because those are valid way of tests.

The idea of this template is make the test readable and flexible for the developer. The use of `updateWrapper` `.setProps` `baseProps` or `MOCK_` are valid options.

The reason on why Im not doing the test more optimal or reuse some things is because Im not doing a `test review`, this is a refactor task.

Thanks for the review and pointing it out this things, It is probably need to update the doc with this cases.

components/combobox/combobox.test.js
juliodialpad: You're right, I didn't see that on the guideline, I'd be good to make that more visible.

components/dropdown/dropdown.test.js
juliodialpad: It's not that you're refactoring too much to fix that, I'd be cool if you can fix some stuff that you find even if it's not what you're currently doing.

That help us to have less tech-debt.

iropolo: Sure, tomorrow will retake this tests migrations and I will do a couple of improvements

components/dropdown/dropdown.test.js
iropolo: Updated!

components/dropdown/dropdown.test.js
iropolo: Updated!

iropolo:
Updated doc and test improvements.

.github/TEST_CONTRIBUTING/TEST_GUIDELINE.md
```diff
@@ -344,20 +342,53 @@ describe('When disabled', () => {
...
+    const updateWrapper = () => {
+      ...
+      ...
+      decorativeSpan = wrapper.find('.d-badge__decorative');
+    }
+    ...
+  describe(...)
+     ...
+    {
+     it(..., async () => {
```
juliodialpad: for this example, this async is not needed here, right?

.github/TEST_CONTRIBUTING/TEST_GUIDELINE.md
```diff
@@ -344,20 +342,53 @@ describe('When disabled', () => {
...
+         expect(decorativeSpan.exists()).toBeTruthy();
+     });
+     
+     it(..., async () => {
+        await wrapper.setProps({disabledSection: false});
+      
+        const decorativeSection = wrapper.find('.d-badge__decorative-section');
+      
+        expect(decorativeSection.exists()).toBeTruthy();
+      });
```
juliodialpad, line 383: Nice, this examples are a lot more clear

.github/TEST_CONTRIBUTING/TEST_GUIDELINE.md
```diff
@@ -403,7 +434,7 @@ describe('Extendability Tests', () => {
 });
 ```
 
-- If you are testing with `vi.fn()` mocks it is necessary to restore it after each test.
+- If you are testing with `vi.fn()` mocks it is necessary/recommend to restore it after each test.
```
juliodialpad:
```suggestion
- If you are testing with `vi.fn()` mocks it is necessary/recommended to restore it after each test.
```

components/dropdown/dropdown.test.js
```diff
@@ -36,6 +36,7 @@ describe('DtDropdown Tests', () => {
   let wrapper;
   let anchorElement;
   let listWrapper;
+  let buttonExists;
```
juliodialpad:
```suggestion
  let closeButton;
```

`buttonExists` doesn't make sense.

components/dropdown/dropdown.test.js
```diff
@@ -52,6 +53,7 @@ describe('DtDropdown Tests', () => {
 
     anchorElement = wrapper.find('#anchor');
     listWrapper = wrapper.find('[data-qa="dt-dropdown-list-wrapper"]');
+    buttonExists = wrapper.findComponent(SrOnlyCloseButton);
```
juliodialpad:
```suggestion
    closeButton = wrapper.findComponent(SrOnlyCloseButton);
```

components/dropdown/dropdown.test.js
```diff
@@ -85,9 +87,7 @@ describe('DtDropdown Tests', () => {
     });
 
     it('should not render the visually hidden close button', () => {
-      const buttonExists = wrapper.findComponent(SrOnlyCloseButton).exists();
-
-      expect(!buttonExists).toBe(true);
+      expect(buttonExists.exists()).toBe(false);
```
juliodialpad:
```suggestion
      expect(closeButton.exists()).toBe(false);
```

components/dropdown/dropdown.test.js
```diff
@@ -112,9 +112,7 @@ describe('DtDropdown Tests', () => {
       });
 
       it('should contain a visually hidden close button', () => {
-        const buttonExists = wrapper.findComponent(SrOnlyCloseButton).exists();
-
-        expect(buttonExists).toBe(true);
+        expect(buttonExists.exists()).toBe(true);
```
juliodialpad:
```suggestion
        expect(closeButton.exists()).toBe(true);
```

juliodialpad (approved):
Documentation looks way more clear, just rename the `buttonExists` variable to a more accurate naming before merging, thanks!

components/dropdown/dropdown.test.js
iropolo: 👍🏼 , will update the name to `closeButton`

PR #1162, opened 2023-09-07
test(emoji, emoji-text-wrapper, image-viewer): new test template 

iropolo:
# Test migrations

- [x] Refactoring

## :book: Description


components/emoji/emoji.test.js
```diff
@@ -1,183 +1,135 @@
...
-const basePropsData = {};
+const MOCK_EXPECTED_SMILE_SRC = 'https://mockstorage.com/emojis/1f604.svg';
+const MOCK_EXPECTED_SMILE_SRC_SMALL = 'https://mockstorage.com/emojis/1f604.png';
+const MOCK_EXPECTED_LAUGHING_SRC = 'https://mockstorage.com/emojis/1f606.svg';
+const MOCK_EXPECTED_POINT_UP_LIGHT = 'https://mockstorage.com/emojis/261d-1f3fb.svg';
+const MOCK_EXPECTED_SHIP_IT = 'https://mockstorage.com/emojis/shipit.png';
 
-describe('DtEmoji Tests', () => {
-  let testContext;
+const baseProps = { code: '' };
```
iropolo, line 16: Sth I need to point here.

The `code` prop in `emoji.vue` the default value is null, if we provide a null value its broke the component.

In our method: 
```
getEmojiData () {
      this.emojiData = codeToEmojiData(this.code);
    },
```
We have this `codeToEmojiData(this.code);` which is doing this:
`if (code.startsWith(':') && code.endsWith(':')) {`

So if you send a null , this `code.startsWith(':')` brokes the component, so what we can do here is a few options:

1. change the default value from `null` to empty string `''`
2. change it as `required` prop 
3. or maybe we can safe check this and return an empty string...

As you can see in the test I provide as a `baseProp` a `code` with empty string. I should not do this because Im bypassing the error.

braddialpad: hmm I'm thinking it would make sense for code to be a required prop.

iropolo: I was thinking required prop is my preferred option but this could be a breaking change...
Maybe `null` to empty string `''` would work also...

How we can move forward with this as `required` prop?
Should I create another pr with this commit as `refactor` and `BREAKING CHANGE`?

braddialpad (approved):
Looks pretty good. Nice reduction in number of lines.

components/emoji/emoji.test.js
braddialpad: hmm do you really think it would break anything? Would anyone really be using emoji without the code prop?

iropolo: dunno, just asking, I will update it here then

PR #1164, opened 2023-09-09
feat: add `avatar-click` event in contact info

tonydialpad:

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

This pull request adds `avatar-click` event in contact info component.

## :bulb: Context

When a user clicks on any of the avatars in the callbar when on a call with multiple other participants (i.e. a group call) it opens the “Manage participants” modal. To implement that, we need to add `avatar-click` event in contact info component.

## :pencil: Checklist

- [ ] I have added tests

## :crystal_ball: Next Steps

Make a new release for dialtone-vue

 No UI update

## :link: Sources

JIRA: https://dialpad.atlassian.net/browse/DP-71224
Vue3 PR: https://github.com/dialpad/dialtone-vue/pull/1165

### DIFF
--- a/recipes/list_items/contact_info/contact_info.vue
+++ b/recipes/list_items/contact_info/contact_info.vue
@@ -7,12 +7,13 @@
     class="dt-contact-info"
   >
     <template #left>
-      <div
+      <button
         v-if="showAvatar"
-        class="d-ps-relative"
+        class="d-ps-relative d-bgc-transparent d-baw0"
         data-qa="contact-info-left"
+        @click="avatarClick"
       >
-        <div
+        <span
           v-if="avatarList"
           class="dt-contact-info--avatars d-mrn4 d-d-flex d-fd-row"
         >
@@ -28,7 +29,7 @@
             :overlay-text="avatar.text"
             :avatar-class="[{ 'd-mln24': index > 0, 'd-bc-brand': !!avatar.halo }]"
           />
-        </div>
+        </span>
         <dt-avatar
           v-else
           :size="avatarSize"
@@ -38,7 +39,7 @@
           :seed="avatarSeed"
           :presence="presence"
         />
-      </div>
+      </button>
     </template>
     <template #default>
       <div data-qa="contact-info-header">
@@ -181,6 +182,14 @@ export default {
     avatarList: {
       type: Array,
       default: null,
+    },
+  },
+
+  emits: ['avatar-click'],
+
+  methods: {
+    avatarClick () {
+      this.$emit('avatar-click');
     },
   },
 };
--- a/recipes/list_items/contact_info/contact_info.vue
+++ b/recipes/list_items/contact_info/contact_info.vue
@@ -11,7 +11,7 @@
         v-if="showAvatar"
         class="d-ps-relative d-bgc-transparent d-baw0 d-c-pointer"
         data-qa="contact-info-left"
-        :aria-label="$i18n('avatar')"
+        :aria-label="avatarAriaLabel"
         @click="avatarClick"
       >
         <span
@@ -156,6 +156,14 @@ export default {
     avatarSize: {
       type: String,
       default: 'lg',
+    },
+
+    /**
+     * The aria label of the avatar
+     */
+    avatarAriaLabel: {
+      type: String,
+      default: '',
     },
 
     /**
--- a/recipes/list_items/contact_info/contact_info.vue
+++ b/recipes/list_items/contact_info/contact_info.vue
@@ -11,7 +11,7 @@
         v-if="showAvatar"
         class="d-ps-relative d-bgc-transparent d-baw0 d-c-pointer"
         data-qa="contact-info-left"
-        :aria-label="avatarAriaLabel"
+        :aria-labelledby="avatarAriaLabel"
         @click="avatarClick"
       >
         <span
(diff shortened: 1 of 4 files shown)


iropolo:
Cool!, could you please add the event on storybook so we can see the event emitted and also some JSDoc, please?

At this moment in the deploy preview I cannot see any event on the Actions section, and the documentation is null.

image

iropolo:
Seems like there is a11y failing test 

`FAIL recipes/list_items/contact_info/contact_info.stories.js`

recipes/list_items/contact_info/contact_info.vue
```diff
@@ -7,12 +7,13 @@
...
   >
     <template #left>
-      <div
+      <button
         v-if="showAvatar"
-        class="d-ps-relative"
+        class="d-ps-relative d-bgc-transparent d-baw0"
         data-qa="contact-info-left"
+        @click="avatarClick"
       >
```
braddialpad: Going to need an aria-label that says what this button does. Probably needs to be a prop for translation reasons.

recipes/list_items/contact_info/contact_info.vue
```diff
@@ -7,12 +7,13 @@
     class="dt-contact-info"
   >
     <template #left>
-      <div
+      <button
         v-if="showAvatar"
-        class="d-ps-relative"
+        class="d-ps-relative d-bgc-transparent d-baw0"
```
braddialpad: It's kind of hard for a user to visually know this is a button.. do we need some kind of visual indication on hover? Let's at least add `d-c-pointer` here. Who was the designer for this change?

recipes/list_items/contact_info/contact_info.vue
```diff
@@ -183,6 +184,14 @@ export default {
       default: null,
     },
   },
+
+  emits: ['avatar-click'],
```
braddialpad, line 197: capture this in storybook, as nacho mentioned.

braddialpad (changes requested):

recipes/list_items/contact_info/contact_info.vue
tonydialpad: I don't think we have a designer for this change, this requirement change comes from product team.

iropolo:
I still cannot see the click event on `action` tab in storybook.

You need to add in `contact_info.stories.js`

```
export const argsData = {
  **onAvatarClick: action('avatar-click'),**
  presence: null,
};
```

and inside `argTypesData`

 ```
 onAvatarClick: {
    table: {
      disable: true,
    },
  },

```

you should add  `@avatar-click="onAvatarClick" ` in `contact_info_default.story.vue` also.

File: recipes/list_items/contact_info/contact_info.vue [hunk 4]
```diff
@@ -7,12 +7,14 @@
     class="dt-contact-info"
   >
     <template #left>
-      <div
+      <button
         v-if="showAvatar"
-        class="d-ps-relative"
+        class="d-ps-relative d-bgc-transparent d-baw0 d-c-pointer"
         data-qa="contact-info-left"
+        :aria-label="avatarAriaLabel"
```
iropolo: Update it to `:aria-labelledby="avatarAriaLabel"` and a11y should be fixed.

iropolo (changes requested):

recipes/list_items/contact_info/contact_info.vue
tonydialpad: I applied the change as you suggested, but the a11y test was still failed: https://github.com/dialpad/dialtone-vue/actions/runs/6165158204/job/16732400646?pr=1164

do you have any idea about why?

iropolo: Oh right, there is another a11y failing test lol, lemme check

iropolo: ye seems like you cannot have a `<button>` inside a `dt-list-item` slot..

iropolo: We have sth similar in `feed_item_pill.vue` but we directly disabled a11y there.
Because the error was triggered on the `feed_item_pill_default.story.vue` not inside the component.

So maybe we can do the same here, but we definetly need to rewrite or rethink how we are using this `dt-item-layout`

So... you can disable the a11y in `recipes/list_items/contact_info/contact_info.stories.js`

```
a11y: {
      **disable: true,**
      config: {
        rules: [
          {
            id: 'aria-allowed-attr',
            enabled: false,
          },
          // Verify why contact info and avatar aren't getting unique IDs
          {
            id: 'duplicate-id',
            enabled: false,
          },
        ],
      },
    },
```

What you think @braddialpad, we could create a ticket to follow this issue and the one in `feed_item_pill_default.story.vue`. But for now disabling it here to not block this PR

iropolo (approved):

iropolo:
Approving it to not block the PR, will check it out the a11y issue with `dt-list-item` later. Jira ticket [DLT-1224](https://dialpad.atlassian.net/browse/DLT-1224)

[DLT-1224]: https://dialpad.atlassian.net/browse/DLT-1224?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

recipes/list_items/contact_info/contact_info.vue
braddialpad: hmm weird that you cannot have a button inside an `li`.. didn't think this was invalid. But yeah we can investigate this in the other ticket

braddialpad (approved):

PR #1165, opened 2023-09-09
feat: add avatar click event in contact info (vue3)

tonydialpad:
## :hammer_and_wrench: Type Of Change

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

This pull request adds `avatar-click` event in contact info component.

## :bulb: Context

When a user clicks on any of the avatars in the callbar when on a call with multiple other participants (i.e. a group call) it opens the “Manage participants” modal. To implement that, we need to add `avatar-click` event in contact info component.

## :pencil: Checklist

- [x] I have reviewed my changes
- [ ] I have added tests
- [x] I have added all relevant documentation
- [x] I have validated components with a screen reader
- [x] I have validated components keyboard navigation
- [x] I have considered the performance impact of my change
- [x] I have checked that my change did not significantly increase bundle size
- [ ] I am exporting any new components or constants in the index.js in the component directory
- [ ] I am exporting any new components or constants in the index.js in the root

## :crystal_ball: Next Steps

Make a new release for dialtone-vue

## :camera: Screenshots / GIFs
 No UI update

## :link: Sources

JIRA: https://dialpad.atlassian.net/browse/DP-71224
Vue2 PR: https://github.com/dialpad/dialtone-vue/pull/1164

PR #1176, opened 2023-09-13
feat(avatar): add clickable behavior

juliodialpad:
# Feat (Avatar): Add clickable behavior

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

- Added `clickable` property to `dt-avatar` to enable the possibility to focus and click on avatars.
- Removed the ability to select the text within the avatar.
- Added clickable avatar examples to variants.
- Added documentation
- Added tests

## :bulb: Context

Callbox component needs to emit an event when clicking on the avatar, also in [this PR](https://github.com/dialpad/dialtone-vue/pull/1164) a clickable avatar was needed and a custom implementation was made, so I thought that maybe a clickable avatar is a good feature to have.

## :pencil: Checklist

- [x] I have added tests

image


iropolo:
a11y test fail  on

`[TEST] FAIL browser: chromium components/avatar/avatar.stories.js`

`[TEST]       │    0    │ 'aria-command-name' │ 'serious' │ 'Ensures every ARIA button, link and menuitem has an accessible name' │   1   │`

iropolo:
Cool, looks good, I would revert the update in `recipes/list_items/contact_info/contact_info.vue` component here [#1164](https://github.com/dialpad/dialtone-vue/pull/1164/files).
And apply this update.

This is my suggested change to `recipes/list_items/contact_info/contact_info.vue` component:
As you will see, Im reverting
 `<button>` tag to `<div>`
 and 
`<span>` to `<div>`

Also it would probably necessary to remove `avatarAriaLabel` prop.

```
    <template #left>
      **<div**
        v-if="showAvatar"
        class="d-ps-relative d-bgc-transparent d-baw0 d-c-pointer"
        data-qa="contact-info-left"
      >
        **<div**
          v-if="avatarList"
          class="dt-contact-info--avatars d-mrn4 d-d-flex d-fd-row"
        >
          <dt-avatar
            v-for="(avatar, index) in avatarList"
            :key="index"
            **:clickable="true"**
            :size="avatarSize"
            :seed="avatar.seed"
            :full-name="avatar.fullName"
            :image-src="avatar.src"
            :icon-name="avatarIcon"
            :overlay-icon="avatar.icon"
            :overlay-text="avatar.text"
            :avatar-class="[{ 'd-mln24': index > 0, 'd-bc-brand': !!avatar.halo }]"
            **@click="avatarClick"**
          />
        **</div>**
        <dt-avatar
          v-else
          **:clickable="true"**
          :size="avatarSize"
          :full-name="avatarFullName"
          :image-src="avatarSrc"
          :icon-name="avatarIcon"
          :seed="avatarSeed"
          :presence="presence"
          **@click="avatarClick"**
        />
      **</div>**
    </template>

```

This is the final result:

image


Ping @tonydialpad here to check out this suggested change.

iropolo (approved):
Pre-approving it but need to fix the a11y issue.

Looks good improvement for avatar component, feel free to update `recipes/list_items/contact_info/contact_info.vue` component here, or create a ticket to update it later.

It will be necessary to update also the vue3 version of it, but is still pending [1177](https://github.com/dialpad/dialtone-vue/pull/1165#pullrequestreview-1625189400)

components/avatar/avatar.vue
```diff
@@ -6,7 +6,17 @@
...
+      :class="[
+        canvasClass,
+        'd-avatar__canvas',
+        {
+          'd-avatar--image-loaded': imageLoadedSuccessfully,
+          'd-avatar--clickable': clickable,
+        },
+      ]"
+      :tabindex="clickable ? 0 : -1"
+      :role="clickable ? 'button' : 'presentation'"
```
braddialpad: Can we change the actual element instead of just setting role? Can we have some sort of indicator on hover to show it is clickable? Maybe overlay it with a dark color with some opacity?

components/avatar/avatar.vue
```diff
@@ -309,6 +337,19 @@ export default {
...
+    avatarListeners () {
+      if (!this.clickable) return {};
+
+      return {
+        click: (e) => this.$emit('click', e),
+        keydown: (e) => {
+          if (!['Space', 'Enter'].includes(e.code)) return;
+          e.preventDefault();
+          this.$emit('click', e);
+        },
```
braddialpad: I think we wouldn't have to worry about all this if the element was an actual button

components/avatar/avatar.vue
juliodialpad: What element should I use instead, a button?
Maybe I can show the focus ring on hover too, I was using  `cursor: pointer` to show it is clickable on hover but maybe the opacity would work better.

@francisrupert @fede-dp please help us on this decision.

components/avatar/avatar.vue
juliodialpad: I'll try with a button

components/avatar/avatar.vue
braddialpad: Yeah it should be button I think

francisrupert:
@juliodialpad,

Sorry this got buried in my notifications.

Yes, button. 

```
.d-avatar--clickable:hover {
  border: var(--dt-size-border-100) solid var(--dt-color-border-default);
}

.d-avatar--clickable:active {
  border-color: var(--dt-color-border-moderate);
}
```

I don't mind the `.98` scaling for `:active`. Stick with that for now, and I can always change it after seeing it in play.

juliodialpad:
> a11y test fail on
> 
> `[TEST] FAIL browser: chromium components/avatar/avatar.stories.js`
> 
> `[TEST] │ 0 │ 'aria-command-name' │ 'serious' │ 'Ensures every ARIA button, link and menuitem has an accessible name' │ 1 │`

Fixed, good catch!

juliodialpad:
> Cool, looks good, I would revert the update in `recipes/list_items/contact_info/contact_info.vue` component here [#1164](https://github.com/dialpad/dialtone-vue/pull/1164/files). And apply this update.
> 
> This is my suggested change to `recipes/list_items/contact_info/contact_info.vue` component: As you will see, Im reverting `<button>` tag to `<div>` and `<span>` to `<div>`
> 
> Also it would probably necessary to remove `avatarAriaLabel` prop.
> 
> ```
>     <template #left>
>       **<div**
>         v-if="showAvatar"
>         class="d-ps-relative d-bgc-transparent d-baw0 d-c-pointer"
>         data-qa="contact-info-left"
>       >
>         **<div**
>           v-if="avatarList"
>           class="dt-contact-info--avatars d-mrn4 d-d-flex d-fd-row"
>         >
>           <dt-avatar
>             v-for="(avatar, index) in avatarList"
>             :key="index"
>             **:clickable="true"**
>             :size="avatarSize"
>             :seed="avatar.seed"
>             :full-name="avatar.fullName"
>             :image-src="avatar.src"
>             :icon-name="avatarIcon"
>             :overlay-icon="avatar.icon"
>             :overlay-text="avatar.text"
>             :avatar-class="[{ 'd-mln24': index > 0, 'd-bc-brand': !!avatar.halo }]"
>             **@click="avatarClick"**
>           />
>         **</div>**
>         <dt-avatar
>           v-else
>           **:clickable="true"**
>           :size="avatarSize"
>           :full-name="avatarFullName"
>           :image-src="avatarSrc"
>           :icon-name="avatarIcon"
>           :seed="avatarSeed"
>           :presence="presence"
>           **@click="avatarClick"**
>         />
>       **</div>**
>     </template>
> ```
> 
> This is the final result:
> 
> image
> Ping @tonydialpad here to check out this suggested change.

Created a ticket to solve this later: https://dialpad.atlassian.net/browse/DLT-1230

components/avatar/avatar.vue
juliodialpad: updated

juliodialpad:
Updated please re-review.

braddialpad (approved):
Great thanks!

The button moves the slightest amount when enabling clickable due to the scale change. Not really a big deal but could maybe cause a misalignment issue in some cases?

juliodialpad:
> Great thanks!
> 
> 
> 
> The button moves the slightest amount when enabling clickable due to the scale change. Not really a big deal but could maybe cause a misalignment issue in some cases?
> 
> 
> 
> 
> 
> 

Yeah, it might, I have two options here:

Making the border inset or adding the border to the non-clickable version which might cause a misalignment too.

Any advice here? @francisrupert

iropolo (approved):
Great, thanks!!

juliodialpad:
Added the border by default to the clickable avatar and just change the color on :hover and :active, this will increase the size of avatars which I don't think it's to much of an issue because this way the size will not change.

I was thinking on adding it to all the avatar, but that'd increase the size of all the avatars across the app and that'd create a lot of issues with QA

PR #1182, opened 2023-09-15
fix(avatar): change avatar seed algorithm to sync with android

braddialpad:
# fix(avatar): change avatar seed algorithm to sync with android

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Meant to do this a while back but forgot. Updated our avatar random generation algorithm to sync with android. 

## :bulb: Context

Our previous hashing algo FNV was not easy to implement on android, so I've updated our code to use the same algo as java's "hashCode". Our user avatar colors on web and mobile will be in sync after this change.

The exact same tests I've implemented here are also implemented on android, so we should be able to remain in sync.

## :pencil: Checklist

- [x] I have added tests

## :crystal_ball: Next Steps

Needs to be updated in Dialpad

## :link: Sources

https://www.w3schools.com/java/ref_string_hashcode.asp

### DIFF
--- a/common/utils.js
+++ b/common/utils.js
@@ -4,7 +4,6 @@ import {
   VALIDATION_MESSAGE_TYPES,
 } from './constants';
 import Vue from 'vue';
-import fnv from 'fnv-plus';
 
 let UNIQUE_ID_COUNTER = 0;
 let TIMER;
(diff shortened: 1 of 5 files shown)


common/utils.js
```diff
@@ -33,13 +32,37 @@ export function getUniqueString (prefix = DEFAULT_PREFIX) {
  */
 export function getRandomElement (array, seed) {
   if (seed) {
-    const hash = fnv.hash(seed);
-    return array[hash.value % array.length];
+    const hash = javaHashCode(seed);
+    return array[Math.abs(hash) % array.length];
   } else {
```
kimd13-dialpad, line 37: Is this function only used for Avatar or are there any other components that make use of it? If not, maybe we can rename it to `getAvatarBackgroundColor` or something more specific?

braddialpad: The idea was it could be used for any array I don't think it needs to be specific to avatar.

kimd13-dialpad: Awesome, having one algo for deciding coloring would be great!

iropolo (approved):
Thanks

PR #1186, opened 2023-09-18
fix(contact-info, feed-item-row, item-layout): alignment issues

juliodialpad:
# Fix (Contact row, Feed item row, Item layout): Alignment issues

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

- Fixed alignment issues with `dt-item-layout` based components
- Fixed class not being applied to  `content-text-wrapper-class`, is this the right styling now or was it correct before the class being applied?
- Renamed `feed-item-row*` references to `dt-feed-item-row*` to correct naming convention.

## :bulb: Context

Alignment issues were reported: https://dialpad.atlassian.net/browse/DP-80497 and found others while fixing them.

In this PR: https://github.com/dialpad/dialtone-vue/pull/1170 absolute positioning was beign used which is not correct, also we shouldn't use utility classes on dialtone-vue components anymore, so fixed it also.

## :pencil: Checklist

- [ ] I have added tests

## :crystal_ball: Next Steps

- Move `.dt-item-layout` styles to dialtone

### DIFF
--- a/components/item_layout/item_layout.vue
+++ b/components/item_layout/item_layout.vue
@@ -72,3 +72,16 @@ export default {
   },
 };
 </script>
+
+<style lang="less" scoped>
+// Move this to dialtone
+.dt-item-layout {
+  align-items: stretch;
+
+  &--content {
+    display: flex;
+    flex-direction: column;
+    justify-content: center;
+  }
+}
+</style>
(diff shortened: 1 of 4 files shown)


iropolo (approved):
Thanks for the update, lets see if `visual_test` actions works. Not sure if it will be triggered.

PR #1196, opened 2023-09-19
feat(callbox): add clickable behavior

juliodialpad:
# Feat(Callbox): Add clickable behavior

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

- Added clickable behavior to `title` and `avatar`
- Refactored to use `<div>` instead of nested `<dt-item-layout>` components because the alignment and spacings can be a mess to fix and that'd be hard to maintain.
- Clicking on either of the components emit the same 'click' event

## :bulb: Context

Needed clickable behavior on callbox

## :pencil: Checklist

- [x] I have added tests

Focus-visible
image
image

Hover
image
image


components/avatar/avatar.vue
```diff
@@ -5,7 +5,7 @@
     :class="avatarClasses"
     data-qa="dt-avatar"
     :aria-label="buttonAriaLabel"
-    @click="handleClick"
+    @click.stop="handleClick"
```
iropolo, line 8: What was the case here?

recipes/leftbar/callbox/callbox.vue
```diff
@@ -189,6 +177,13 @@ export default {
       return CALLBOX_BORDER_COLORS[this.borderColor];
     },
   },
+
+  methods: {
+    handleClick (e) {
+      if (!this.clickable) return;
+      this.$emit('click', e);
+    },
```
iropolo, line 185: thinking if it would be better to have separated events like emit `avatar-click` and `title-click`

recipes/leftbar/callbox/callbox.vue
```diff
@@ -199,26 +194,9 @@ export default {
...
-  }
-
-  &:deep(.dt-item-layout--right) {
-    flex-shrink: 0;
-  }
-
   &--video {
     border-radius: var(--dt-size-radius-200) var(--dt-size-radius-200) 0 0;
-    overflow: hidden;
+    overflow: clip;
```
iropolo, line 199: All this `<style>` comes from initially the `dt-item-layout` component and you update it a little bit?

iropolo (commented):
Could you provide us the callbox figma please?

braddialpad:
> Could you provide us the callbox figma please?

https://www.figma.com/file/oTY56LgQha8ljb4c8w4GtO/DT-Shared%3A-DP-App-Components?type=design&node-id=9240-7131&mode=design&t=H996k8ZImRUzD7o5-0

components/avatar/avatar.vue
juliodialpad: The event was being emitted twice on tests, the avatar click and the callbox click

recipes/leftbar/callbox/callbox.vue
juliodialpad: Both events do exactly the same on product, that's why I decided to keep it that way.

recipes/leftbar/callbox/callbox.vue
juliodialpad: No,  those custom styles are needed, with or without the `dt-item-layout`

recipes/leftbar/callbox/callbox.vue
braddialpad: Yeah I think it's fine to keep as one since this is a recipe for a specific product feature, it would be a different story if it was a generic component.

iropolo (approved):
Approving as we discussed, will need a migration to `dt-item-layout` after refactorized.

PR #1198, opened 2023-09-20
fix(contact-row): add space between presence and status text

braddialpad:
# fix(contact-row): add space between presence and status text

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Adds space between status and presence text.

## :bulb: Context

related to this DP 2.0 issue: https://dialpad.atlassian.net/browse/DP-79063

For some reason this is only a problem on vue 3.

## :pencil: Checklist

## :crystal_ball: Next Steps

Let DP 2.0 team know

Before:
Screenshot 2023-09-20 at 3 48 41 PM

After:
Screenshot 2023-09-20 at 3 48 25 PM


juliodialpad (commented):
Hmm, can you check if the `dt-leftbar-row__meta-custom` is being applied?

Maybe it has some left margin, but `dt-emoji-text-wrapper` has inheritAttrs: false and the class is not binding correctly.

braddialpad:
> Hmm, can you check if the `dt-leftbar-row__meta-custom` is being applied?
> 
> Maybe it has some left margin, but `dt-emoji-text-wrapper` has inheritAttrs: false and the class is not binding correctly.

they are certainly both being applied but I don't see any styles actually being applied by the classes. I'm a bit confused by this code in leftbar_row.less, but it seems like this class will not be applied unless the element with `dt-leftbar-row__meta-context` is within `dt-leftbar-row__meta-custom`? Currently they are siblings

```
&__meta-context {
    //
  }

  &__meta-custom:not(:empty) {
    &:deep(.dt-leftbar-row__meta-context ~ &:before) {
      content: ' • ';
      color: var(--dt-theme-sidebar-status-color-foreground);
    }
  }
```

iropolo (approved):

recipes/leftbar/contact_row/contact_row.vue
```diff
@@ -62,7 +62,7 @@ import DtAvatar from '@/components/avatar/avatar.vue';
 import { extractVueListeners, safeConcatStrings } from '@/common/utils';
 
 export default {
-  name: 'DtRecipeGroupRow',
+  name: 'DtRecipeContactRow',
```
iropolo, line 65: was this a leftover?

iropolo (approved):

recipes/leftbar/contact_row/contact_row.vue
braddialpad: Looks like it was always wrong

braddialpad:
I compared the CSS with app framework (https://dialpad.github.io/app-framework/inbox.html) and it seems the logic is not really correct.. seems kinda backwards. Appears to have broken around the same time we implemented `:deep`

see: [https://github.com/dialpad/dialtone-vue/commit/7e27e9c1b5ef5a7c77a9957d09b26c4fde9d38b3#diff-ac651f54cd1f090ca715d5a[…]836e4453c905af06240aa734dR276](https://github.com/dialpad/dialtone-vue/commit/7e27e9c1b5ef5a7c77a9957d09b26c4fde9d38b3#diff-ac651f54cd1f090ca715d5aa71de8c4e4b2bd2a836e4453c905af06240aa734dR276)

@juliodialpad I'm not sure exactly what this change was made for in the first place, please verify I'm not breaking anything else by changing this.

I think I will have to make this same change on the vue 2 branch now as well.

braddialpad:
Now correctly shows the dot

Screenshot 2023-09-21 at 4 24 38 PM

recipes/leftbar/contact_row/contact_row.vue
```diff
@@ -247,3 +247,7 @@ export default {
   },
 };
 </script>
+
+<style lang="less" scoped>
+@import "../style/leftbar_row.less";
+</style>
```
juliodialpad: This is not necessary as this component is using `dt-recipe-general-row` as base, which already includes the stylings

recipes/leftbar/group_row/group_row.vue
```diff
@@ -110,3 +110,7 @@ export default {
   },
 };
 </script>
+
+<style lang="less" scoped>
+@import "../style/leftbar_row.less";
+</style>
```
juliodialpad: This is not necessary as this component is using `dt-recipe-general-row` as base, which already includes the stylings

recipes/leftbar/style/leftbar_row.less
```diff
@@ -272,15 +272,10 @@
     white-space: nowrap;
   }
 
-  &__meta-context {
-    //
-  }
```
juliodialpad, line 277: It's ok to remove this

recipes/leftbar/style/leftbar_row.less
```diff
@@ -272,15 +272,10 @@
...
 
-  &__meta-custom:not(:empty) {
-    &:deep(.dt-leftbar-row__meta-context ~ &:before) {
-      content: ' • ';
-      color: var(--dt-theme-sidebar-status-color-foreground);
-    }
```
juliodialpad, line 283: Just remove the `&:deep` and it'd be working correctly.
I added that by mistake.

```css
&__meta-custom:not(:empty) {
  .dt-leftbar-row__meta-context ~ &:before {
    content: ' • ';
    color: var(--dt-theme-sidebar-status-color-foreground);
  }
}
```

juliodialpad (changes requested):
Make the appropriate changes, CP to vue 2 version and enable visual tests, please

recipes/leftbar/style/leftbar_row.less
francisrupert: Yeah, good catch on that tiny detail — I totally forgot about that. The `•` should indeed appear when `custom` follow `context` inline.

recipes/leftbar/contact_row/contact_row.vue
braddialpad: It seems to actually need this. Scoped isolates the styles to only the general_row component, so any of the classes in leftbar_row.less can not be used in contact_row

recipes/leftbar/style/leftbar_row.less
braddialpad: Updated

francisrupert: I recognize this is kind of a CSS workaround, such that the logic is in the CSS rather than in the Vue. 

Would it be cleaner to do a conditional, which can just insert it in between? Like say...

```html
<div class="dt-leftbar-row__status">
  <span class="dt-leftbar-row__meta-context">{presenceText}</span>
  <span aria-hidden="true">&bull;</span>
  <span class="dt-leftbar-row__meta-custom">{userstatus}</span>
</div>
```

recipes/leftbar/style/leftbar_row.less
juliodialpad, line 1: Your first suggestion was indeed very close and it's more readable, just move the :deep selector, I tested and seems to be working well.

Also removed the style imports from the other files and everything works.

Let me know if you're having any other issues so we can check what's happening.
```css
&:deep(&__meta-context ~ &__meta-custom:not(:empty):before) {
    content: ' • ';
    color: var(--dt-theme-sidebar-status-color-foreground);
  }
```

juliodialpad (commented):
You shouldn't need to import the file multiple times due to scoped as they're using the same base, it was just an issue with the :deep selector, please try my suggested change and let me know if something else is not working so I can check it more deeply

recipes/leftbar/style/leftbar_row.less
braddialpad: okay thanks looks good now with the imports removed

PR #1200, opened 2023-09-21
chore: update storybook to 7.4.2

braddialpad:
Vue 3


iropolo (approved):
looks good

braddialpad:
Looking into test break.

braddialpad:
vitest upgrade caused test failure. Don't really want to debug this right now so downgraded to 0.33.0 which fixes the issue.

PR #1209, opened 2023-09-25
feat: attachment carousel for generic media files

sanjeevdialpad:
# Attachment carousel

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Created the attachment carousel recipe that will be used in message input. This can handle any type of media files. The list of media files include: images, video, files.

## :bulb: Context

This PR contains changes pertaining to attachment carousel. For now i have used the generic DtImageViewer but it needs to be replaced with generic type that can handle all different types of media.
Each of these media type will also have a loader and preload state defined.

Figma: https://www.figma.com/file/GJiRmWORmDi0zxLJfwn8yp/DP-2.0---Conversation-View?type=design&node-id=2649-89072&mode=design&t=xHQr2H4s3pCjWKeU-0
JIRA: https://dialpad.atlassian.net/browse/DLT-1159

I will be creating separate PR to be merged into this PR for each component type.

## :pencil: Checklist

- [ ] I have added tests


braddialpad:
> The list of media files include: images, video, files.

What happens in the case of a user uploading a text file or an audio file or something? It doesn't use this carousel?

sanjeevdialpad:
@braddialpad I just checked dialpad upload, it only allows video/Image file. not even Text files. Do we want to support all types of files for this recipe?

I will ask our PM on what the requirement would be for dp 2.0

braddialpad:
> @braddialpad I just checked dialpad upload, it only allows video/Image file. not even Text files. Do we want to support all types of files for this recipe?

Fair enough, I think this is not great functionality, but if our goal is just to reach current DP 1.0 functionality then this is fine. Definitely don't let it hold back this PR. possible future update though.

recipes/conversation_view/attachment_carousel/attachment_carousel.mdx
braddialpad, line 1: This purposely not done yet?

recipes/conversation_view/attachment_carousel/attachment_carousel.vue
```diff
@@ -0,0 +1,263 @@
...
+        @mouseenter="closeButton(true, index)"
+        @mouseleave="closeButton(false, index)"
+      >
+        <dt-image-viewer
+          image-button-class="d-h64 d-w64 d-bar4 d-ba d-baw6 d-bc-subtle"
+          :image-src="mediaItem.path"
+          :image-alt="mediaItem.altText"
+          :close-aria-label="closeAriaLabel"
+          :aria-label="clickToOpenAriaLabel"
+        />
```
braddialpad: Maybe more of an image viewer problem, but should the image maintain its aspect ratio instead of being cut off like this?

Screenshot 2023-09-26 at 11 34 35 AM

recipes/conversation_view/attachment_carousel/attachment_carousel.vue
```diff
@@ -0,0 +1,263 @@
...
+    <dt-button
+      v-if="showRightArrow && showCarouselArrows"
+      tabindex="-1"
+      class="dt-attachment-carousel--right-arrow d-ba d-baw1 d-bc-default"
+      circle
+      size="xs"
+      importance="clear"
+      :aria-label="rightArrowAriaLabel"
+      @click="rightScroll"
+    >
```
braddialpad: Does the arrow need to dynamically reposition when you delete images?

Screenshot 2023-09-26 at 11 36 59 AM

recipes/conversation_view/attachment_carousel/attachment_carousel.vue
```diff
@@ -0,0 +1,263 @@
...
+      </template>
+    </dt-button>
+  </div>
+</template>
+
+<script>
+import { DtIcon } from '@/components/icon';
+import { DtImageViewer } from '@/components/image_viewer';
+import { DtButton } from '@/components/button';
+import {} from './attachment_carousel_constants';
```
braddialpad: can remove this if not used

recipes/conversation_view/attachment_carousel/attachment_carousel.vue
```diff
@@ -0,0 +1,263 @@
...
+      <div
+        v-for="(mediaItem, index) in mediaList"
+        :key="`media-${index}`"
+        role="presentation"
+        class="d-ps-relative"
+        @focusin="closeButton(true, index)"
+        @focusout="closeButton(false, index)"
+        @mouseenter="closeButton(true, index)"
+        @mouseleave="closeButton(false, index)"
+      >
```
braddialpad: Hmm something seems messed with keyboard functionality. Tabbing between them makes the focus go to some "hidden" place before going to the next item. Maybe it is going to the close button but it's invisible if it is.

recipes/conversation_view/attachment_carousel/attachment_carousel.vue
braddialpad: don't think these arrows need an aria label right? They are for visual purposes only.

recipes/conversation_view/attachment_carousel/attachment_carousel.vue
```diff
@@ -0,0 +1,263 @@
...
+    @focusin="showCarouselArrows = true"
+    @focusout="showCarouselArrows = false"
+  >
+    <div
+      v-if="mediaList.length > 0"
+      ref="carousel"
+      class="dt-attachment-carousel--media-list"
+      :aria-label="attachmentAriaLabel"
+      @scroll="handleScroll"
+    >
```
braddialpad: accessbiility warning here

```
.dt-attachment-carousel--media-list

Serious
aria-label attribute is not well supported on a div with no valid role attribute.
```

is there a better semantic element or role?

recipes/conversation_view/attachment_carousel/attachment_carousel.vue
```diff
@@ -0,0 +1,263 @@
...
+        @focusout="closeButton(false, index)"
+        @mouseenter="closeButton(true, index)"
+        @mouseleave="closeButton(false, index)"
+      >
+        <dt-image-viewer
+          image-button-class="d-h64 d-w64 d-bar4 d-ba d-baw6 d-bc-subtle"
+          :image-src="mediaItem.path"
+          :image-alt="mediaItem.altText"
+          :close-aria-label="closeAriaLabel"
+          :aria-label="clickToOpenAriaLabel"
```
braddialpad: Should you add the alt text in here so it says "click to open" and then a description of the image?

recipes/conversation_view/attachment_carousel/attachment_carousel.vue
```diff
@@ -0,0 +1,263 @@
...
+        :key="`media-${index}`"
+        role="presentation"
+        class="d-ps-relative"
+        @focusin="closeButton(true, index)"
+        @focusout="closeButton(false, index)"
+        @mouseenter="closeButton(true, index)"
+        @mouseleave="closeButton(false, index)"
+      >
+        <dt-image-viewer
+          image-button-class="d-h64 d-w64 d-bar4 d-ba d-baw6 d-bc-subtle"
```
braddialpad: make a custom class for this using tokens, we try not to use utility classes in dialtone vue anymore.

recipes/conversation_view/attachment_carousel/attachment_carousel.vue
```diff
@@ -0,0 +1,263 @@
...
+          </template>
+        </dt-button>
+      </div>
+    </div>
+
+    <!-- Carousel Arrows -->
+    <dt-button
+      v-if="showLeftArrow && showCarouselArrows"
+      tabindex="-1"
+      class="dt-attachment-carousel--left-arrow d-ba d-baw1 d-bc-default"
```
braddialpad: Remove utility classes and add them into the custom class (same goes for any others in this PR)

recipes/conversation_view/attachment_carousel/attachment_carousel.vue
```diff
@@ -0,0 +1,263 @@
...
+    };
+  },
+
+  computed: {
+
+  },
+
+  mounted: function () {
+    this.showLeftArrow = this.$refs.carousel.scrollLeft > 0;
+    this.showRightArrow = this.$refs.carousel.scrollWidth > this.$refs.carousel.clientWidth;
```
braddialpad, line 158: can you just call handleScroll?

recipes/conversation_view/attachment_carousel/attachment_carousel.vue
```diff
@@ -0,0 +1,263 @@
...
+
+.dt-attachment-carousel--close-button {
+  position: absolute;
+  color: var(--dt-color-neutral-white);
+  top: var(--dt-size-100);
+  right: var(--dt-size-100);
+  background-color: var(--dt-color-black-400);
+}
+.dt-attachment-carousel--left-arrow {
+  z-index: 100;
```
braddialpad: Do you need z-index here at all or does it work without?

recipes/conversation_view/attachment_carousel/attachment_carousel.stories.js
```diff
@@ -0,0 +1,118 @@
...
+
+const VariantsTemplate = (args) => createTemplateFromVueFile(
+  args,
+  DtRecipeAttachmentCarouselVariantsTemplate,
+);
+
+export const Variants = {
+  render: VariantsTemplate,
+  args: {},
+};
```
braddialpad: Can remove variants if it's not any different than default.

recipes/conversation_view/attachment_carousel/attachment_carousel_variants.story.vue
braddialpad, line 1: Remove file if not needed.

braddialpad (changes requested):
Lookin pretty good!

recipes/conversation_view/attachment_carousel/attachment_carousel.mdx
sanjeevdialpad: ye am probably gonna finish the documentation, variants and test cases in a separate PR. For now i will redo this with the most basic default story in it and description

recipes/conversation_view/attachment_carousel/attachment_carousel.vue
sanjeevdialpad: damn, i actually have this condition on remove media to handle this bug but seems like theres more corner cases. Let me try out.
```
removeMediaItem (index) {
      // make sure the carousel arrows is updated. 64 is the width of each media item
      this.showRightArrow = this.$refs.carousel.scrollWidth > (this.$refs.carousel.clientWidth + 64);
      this.$emit('remove-media', index);
    },
    ```

sanjeevdialpad: oh damn, i put that code in the other PR. Let me move it here

recipes/conversation_view/attachment_carousel/attachment_carousel.mdx
braddialpad: k sounds good.

sanjeevdialpad: added the docs for base style.

recipes/conversation_view/attachment_carousel/attachment_carousel.vue
sanjeevdialpad: right. removed.

recipes/conversation_view/attachment_carousel/attachment_carousel.vue
sanjeevdialpad: the showRightArrow is different in handleScroll. Once mounted, we check if there is more items than what can fit in the width. While on scroll we check if we reached the last item.

recipes/conversation_view/attachment_carousel/attachment_carousel.vue
sanjeevdialpad: works without it. removing.

recipes/conversation_view/attachment_carousel/attachment_carousel.vue
sanjeevdialpad: yea i noticed this. we manually set the image class but i think image viewer should have a way to set the height and width mainting the aspect ratio somehow passing that value as prop. if you think that idea is fine, we can create a ticket and fix this later.

recipes/conversation_view/attachment_carousel/attachment_carousel.vue
sanjeevdialpad: Yea this is something i wasnt able to fix. Still trying. It doesnt tab to the close button. Coz i kept a focus listener and it didnt get triggered. But i do see this issue.

recipes/conversation_view/attachment_carousel/attachment_carousel.vue
sanjeevdialpad: `image-alt`  is passed onto the image viewer alt text. Is that what you are talking about?

recipes/conversation_view/attachment_carousel/attachment_carousel.vue
sanjeevdialpad: changed it in both PRs.

recipes/conversation_view/attachment_carousel/attachment_carousel.vue
iropolo: maybe we can use the `object-fit: cover` css property

braddialpad: > yea i noticed this. we manually set the image class but i think image viewer should have a way to set the height and width mainting the aspect ratio somehow passing that value as prop. if you think that idea is fine, we can create a ticket and fix this later.

I don't think it needs to be that complicated, it should probably just always maintain the aspect ratio.

recipes/conversation_view/attachment_carousel/attachment_carousel.vue
braddialpad: I was thinking something like.. ``:image-alt="`Click to open ${mediaItem.altText}`"``

But I realized this is an i18n issue so nevermind.

recipes/conversation_view/attachment_carousel/attachment_carousel.vue
braddialpad: Looks good after update.

recipes/conversation_view/attachment_carousel/attachment_carousel.vue
sanjeevdialpad: > maybe we can use the `object-fit: cover` css property

this works perfectly.

recipes/conversation_view/attachment_carousel/attachment_carousel.vue
sanjeevdialpad: i just put this on the parent one since it has a role presentation. but theres accessbility issue on progress bar am gonna fix.

recipes/conversation_view/attachment_carousel/attachment_carousel.vue
sanjeevdialpad: solved using opacity to show hide buttons

sanjeevdialpad:
@braddialpad this is ready for final review!

braddialpad (approved):
Looks good once you fix the accessibility error!

PR #1218, opened 2023-09-27
fix(callbox, avatar): events and styling

juliodialpad:
# Fix (Callbox, Avatar): Events and styling

## :hammer_and_wrench: Type Of Change

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

- Removed `.stop` modifier from click events to allow click events to bubble up.
- Added `emits` to callbox to avoid duplicating native click event on vue 3
- Changed clickable title styling to use underline instead of background color on :hover and :active.
- Updated callbox variants to use `<dt-stack>` instead of custom divs with `d-d-flex`

## :bulb: Context

Received [design feedback](https://dialpad.atlassian.net/browse/DP-81452) on callbox implementation
Received a [bug report](https://dialpad.atlassian.net/browse/DP-81391) that avatar wasn't triggering the click event correctly.

## :pencil: Checklist

- [x] I have reviewed my changes
- [ ] I have added tests
- [ ] I have added all relevant documentation
- [ ] I have validated components with a screen reader
- [x] I have validated components keyboard navigation
- [ ] I have considered the performance impact of my change
- [ ] I have checked that my change did not significantly increase bundle size
- [ ] I am exporting any new components or constants in the index.js in the component directory
- [ ] I am exporting any new components or constants in the index.js in the root

## :crystal_ball: Next Steps

Release and update on product.

## :camera: Screenshots / GIFs

image
image

PR #1219, opened 2023-09-27
fix(callbox, avatar): events and styling

juliodialpad:
# Vue 3 version of #1218

### DIFF
--- a/components/avatar/avatar.vue
+++ b/components/avatar/avatar.vue
@@ -5,7 +5,7 @@
     :class="avatarClasses"
     data-qa="dt-avatar"
     :aria-label="buttonAriaLabel"
-    @click.stop="handleClick"
+    @click="handleClick"
   >
     <div
       ref="canvas"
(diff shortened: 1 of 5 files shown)


iropolo:
In this branch all clicks events (on variants) are shown in the `action` tab of storybook, this does not happen in the vue2 branch.
Idk if it is sth intendeed or in vue2 it is not working.

Pre-approving but take a look on it.
Thanks,

iropolo (approved):
In this branch all clicks events (on variants) are shown in the action tab of storybook, this does not happen in the vue2 branch.
Idk if it is sth intendeed or in vue2 it is not working.

Pre-approving but take a look on it.
Thanks,

juliodialpad:
> In this branch all clicks events (on variants) are shown in the action tab of storybook, this does not happen in the vue2 branch. Idk if it is sth intendeed or in vue2 it is not working.
> 
> Pre-approving but take a look on it. Thanks,

That's kind of a bug with storybook, because the event name is called `onClick` which is the native click event.

Added the event to `emits` so storybook know I'm going to handle the event myself so no click events are being triggered on that variants_story.js

Every component that emits click event might have this behavior too, well need to verify.

PR #1221, opened 2023-09-28
fix(list-item): apply correct color design tokens

francisrupert:
# List item: apply correct color design tokens

image

## :hammer_and_wrench: Type Of Change

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :pencil: Checklist

- [x] I have reviewed my changes
- [ ] I have added tests
- [ ] I have added all relevant documentation
- [ ] I have validated components with a screen reader
- [x] I have validated components keyboard navigation
- [x] I have considered the performance impact of my change
- [x] I have checked that my change did not significantly increase bundle size
- [ ] I am exporting any new components or constants in the index.js in the component directory
- [ ] I am exporting any new components or constants in the index.js in the root

PR #1223, opened 2023-09-28
fix(list-item): apply correct color design tokens - vue3

braddialpad:
vue 3 https://github.com/dialpad/dialtone-vue/pull/1221/files


iropolo (approved):
Thanks

PR #1227, opened 2023-09-29
chore: add release script

juliodialpad:
# Chore: Add release script

- [ ] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Added a release script to automate the release command a bit more. This way you only need to be on either staging/staging-vue3 and run `npm run release` and it'd pull the latest changes, run semantic release, checkout to corresponding branch, merge and push the changes.

## :bulb: Context

I always struggle having to read documentation on how to release dialtone-vue again because I forgot the commands and the order and this might lead to some errors.

## :pencil: Checklist

- [ ] I have added tests

## :crystal_ball: Next Steps

Update documentation and remove the unneccesary steps.

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "lint:docs": "markdownlint '**/*.mdx'",
     "prepublishOnly": "npm run build",
     "pretest": "npm run clean",
-    "release": "bash ./scripts/release",
+    "release": "./scripts/release.sh",
     "start": "npm run storybook",
     "storybook": "STORYBOOK_ENV=development storybook dev -p 9010",
     "storybook:clean": "rm -rf ./storybook-static",
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "lint:docs": "markdownlint '**/*.mdx'",
     "prepublishOnly": "npm run build",
     "pretest": "npm run clean",
-    "release": "semantic-release --no-ci --extends ./release-local.config.cjs",
+    "release": "bash ./scripts/release",
     "start": "npm run storybook",
     "storybook": "STORYBOOK_ENV=development storybook dev -p 9010",
     "storybook:clean": "rm -rf ./storybook-static",
--- a/scripts/release
+++ b/scripts/release
@@ -1,0 +1,26 @@
+#!/bin/sh
+set -e
+
+branch="$(git rev-parse --abbrev-ref HEAD)"
+
+if [[ $branch != "staging" || $branch != "staging-vue3" ]]; then
+  echo "This script can only be run while on staging or staging-vue3 branches"
+  exit 1;
+fi
+
+git pull &&
+semantic-release --no-ci --extends ./release-local.config.cjs;
+
+case $branch in
+  "staging")
+    git checkout production &&
+    git merge --ff-only staging &&
+    git push -u origin production
+    ;;
+
+  "staging-vue3")
+    git checkout vue3 &&
+    git merge --ff-only staging-vue3 &&
+    git push -u origin vue3
+    ;;
+esac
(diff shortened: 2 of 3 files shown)


braddialpad (changes requested):
I thought of adding this a while back!

We'll need to update the releasing documentation as it will still say to do this the manual way.

My only real concern is error handling.. if some kind of error occurs and this process only "partially" completes it might be a bit confusing how to fix it. But we'll try it out and see how it goes, might not be that much of a problem.

juliodialpad:
> I thought of adding this a while back!
> 
> 
> 
> We'll need to update the releasing documentation as it will still say to do this the manual way.
> 
> 
> 
> My only real concern is error handling.. if some kind of error occurs and this process only "partially" completes it might be a bit confusing how to fix it. But we'll try it out and see how it goes, might not be that much of a problem.

Cool, I'll update the documentation then.

Yes, I was thinking about error handling, I'll think on what should I add to documentation regarding that, all errors should output to console anyways.

scripts/release
braddialpad, line 1: add extension `release.sh`

scripts/release
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
```
braddialpad: If this needs to be executed in bash it should be `#!/usr/bin/env bash`. `#!/bin/sh` refers to "current system shell" which could be something other than bash.

package.json
```diff
@@ -16,7 +16,7 @@
     "lint:docs": "markdownlint '**/*.mdx'",
     "prepublishOnly": "npm run build",
     "pretest": "npm run clean",
-    "release": "semantic-release --no-ci --extends ./release-local.config.cjs",
+    "release": "bash ./scripts/release",
```
braddialpad: because you have the shebang `#!/usr/bin/env bash` you can just execute this without 'bash'

`"release": "./scripts/release.sh",`

scripts/release
braddialpad: I'm not sure if I can even check this on github, but did you make this executable via git add? I know you usually have to do this when adding executable files.

braddialpad (changes requested):

scripts/release
juliodialpad: I think it was added automatically as I did `chmod +x release.sh` before committing, but ran this commands anyways just in case: https://stackoverflow.com/a/38285462

scripts/release
juliodialpad: ~It can be executed in any shell it's not specific to bash I think, executed with `bash ./scripts/release` on package.json because I was having some errors but weren't related, will remove that and preserve the  #!/bin/sh.~

Updated shebang, it was using the bash feature `[[ ]]` so its necessary to execute this on bash. Thanks for the suggestion!

scripts/release.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/sh
+set -e
+
+branch="$(git rev-parse --abbrev-ref HEAD)"
+
+if [[ $branch != "staging" || $branch != "staging-vue3" ]]; then
```
iropolo: hmm... this will always `true`... is this on purpose?

I mean.. `$branch` will always not be equal to one of the two values...
if branch is `staging` then it will not be `staging-vue3`so it is `true`, Do Im missing sth?

Should be `&&` ... right? Then if you are on either the `staging` or `staging-vue3` will continue but on any other branch the script will echo the error message and exit.

juliodialpad: Good catch, updated the if logic, thanks!

scripts/release
braddialpad: Yeah I remember having to do `git add --chmod=+x` before even if the file was locally executable.. Anyways should be good.

braddialpad (approved):
Looks good!

PR #1231, opened 2023-10-04
fix(root-layout): remove fixed height on header/footer

braddialpad:
# fix(root-layout): remove fixed height on header/footer

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Still having issues with root layout not collapsing the footer when it's empty. I'm just removing the fixed height and having it take the height of whatever element is passed into so the consumer has full control. Might be a breaking change so I'll check usages before releasing this. I don't think it's used much and was requested by DP 2.0 team so I'll guide them to make the update.

Also removed the root_layout_body component and integrated it into the main component, as it is no longer necessary after we refactored this to use grid.

## :bulb: Context

min-height on the footer and header components is causing the footer to remain at it's height even with no content in it. Allow the consumer to control how they manage the content of the header footer instead of us since there are many potential variations.

## :pencil: Checklist

- [x] I have added tests

## :crystal_ball: Next Steps

Notify product teams of potential breaking change.


iropolo (approved):
I forget to click on approve here earlier..

components/root_layout/root_layout.vue
```diff
@@ -84,8 +75,7 @@ export default {
     },
 
     /**
-     * The height of the header
-     * Possible units rem|px|%|em
+     * DEPRECATED: set the height of the inner element instead.
```
iropolo, line 78: When we will remove these `DEPRECATED` props? after checking it with the usage on web dev?

braddialpad: > When we will remove these `DEPRECATED` props? after checking it with the usage on web dev?

Yep exactly, after we check and fix any code still using them.

PR #1232, opened 2023-10-04
fix: alignment for emoji and emoji row

ninamarina:
# fix: alignment for emoji and emoji row
Jira ticket: https://dialpad.atlassian.net/browse/DLT-1247
Fixes the alignment of the emoji inside the pill from the emoji row.

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description
The `span` container was using the class `emojiSize`, but that was doing nothing since the variable is not used without the class `d-icon`. So I added that class, and also to be able to set the height I needed to add `display: inline-block`.
This change is changing the height of the pill, which I think is fine, but that makes the visual tests fail. If the reviewers agree with these changes, I will change the snapshots used in the tests.

## :pencil: Checklist

- [ ] I have added tests

| Before these changes | After these changes |
|------|--------|
| image | image |


components/emoji/emoji.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-  <span :class="emojiSize">
+  <span :class="['d-d-inline-block', 'd-icon', emojiSize]">
```
braddialpad: We try not to use utility classes in dialtone-vue anymore due to their immutability. I know there are still tons of components in dialtone-vue using them, but any new stuff we put in should use defined css classes instead, even if it's super simple like this.

You can simply define a class like this in the <style> section:

```
.dt-emoji {
  display: inline-block;
}
```

recipes/conversation_view/emoji_row/emoji_row.vue
```diff
@@ -26,7 +26,7 @@
             :attrs="attrs"
             @click="emojiClicked(reaction)"
           >
-            <span class="d-mr4">
+            <span class="d-mr4 d-d-inline-flex">
```
braddialpad, line 29: this one is fine to leave since this file is already full of utility classes and we'll have to do a whole migration anyways.

braddialpad:
Hmm for some reason they still look kinda misaligned on my screen at normal zoom. If I zoom in a few levels though it looks fine.... Some sort of scaling issue? It's pretty minor, this just happening to me?

Screenshot 2023-10-04 at 1 56 42 PM

vs browser zoom 175%

Screenshot 2023-10-04 at 1 59 12 PM

recipes/conversation_view/emoji_row/emoji_row.vue
iropolo: Yes they are a lot on our recipes, do we have a ticket to make this migration?

iropolo (approved):
Pre-approved

recipes/conversation_view/emoji_row/emoji_row.vue
braddialpad: Yep it's here https://dialpad.atlassian.net/browse/DLT-1125

iropolo:
> Hmm for some reason they still look kinda misaligned on my screen at normal zoom. If I zoom in a few levels though it looks fine.... Some sort of scaling issue? It's pretty minor, this just happening to me?
> 
> Screenshot 2023-10-04 at 1 56 42 PM
> 
> vs browser zoom 175%
> 
> Screenshot 2023-10-04 at 1 59 12 PM

I'm seeing them right.

Prev: 
Screenshot 2023-10-04 at 18 59 24

Now:
Screenshot 2023-10-04 at 19 00 41

braddialpad:
> I'm seeing them right.

Interesting.. I think I found the issue.

@ninamarina Can you try removing the font-size from `dt-emoji-row__reaction` in emoji_row.vue?

ninamarina:
> dt-emoji-row__reaction

@braddialpad yes, that solves the issue!

components/emoji/emoji.vue
ninamarina: Fixed in b148e5fd291f98af59d7ca8c65417f79a0be2a6b
I still had to leave the classes `d-icon` and `emojiSize` since that class defines the variables for the sizes, I suppose thats ok?

braddialpad: Yep, those ones are not utility classes

braddialpad (approved):
Looks great thanks!

PR #1233, opened 2023-10-04
fix: alignment for emoji and emoji row - vue3

ninamarina:
# fix: alignment for emoji and emoji row
Jira ticket: https://dialpad.atlassian.net/browse/DLT-1247
Fixes the alignment of the emoji inside the pill from the emoji row.

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description
The `span` container was using the class `emojiSize`, but that was doing nothing since the variable is not used without the class `d-icon`. So I added that class, and also to be able to set the height I needed to add `display: inline-block`.
This change is changing the height of the pill, which I think is fine, but that makes the visual tests fail. If the reviewers agree with these changes, I will change the snapshots used in the tests.

## :pencil: Checklist

- [ ] I have added tests

| Before these changes | After these changes |
|------|--------|
| image | image |


iropolo (approved):
Pre-approved.
Dont forge to update this pr with the updates of #1232 .

braddialpad (approved):

PR #1237, opened 2023-10-05
feat(icons): update dialtone icons

iropolo:
# Update dialtone icons

Updating dialtone icons package in vue3.


braddialpad (changes requested):
Hey don't worry about this, I just did it, releasing currently.

PR #1242, opened 2023-10-05
fix(callbar-button): slot not working

juliodialpad:
# Fix (Callbar button): Slot not working

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Fixed button icon slot that was incorrectly using <slot> instead of <template>

## :bulb: Context

Issues were reported on callbar button with popover on vue 3 version, but found it was incorrect on vue 2 version too.

## :pencil: Checklist

- [ ] I have added tests

### DIFF
--- a/recipes/buttons/callbar_button/callbar_button.vue
+++ b/recipes/buttons/callbar_button/callbar_button.vue
@@ -195,7 +195,7 @@ export default {
     callbarButtonListeners () {
       return {
         ...this.$listeners,
-        click: (event) => this.$emit('click'),
+        click: (event) => this.$emit('click', event),
       };
     },
   },
--- a/recipes/buttons/callbar_button/callbar_button.vue
+++ b/recipes/buttons/callbar_button/callbar_button.vue
@@ -19,10 +19,9 @@
           v-on="$listeners"
         >
           <slot />
-          <slot
-            slot="icon"
-            name="icon"
-          />
+          <template #icon>
+            <slot name="icon" />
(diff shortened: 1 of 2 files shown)


recipes/buttons/callbar_button/callbar_button.vue
```diff
@@ -192,6 +192,13 @@ export default {
       }
       return this.circle ? 'outlined' : 'clear';
     },
+
+    callbarButtonListeners () {
+      return {
+        ...this.$listeners,
+        click: (event) => this.$emit('click', event),
+      };
+    },
```
braddialpad, line 201: Hmm I am confused by this. We are now capturing an click from the tooltip wrapper, it should be captured from the button right? since that is the interactive element.

juliodialpad: We were already capturing the click from the tooltip itself because we had `inheritAttrs: true` we were just duplicating the listener on the button too. I did it this way to avoid breaking the tests but I can move the events to the button and fix the tests.

braddialpad: Hmm yeah I think we should do that, it's the correct way to do it.

juliodialpad: Just moved the events to button, vue2 and vue3 tests were out of sync too.

braddialpad (approved):

PR #1243, opened 2023-10-05
fix(callbar-button): slot not working

juliodialpad:
# Vue 3 version of #1242 

Screenshot 2023-10-05 at 2 27 15 p m
Screenshot 2023-10-05 at 2 27 30 p m
Screenshot 2023-10-05 at 2 31 23 p m
Screenshot 2023-10-05 at 2 31 33 p m
Screenshot 2023-10-05 at 2 31 42 p m

### DIFF
--- a/recipes/buttons/callbar_button/callbar_button.vue
+++ b/recipes/buttons/callbar_button/callbar_button.vue
@@ -16,7 +16,6 @@
           :label-class="callbarButtonTextClass"
           :width="buttonWidth"
           :class="callbarButtonClass"
-          v-on="callbarButtonListeners"
         >
           <slot />
           <template #icon>
@@ -39,8 +38,6 @@ export default {
   name: 'DtRecipeCallbarButton',
 
   components: { DtButton, DtTooltip },
-
-  inheritAttrs: true,
 
   props: {
     /**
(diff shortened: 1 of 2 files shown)


recipes/buttons/callbar_button/callbar_button.vue
```diff
@@ -142,6 +143,16 @@ export default {
...
 
+  emits: [
+    /**
+     * Native click event
+     *
+     * @event click
+     * @type {PointerEvent | KeyboardEvent}
+     */
+    'click',
+  ],
```
braddialpad, line 154: If we remove this and the click event from `callbarButtonListeners()` and have `v-on="callbarButtonListeners"` on the button it should natively emit this event without us having to manually emit.

In vue 3 setting something in emits actually prevents it from natively emitting.

juliodialpad: Yeah I know, but that way it wouldn't be on the generated documentation, I think we need to add this to our components so it actually outputs to the documentation and it is explicit that we're emitting a events from our components.

braddialpad: Fair enough

braddialpad (approved):

PR #1244, opened 2023-10-05
fix(link): href null on link

sanjeevdialpad:
# fix: ability to not navigate when href is not passed to dt-link

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

We will reuse dt-link on mention for message input in which case we do not want to navigate anywhere else. this change will allow to do that coupled with the native event link so user can do specific actions without navigating away.


braddialpad:
Hey the accessibility error is happening because the BreadcrumbItem story is pretty messed up. This PR should fix it https://github.com/dialpad/dialtone-vue/pull/1246

braddialpad (approved):
Merge in staging and you're good to go when all tests pass.

iropolo (approved):

iropolo (approved):
Thanks!

PR #1246, opened 2023-10-05
fix(breadcrumbs): fix story configuration of BreadcrumbItem

braddialpad:
# fix(breadcrumbs): fix story configuration of BreadcrumbItem

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Breadcrumb Item story config was not great, it was being passed the props from breadcrumbs which are completely different. Figured out how to make it it's own story but still be under "breadcrumbs" this pattern will be useful for future sub components like this.

## :bulb: Context

Caused accessibility error in sanjeevs PR.

## :pencil: Checklist

- [x] I have added tests


components/breadcrumbs/breadcrumbs.stories.js
```diff
@@ -1,6 +1,5 @@
 import { createTemplateFromVueFile } from '@/common/storybook_utils';
```
sanjeevdialpad, line 1: in this file, we have a description here: https://github.com/dialpad/dialtone-vue/pull/1246/files#diff-88d4eddbc5b117284ff3294a480b8891ac689e6d85117ea205b5d8f4f002f10aR50

Where the object contains url instead of href which is misleading a bit. 
```
summary: '{ url: string, label: string }[]',
```

sanjeevdialpad (approved):
ty so much!

components/breadcrumbs/breadcrumbs.stories.js
braddialpad: thanks, definitely outdated.

PR #1248, opened 2023-10-05
feat(rich-text-editor): contact mentions

sanjeevdialpad:
# feat(rich-text-editor): Contact mentions

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

The tiptap Mention plugin is apt for our use case of mentioning users/ channels within the editor. It also creates the respective nodes with ability for user to customize the trigger character. The dev needs to pass in the suggestion object which provides the plugin with the function to search and display the suggested users with the keyed in characters(query).

Currently this PR only enables the CONTACT mention plugin. We will have to duplicate and create a custom 2nd plugin for channel mentions in a similar way.

## How to use the mention plugin.

When using the Rich Text editor, the developer would have to pass in the suggestion object prop which MUST contain an items function that takes in a query and returns a list of object with the properties: `{ contactId: String, name: String, avatarSrc: String }`. There can be additional properties that override the mention plugin and more details can be seen here: https://tiptap.dev/api/utilities/suggestion . 

For example:
```
<dt-rich-text-editor
    v-model="inputValue"
    :mention-suggestion="mentionSuggestion"
  />
```

If this mentionSuggestion Object prop is not supplied, the mention plugin is disabled by default.

To see it in action type '@' into rich editor here:  https://vue.dialpad.design/deploy-previews/pr-1248/?path=/story/components-rich-text-editor--with-mention-suggestions

## :pencil: Checklist

- [ ] I have added tests

## :crystal_ball: Next Steps

We need to introduce a new prop, channelSuggestion which is the exact suggestion object for contact mentions but will search through channels. This also needs to have a separate plugin key and name so tiptap does not complain. this way we can extend the mention plugin to multiple instances as needed.

## :link: Sources

Multiple Mention capability: https://github.com/ueberdosis/tiptap/issues/2219


components/rich_text_editor/rich_text_editor.vue
```diff
@@ -254,6 +273,10 @@ export default {
   },
 
   methods: {
+    onclick (event) {
```
sanjeevdialpad: will remove this. was experimenting something. got left over.

components/rich_text_editor/mention_suggestion.js
```diff
@@ -0,0 +1,5 @@
+export default {
```
sanjeevdialpad, line 2: this file/object would need to come from the product and it is how product can override any behaviour of the mention plugin. or like this it can expose the function that gets the query to return list of suggestions to be shown

iropolo:
What are the instructions or the steps we should follow to replicate the behaviours? thanks.

sanjeevdialpad:
Oh my bad. Just finished working on this late evening yesterday. Forgot to edit the description. Will add more details. 

I have added another storybook doc called with channel mentions. In that editor you just type @ and it will pop up 3 test result for now.

sanjeevdialpad:
> What are the instructions or the steps we should follow to replicate the behaviours? thanks.

Hey @iropolo I have edited the description with kind of a Readme gist of usage.  Also adding this to storybook.

braddialpad:
This is not just contact mentions but also emoji suggestions right?

components/rich_text_editor/extensions/emoji/EmojiSuggestion.vue
```diff
@@ -0,0 +1,40 @@
...
+    class="dt-emoji-suggestion"
+  >
+    <dt-emoji
+      size="200"
+      :code="item"
+    />
+    <div class="dt-emoji-suggestion--text">
+      {{ item }}
+    </div>
+  </div>
```
braddialpad: can probably use dt-stack with gap here instead of margin. Always avoid using margin for spacing between items if possible.

components/rich_text_editor/extensions/mentions/MentionComponent.vue
```diff
@@ -0,0 +1,50 @@
...
+  color: var(--dt-color-link-primary);
+  border-radius: var(--dt-space-200);
+  height: 1.8rem;
+}
+.dt-link--mention:hover {
+  background-color: hsl(var(--dt-color-link-primary-hsl) / 10%);
+  color: var(--dt-color-link-primary);
+  border-radius: var(--dt-space-200);
+  height: 1.8rem;
+}
```
braddialpad: Are both these classes supposed to be identical?

components/rich_text_editor/extensions/mentions/MentionSuggestion.vue
```diff
@@ -0,0 +1,49 @@
...
+  <div class="dt-mention-suggestion">
+    <dt-avatar
+      :full-name="name"
+      :image-src="avatarSrc"
+      size="xs"
+    />
+    <div class="dt-mention-suggestion---text">
+      {{ name }}
+    </div>
+  </div>
```
braddialpad: dt-stack

components/rich_text_editor/extensions/suggestion/SuggestionList.vue
```diff
@@ -95,37 +102,33 @@ export default {
...
-  overflow-y: scroll !important;
-}
-.dt-emoji-suggestion-list .dt-item-layout--title {
-  display: flex;
 }
-.dt-emoji-suggestion-list-item {
-  display: flex !important;
+
+.dt-suggestion-list--item {
+  display: list-item;
```
braddialpad: not necessary, this is the default.

components/rich_text_editor/extensions/suggestion/SuggestionList.vue
```diff
@@ -95,37 +102,33 @@ export default {
...
 </script>
 
 <style>
-.dt-emoji-suggestion-list {
-  background-color: var(--dt-color-surface-secondary) !important;
-  border-color: var(--dt-color-border-subtle) !important;
+.dt-suggestion-list {
+  position: relative;
+  padding: 0rem;
+  margin: .4rem;
```
braddialpad: instead set padding to `var(--dt-size-300)` and remove margin.

components/rich_text_editor/extensions/suggestion/SuggestionList.vue
```diff
@@ -95,37 +102,33 @@ export default {
...
+
+.dt-suggestion-list--item {
+  display: list-item;
   border: var(--dt-size-100) solid transparent;
-  min-width: var(--dt-size-850);
-  width: var(--dt-size-100-percent) !important;
+  /* min-width: var(--dt-size-850); */
 
   &.is-selected {
     border-color: var(--bc-bold);
```
braddialpad: `var(--dt-color-border-bold)`

components/rich_text_editor/rich_text_editor.vue
```diff
@@ -114,6 +117,15 @@ export default {
       type: [Boolean, Object],
       default: false,
     },
+
+    /**
+     * suggestion object containing the items query function. When null, it does not add the plugin.
+     */
```
braddialpad: Describe what the required keys of the objects passed into this are.

components/rich_text_editor/rich_text_editor.vue
```diff
@@ -114,6 +117,15 @@ export default {
...
+    mentionSuggestion: {
+      type: Object,
+      default: null,
+      // required: true,
+    },
```
braddialpad: Hmm is this an object type? Appears to be a function that returns an array of objects?

braddialpad (changes requested):

sanjeevdialpad:
> This is not just contact mentions but also emoji suggestions right?

I kind of just abstracted the suggestion list component out from emoji and is used in contact, and in future for channels or any plugin that would want to show suggestions. but yes, there are changes on emoji suggestion file.

components/rich_text_editor/extensions/mentions/MentionComponent.vue
sanjeevdialpad: I should have used &: hover and am missing something am sure, thanks for pointing out.

sanjeevdialpad: made changes but need to replace hsl values later.

components/rich_text_editor/rich_text_editor.vue
sanjeevdialpad: changed the description to explain the object. Also introduced doc changes to make it easier to use this.

sanjeevdialpad: its an object that contains the function, i did it this way so it can be further customized by product if needed. for example if they want to override the render function or the trigger character, all can be done in this object. I have documented it here:
https://github.com/dialpad/dialtone-vue/pull/1248/files#diff-8e1296ebf37f15b6140a0f899b263bf8b5461c291a8a33f3bb32f09006ed6534R71

braddialpad (approved):
Looks good.

iropolo:
Some functionalities where I have doubts.

- when I write `@test2` and press enter it select `@test`.
- there is no filters on the list, if I write `@test2` I will expect the list will only shows me `test2` matchs.

components/rich_text_editor/extensions/mentions/MentionSuggestion.vue
```diff
@@ -0,0 +1,43 @@
...
+      :full-name="name"
+      :image-src="avatarSrc"
+      size="xs"
+    />
+    {{ name }}
+  </dt-stack>
+</template>
+
+<script>
+import DtAvatar from '@/components/avatar/avatar.vue';
```
iropolo: import { DtAvatar } from '@/components/avatar';

components/rich_text_editor/extensions/suggestion/SuggestionList.vue
```diff
@@ -95,37 +102,30 @@ export default {
...
       const item = this.items[index];
 
-      if (item) {
-        this.command({ code: item });
+      switch (this.itemType) {
+        case 'emoji':
+          this.command({ code: item });
+          return;
+        case 'mention':
+          this.command({ name: item.name, contactId: item.contactId, avatarSrc: item.avatarSrc });
```
iropolo, line 110:
```
this.command({ name: item.name, contactId: item.contactId, avatarSrc: item.avatarSrc });
break;
```

iropolo (approved):
looks good, just doubts in my previous comment, but if that is sth intended or will be tackled later, will recommend to create ticket to handle it.

sanjeevdialpad:
> Some functionalities where I have doubts.
> 
> * when I write `@test2` and press enter it select `@test`.

When i select using click and enter,  it does select the second one or third one. I will probably reach to you to see this bug.
Screenshot 2023-10-12 at 11 16 28 AM

> * there is no filters on the list, if I write `@test2` I will expect the list will only shows me `test2` matchs.

Ah this is my lazy self not hooking up the items function to filter the list. Right now am just returning a static result of 3 strings no matter the input string. You could see this here:
https://github.com/dialpad/dialtone-vue/pull/1248/files#diff-275bf53e59452b453bcbd64dba5aab954de917182235a2bce6d0e132b823a5f7R1

I will change the function to filter it using the query string but this function basically comes from product.

braddialpad:
> I will change the function to filter it using the query string but this function basically comes from product.

I think we might be better off leaving it as it is as the component doesn't actually do this. It will be unclear whether the component does the filtering itself or not

PR #1253, opened 2023-10-11
fix(root-layout): convert to pure CSS grid method

braddialpad:
# fix(root-layout): convert to pure CSS grid method

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Well it's been a long time coming but root-layout has now been fully converted to CSS grid. Everything in the `<style>` section will be removed from this PR before merging, and will replace the root-layout css in the dialtone repo. Just for preview purposes of this PR but you may still review the CSS code here.

Please throughly review that all the options on storybook work correctly as we've had a number of problems with this component. Also mention any edge cases I may have missed.

## :bulb: Context

Root layout broke on doc site after recent changes made, needed to be converted to full grid for a long time anyways. This should make future layout changes much easier.

## :pencil: Checklist

- [x] I have added tests

## :crystal_ball: Next Steps

Notify DP 2.0 team of the changes

### DIFF
--- a/components/root_layout/root_layout.stories.js
+++ b/components/root_layout/root_layout.stories.js
@@ -8,14 +8,15 @@ import {
 import DtRootLayout from './root_layout.vue';
 
 import DtRootLayoutDefaultTemplate from './root_layout_default.story.vue';
+import DtRootLayoutStickyTemplate from './root_layout_sticky.story.vue';
 
 // Default Prop Values
 export const argsData = {
   sidebarPosition: 'left',
-  header: '<div class="d-bgc-purple-200 d-h64 d-h100p">Header</div>',
-  footer: '<div class="d-bgc-gold-200 d-h64 d-h100p">Footer</div>',
+  header: '<div class="d-bgc-purple-200 d-h64">Header</div>',
+  footer: '<div class="d-bgc-gold-200 d-h64">Footer</div>',
   sidebar:
-    '<div class="d-bgc-black-200 d-h100p d-w264"><div>Sidebar item 1</div><div>Sidebar item 2</div><div>Sidebar item 3</div></div>',
+    '<div class="d-bgc-black-200 d-wmn264 d-h100p"><div>Sidebar item 1</div><div>Sidebar item 2</div><div>Sidebar item 3</div></div>',
   default: `Lorem ipsum dolor sit amet, consectetur adipiscing elit. Aliquam dignissim eleifend condimentum.
   Vestibulum euismod leo at finibus mattis. Integer ut dui id ligula tincidunt pellentesque. Vestibulum a ullamcorper
   risus. Ut tristique sapien eget magna lacinia, non interdum lacus malesuada. Proin augue lacus, finibus eget aliquam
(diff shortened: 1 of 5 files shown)


components/root_layout/root_layout.stories.js
```diff
@@ -8,14 +8,15 @@ import {
...
 // Default Prop Values
 export const argsData = {
   sidebarPosition: 'left',
-  header: '<div class="d-bgc-purple-200 d-h64 d-h100p">Header</div>',
-  footer: '<div class="d-bgc-gold-200 d-h64 d-h100p">Footer</div>',
+  header: '<div class="d-bgc-purple-200 d-h64">Header</div>',
+  footer: '<div class="d-bgc-gold-200 d-h64">Footer</div>',
   sidebar:
-    '<div class="d-bgc-black-200 d-h100p d-w264"><div>Sidebar item 1</div><div>Sidebar item 2</div><div>Sidebar item 3</div></div>',
+    '<div class="d-bgc-black-200 d-wmn264 d-h100p"><div>Sidebar item 1</div><div>Sidebar item 2</div><div>Sidebar item 3</div></div>',
```
juliodialpad, line 19: `d-h100p` shouldn't be necessary here anymore, right?

components/root_layout/root_layout.vue
```diff
@@ -161,17 +155,103 @@ export default {
...
+  &__sidebar {
+    grid-area: sidebar;
+    height: 100%;
+    box-shadow: none;
+
+    &--sticky {
+      position: sticky;
+      top: 0;
+      overflow: auto;
+    }
```
juliodialpad: I cannot find the usage of this class

components/root_layout/root_layout_sticky.story.vue
```diff
@@ -1,35 +1,26 @@
...
-    :header-height="headerHeight"
     :content-class="contentClass"
-    :content-wrap-width-percent="contentWrapWidthPercent"
     :sidebar-class="sidebarClass"
     :sidebar-position="sidebarPosition"
-    :sidebar-width="sidebarWidth"
     :footer-class="footerClass"
-    :footer-height="footerHeight"
-    :fixed="fixed"
+    :fixed="false"
```
juliodialpad: why are we setting this variable, it'd be strange that you toggle it on the panel and it doesn't do anything.

juliodialpad (commented):
Looking pretty good!

Just have some questions, but seems to be working pretty flawlessly.

components/root_layout/root_layout.stories.js
braddialpad: it's basically just for the example. The sidebar itself takes up the full height whether you have this or not, but this element within the sidebar will not take up the full height unless you tell it to. Basically just using it to set the whole sidebar to a color.

components/root_layout/root_layout.vue
braddialpad: hmmm was the sidebar supposed to stay in place when sticky is enabled? can't even remember what it did before

components/root_layout/root_layout_sticky.story.vue
braddialpad: ah yeah thanks, was debugging something and forgot to change it back.

components/root_layout/root_layout.vue
braddialpad: I checked an earlier version and seemed it does not. So I think the current functionality is fine and this class is not needed anymore.

braddialpad: Removed.

iropolo (approved):
Thanks

PR #1255, opened 2023-10-12
fix: suggestion list CSS changes

sanjeevdialpad:
# fix: suggestion list CSS changes

Some changes suggested by design on review


components/rich_text_editor/extensions/suggestion/SuggestionList.vue
```diff
@@ -89,10 +91,26 @@ export default {
...
     upHandler () {
       this.selectedIndex = ((this.selectedIndex + this.items.length) - 1) % this.items.length;
+
+      const activeElement = this.$refs.suggestionList.querySelector('.dt-list-item--highlighted');
+      if (activeElement) {
+        activeElement.scrollIntoView({
+          behaviour: 'smooth',
+          block: 'center',
+        });
+      }
```
braddialpad: Instead of repeating in both functions, isolate in a single function `scrollActiveElementIntoView` or somethin like that.

components/rich_text_editor/extensions/suggestion/SuggestionList.vue
```diff
@@ -24,6 +25,7 @@
   </div>
 </template>
 
+import type { resolveFocusPosition } from '@tiptap/vue-3';
```
braddialpad: does this need to be used?

components/rich_text_editor/extensions/suggestion/SuggestionList.vue
```diff
@@ -2,14 +2,15 @@
   <div class="d-popover__dialog">
     <ul
       v-show="items.length"
+      ref="suggestionList"
       class="dt-suggestion-list"
     >
       <dt-list-item
         v-for="(item, index) in items"
         :key="index"
```
braddialpad: Don't use index for key here it's actually causing a warning to be thrown in vue. Need to use some string that uniquely identifies the item.

components/rich_text_editor/extensions/suggestion/SuggestionList.vue
braddialpad, line 1: Hmm seems like the suggestion list is not working anymore when you type @? Actually doesn't seem related to this PR at all it's happening on vue.dialpad.design/vue3 as well. Pretty strange error...

Screenshot 2023-10-16 at 10 10 29 AM

braddialpad (changes requested):

components/rich_text_editor/extensions/suggestion/SuggestionList.vue
sanjeevdialpad: ah my bad, didnt notice this sneak in , vs code autocomplete :(

components/rich_text_editor/extensions/suggestion/SuggestionList.vue
sanjeevdialpad: the @ works only on the third storybook with channel mentions and its working. Let me see recent changes to check if this is introduced recently. i cant find anything in stacktrace

sanjeevdialpad: damn its because of this comment on the docstring:
```
The only required key is the items function which is used to query the contacts for suggestion.
     * items({ query }) => { return [<ContactObj>]; }
     * ContactObj format:
     * { name: string, avatarSrc: string, contactId: string }
```

sanjeevdialpad: fixed it

braddialpad: Not working once it's deployed. Getting this:

```
[Vue warn]: Avoid app logic that relies on enumerating keys on a component instance. The keys will be empty in production mode to avoid performance overhead.
```

sanjeevdialpad: i think its another unrelated issue, taking a look

components/rich_text_editor/extensions/suggestion/SuggestionList.vue
braddialpad: Fix was not exactly what I meant. Use contactId for the key instead of index.

components/rich_text_editor/extensions/suggestion/SuggestionList.vue
braddialpad: Does the @ list show up for you though?

sanjeevdialpad: Yes in the storybook called with mention suggestion. Type @T and it will show up.

braddialpad: ohh got ya. Didn't know I needed to type "t"

braddialpad: We should put a string above the component in the example saying how to do this, because nobody is going to know they have to put "t" if we don't tell them.

sanjeevdialpad: LEt me change the filter function to return all the list when its empty. That would be better

components/rich_text_editor/extensions/suggestion/SuggestionList.vue
sanjeevdialpad: the problem is that this component is used for contact suggestion, emoji suggestion and in future channel suggestion. So probab;y have to pass a unique index ?

components/rich_text_editor/extensions/suggestion/SuggestionList.vue
braddialpad: Yeah thats fine

components/rich_text_editor/extensions/suggestion/SuggestionList.vue
sanjeevdialpad: btw the warning you talking about is this:
```
[Vue warn]: Avoid app logic that relies on enumerating keys on a component instance. The keys will be empty in production mode to avoid performance overhead.
```

this is because of listening to events on blur/focus/update and this has been happening since before i took this component. 
```
/**
     * The Editor exposes event hooks that we have to map our emits into. See
     * https://tiptap.dev/api/events for all events.
     */
    addEditorListeners () {
      // The content has changed.
      this.editor.on('update', () => {
        const value = this.getOutput();
        this.$emit('input', value);
        this.$emit('update:modelValue', value);
      });

      // The editor is focused.
      this.editor.on('focus', ({ event }) => {
        this.$emit('focus', event);
      });

      // The editor isn’t focused anymore.
      this.editor.on('blur', ({ event }) => {
        this.$emit('blur', event);
      });
    },
```

braddialpad: Then make it a generic id field that all of these would use

braddialpad: The problem with using index, is it changes if the order of the item in the list changes, then things get all kinds of messed up with reactivity and such, so it's best practice to not use it.

braddialpad: > this is because of listening to events on blur/focus/update and this has been happening since before i took this component.

All good, not relevant to this PR then.

sanjeevdialpad: So i used item.id and i changed the emoji implementation to add a atrribute of id. Unfortunately we dont have a unique id for emoji, so i made id = unicode character which will be unique in either way. Now key can just refer item.id.

braddialpad: yeah thats fine, id just has to be SOMETHING unique even if it's a unicode char

braddialpad (approved):

PR #1256, opened 2023-10-14
feat(dt-tooltip-directive): add dt-tooltip directive

juliodialpad:
# Feat (Tooltip Directive): Add dt-tooltip-directive.

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

### Tooltip component
- Added `externalAnchor` prop to be able to handle the tooltip from external anchors.

### Tooltip directive
- Created `dt-tooltip-directive` using `dt-tooltip` component.
- The directive creates a new Vue instance that will manage the dt-tooltip instances
- The directives are exported from a separated package `directives`.
- The directive is exported as `DtTooltipDirective` to avoid naming conflicts with `DtTooltip` component but gets installed with `dt-tooltip` on the Vue instance.

## :bulb: Context

To use `dt-tooltip` component it was needed to add wrappers and a lot of boilerplate around the anchor element, even for just text tooltip.

## :pencil: Checklist

- [x] I have added tests

## :crystal_ball: Next Steps

- Export to Vue 3

### Tooltip component
Screenshot 2023-10-13 at 7 53 02 p m

### Tooltip directive
Screenshot 2023-10-13 at 7 51 52 p m
Screenshot 2023-10-13 at 7 52 04 p m
Screenshot 2023-10-13 at 7 52 12 p m

## :link: Sources

- https://v2.vuejs.org/v2/guide/custom-directive
- https://v2.vuejs.org/v2/guide/render-function


directives/tooltip/tooltip.js
braddialpad, line 1: Hmm curious if storybook even has a built in way of documenting directives? If not it's not a big deal, we can just document manually via MDX.

package.json
```diff
@@ -172,6 +172,16 @@
         "default": "./dist/emoji.cjs"
       }
     },
+    "./directives": {
+      "import": {
+        "types": "./dist/types/directives.d.ts",
+        "default": "./dist/directives.js"
```
braddialpad, line 178: This would have to be manually imported if you want to use dialtone directives right? If so we should add as part of the initial setup in readme.md

braddialpad (commented):
Looks pretty good I think. We'll have to be careful about changing events and prop validators on tooltip since the changes will also need to be done here.

directives/tooltip/tooltip.js
juliodialpad: Looks like it doesn't have any way of documenting directives, we'll need to go with manual MDX documentation.

package.json
juliodialpad: Yes, it should be manually imported, I'll add that to the README.md

directives/tooltip/tooltip.stories.js
```diff
@@ -0,0 +1,24 @@
...
+export const argTypesData = {};
+
+// Story Collection
+export default {
+  title: 'Directives/Tooltip',
+  component: '<div></div>',
+  args: argsData,
+  argTypes: argTypesData,
+  excludeStories: /.*Data$/,
+};
```
braddialpad, line 15: We can disable controls and actions for this story so only the accessibility panel shows.

directives/tooltip/tooltip.test.js
```diff
@@ -0,0 +1,117 @@
...
+  let anchor;
+
+  const updateWrapper = () => {
+    wrapper = mount(WrapperComponent, {
+      localVue: testContext.localVue,
+      propsData: { ...baseProps, ...mockProps },
+      stubs: {
+        // this gets around transition async problems. See https://v1.test-utils.vuejs.org/guides/common-tips.html
+        transition: MOCK_TRANSITION_STUB(),
+      },
```
braddialpad, line 52: should this be in a ...baseStubs, ...mockStubs type thing like the others?? I'm actually not sure.

directives/tooltip/tooltip_directive_default.story.vue
```diff
@@ -0,0 +1,28 @@
...
+    direction="row"
+    gap="600"
+    class="d-mt64"
+  >
+    <dt-button
+      v-dt-tooltip:top-end="'Tooltip text on top-end'"
+      importance="clear"
+      kind="danger"
+    >
+      Trying complex examples
```
braddialpad, line 12: what does "trying complex examples" mean exactly?

braddialpad (approved):
Couple small things. Pre-approving as this looks good to release to me.

braddialpad:
Couple of percy comments I've mentioned as well.

directives/tooltip/tooltip.test.js
juliodialpad: I don't think so, we don't do this stubbing to often so I think it's ok to leave it like this.

directives/tooltip/tooltip_directive_default.story.vue
juliodialpad: That was a leftover, as I was copying the vnodes, I was trying a "complex" example of having children, properties and several stuff but now that's not important as I'm not longer cloning vnodes.

Will change wording there.

braddialpad:
Good to merge thanks!

PR #1261, opened 2023-10-18
feat(call-pill): add default toggled state

fabiangarga:
# PR Title

## :hammer_and_wrench: Feat

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Add a default behavior to control the default state of the collapsible component

## :bulb: Context

For some of our implementation, there is a need to render the call pill with the default state of the collapsible component being open

## :pencil: Checklist

- [x] I have added tests

https://github.com/dialpad/dialtone-vue/assets/5950950/93a0298c-faaa-4205-9b2c-38220467d77c


recipes/conversation_view/feed_pill/feed_item_pill.test.js
```diff
@@ -1,6 +1,6 @@
 import { mount } from '@vue/test-utils';
 import DtRecipeFeedItemPill from './feed_item_pill.vue';
-import { beforeEach } from 'vitest';
+import { beforeEach, describe } from 'vitest';
```
fabiangarga: Not sure if `describe` or `beforeEach` is actually needed, noticed that runs ok without it.

recipes/conversation_view/feed_pill/feed_item_pill.vue
```diff
@@ -18,7 +18,7 @@
               <template #left>
                 <dt-icon
                   size="300"
-                  class="dt-feed-item-pill--icon d-pr8"
+                  class="dt-feed-item-pill--icon d-pr8 d-box-content"
```
fabiangarga: Might need to handle this in a different way, but `box-content` is required to avoid the icon shrinking when external usage has a `normalize` library that's set's pretty much everything to `border-box` that ends up in:

image

recipes/conversation_view/feed_pill/feed_item_pill.vue
```diff
@@ -103,6 +103,11 @@ export default {
       default: () => true,
     },
 
+    defaultToggled: {
```
fabiangarga, line 108: Open for feedback for a better naming, `initialyToggled`?

juliodialpad: You can use `isToggled` use the prefix convention `is`, `has`, `can` when you're defining boolean values.

recipes/conversation_view/feed_pill/feed_item_pill.vue
juliodialpad: We're no longer using utility classes at component level, just for consumer side.

You might want to add that to `.dt-feed-item-pill--icon` defined below.

That image is after or before applying `box-content`?

recipes/conversation_view/feed_pill/feed_item_pill.mdx
```diff
@@ -18,6 +18,16 @@ import * as FeedItemPillStories from './feed_item_pill.stories';
 
 ## Variants
 
+Can enable the initial toggled state
```
juliodialpad:
```suggestion
Set the initial toggled state
```

recipes/conversation_view/feed_pill/feed_item_pill.mdx
```diff
@@ -18,6 +18,16 @@ import * as FeedItemPillStories from './feed_item_pill.stories';
...
+<Canvas>
+  <Story of={FeedItemPillStories.DefaultToggledVariant} />
+</Canvas>
+
+Can enable the initial toggled state, and prevent to be toggled
```
juliodialpad:
```suggestion
Set the initial toggled state, and prevents toggle.
```

recipes/conversation_view/feed_pill/feed_item_pill.stories.js
```diff
@@ -73,6 +73,31 @@ export const Default = {
...
+  render: DefaultTemplate,
+  parameters: { a11y: { disable: true } },
+  args: {
+    iconName: 'video',
+    title: 'This meeting has ended',
+    wrapperClass: 'd-w628',
+    toggleable: false,
+    defaultToggled: true,
+  },
+};
```
juliodialpad: Please avoid duplicating code, if you're using the same args for multiple histories, extract them into a constant and just override the needed properties.

Cleanup the objects below if you can please, they're all using the same args. check other `.stories.js` files for guidance on the best way to do that.

recipes/conversation_view/feed_pill/feed_item_pill.test.js
juliodialpad: They're not needed, please remove them.

recipes/conversation_view/feed_pill/feed_item_pill.test.js
```diff
@@ -14,6 +14,7 @@ describe('DtRecipeFeedItemPill Tests', function () {
     ariaLabel: MOCK_ARIA_LABEL,
     buttonClass: '',
     toggleable: true,
+    defaultToggled: false,
```
juliodialpad: As this is the default value, you don't need to set it here.

recipes/conversation_view/feed_pill/feed_item_pill.test.js
```diff
@@ -113,5 +114,31 @@ describe('DtRecipeFeedItemPill Tests', function () {
         expect(icon.attributes('data-name')).toBe('Chevron Right');
       });
     });
+
+    describe('Default toggled state close', function () {
+      beforeAll(() => {
+        mockProps = {
+          defaultToggled: false,
```
juliodialpad: No need to mock a default value

juliodialpad (changes requested):
Please attend comments and upload the media on PR description again.

Screenshot 2023-10-18 at 1 25 02 p m

recipes/conversation_view/feed_pill/feed_item_pill.vue
fabiangarga: Without applying the `box-content` in some apps we normalize with `* { box-sizing: border-box }` so the `svg` will have that as a value, and causes the icon to shrink.

Didn't know that we are not using utlity classes at the component level. Does this mean that `d-pr8` which is a util class should also be removed?

recipes/conversation_view/feed_pill/feed_item_pill.vue
fabiangarga: Yes, my doubt with `isToggled` is that it kind of also infers the meaning that can be toggled through that, like we do with the `open` property from the `collapsible` component

fabiangarga:
@juliodialpad thanks for the feedback, changes are good to review.
@alexdialpad are the changes ok from your side 🙏?

fabiangarga:
Updated the code once again, the last commit https://github.com/dialpad/dialtone-vue/pull/1261/commits/ba8738c93a99e21a265ff95373892adc590b76fa added a super small change but is disruptive with the current convention of rendering the default component at the beginning of the doc. Since it is closer to the default values that we use in the template which are the ones reactive with the control.

recipes/conversation_view/feed_pill/feed_item_pill.vue
juliodialpad: Yes, please remove `d-pr8`, we're trying to not introduce any new utility class and we're planning to make a cleanup of every component soon.

recipes/conversation_view/feed_pill/feed_item_pill.vue
juliodialpad: Hmm I understand, so `defaultExpanded` might work I think

recipes/conversation_view/feed_pill/feed_item_pill.mdx
```diff
@@ -38,6 +32,12 @@ Can disable the toggleable behaviour (Cursor is reset, no changing of icon on ho
   <Story of={FeedItemPillStories.NoToggleVariant} />
 </Canvas>
 
+## Base Style
+
+<Canvas>
+  <Story of={FeedItemPillStories.Default} />
+</Canvas>
+
```
juliodialpad: Base style should always be the first example

recipes/conversation_view/feed_pill/feed_item_pill.stories.js
```diff
@@ -61,66 +65,38 @@ const DefaultTemplate = (args, { argTypes }) => createTemplateFromVueFile(
...
-  args: {
-    iconName: 'video',
-    title: 'This meeting has ended',
-    ariaLabel: 'Click to expand',
-    buttonClass: '',
-    toggleable: true,
-    wrapperClass: 'd-w628',
-    borderColor: 'default',
-  },
+  parameters: { options: { showPanel: false }, a11y: { disable: true }, controls: { disable: true } },
```
juliodialpad: You should not disable the controls panel on default story.

```suggestion
  parameters: { a11y: { disable: true } },
```

recipes/conversation_view/feed_pill/feed_item_pill.stories.js
juliodialpad, line 1: I think I'd be better if you make just one `feed_item_pill_variants.story.vue` template instead of exporting multiple stories depending on default template.

That helps us while visual testing as we just need to take one snapshot and we can see every component variant.

Please refer to: [avatar.stories.js](https://github.com/dialpad/dialtone-vue/blob/staging/components/avatar/avatar.stories.js), [avatar_variants.story.js](https://github.com/dialpad/dialtone-vue/blob/staging/components/avatar/avatar_variants.story.vue) and [avatar.mdx](https://github.com/dialpad/dialtone-vue/blob/staging/components/avatar/avatar.mdx) to get a reference on how should component documentation be written please.

recipes/conversation_view/feed_pill/feed_item_pill_constants.js
```diff
@@ -1,7 +1,12 @@
...
-  ai: 'dt-feed-item-pill--border-ai',
-  critical: 'dt-feed-item-pill--border-critical',
+  default: 'dt-recipe-feed-item-pill--border-default',
+  ai: 'dt-recipe-feed-item-pill--border-ai',
+  critical: 'dt-recipe-feed-item-pill--border-critical',
+};
+
+export const FEED_ITEM_PILL_DATA_QA = {
+  PILL: 'dt-recipe-feed-item-pill',
+  PILL_ICON: 'dt-recipe-feed-item-pill-icon',
```
juliodialpad: Don't export constants that are only used for QA, you can just put the content right in place.

recipes/conversation_view/feed_pill/feed_item_pill.vue
```diff
@@ -1,25 +1,25 @@
...
-    <div class="d-p8 d-bgc-secondary">
+  <div :class="['dt-recipe-feed-item-pill--border', borderClass, wrapperClass]">
+    <div class="dt-recipe-feed-item-pill--wrapper">
       <dt-collapsible :open="expanded">
         <template #anchor>
           <button
             :aria-label="ariaLabel"
-            data-qa="dt-feed-item-pill"
-            :class="['d-baw0 d-bgc-moderate d-bar-pill d-w100p d-ta-left d-btn--circle', toggleableClass, buttonClass]"
+            :data-qa="DATA_QA.PILL"
```
juliodialpad: We prefer to put the `data-qa` attributes as text don't need to use constants.

recipes/conversation_view/feed_pill/feed_item_pill.vue
```diff
@@ -1,25 +1,25 @@
 <template>
-  <div :class="['dt-feed-item-pill--border', borderClass, wrapperClass]">
-    <div class="d-p8 d-bgc-secondary">
+  <div :class="['dt-recipe-feed-item-pill--border', borderClass, wrapperClass]">
+    <div class="dt-recipe-feed-item-pill--wrapper">
```
juliodialpad:
```suggestion
    <div class="dt-recipe-feed-item-pill__wrapper">
```

recipes/conversation_view/feed_pill/feed_item_pill.vue
```diff
@@ -1,25 +1,25 @@
...
       <dt-collapsible :open="expanded">
         <template #anchor>
           <button
             :aria-label="ariaLabel"
-            data-qa="dt-feed-item-pill"
-            :class="['d-baw0 d-bgc-moderate d-bar-pill d-w100p d-ta-left d-btn--circle', toggleableClass, buttonClass]"
+            :data-qa="DATA_QA.PILL"
+            :class="['dt-recipe-feed-item-pill--button', toggleableClass, buttonClass]"
```
juliodialpad:
```suggestion
            :class="['dt-recipe-feed-item-pill__button', toggleableClass, buttonClass]"
```

recipes/conversation_view/feed_pill/feed_item_pill.test.js
juliodialpad, line 1: Remove constants please, that's not part of our standards.

recipes/conversation_view/feed_pill/feed_item_pill.vue
```diff
@@ -1,25 +1,25 @@
...
           >
-            <dt-item-layout class="d-w100p d-p8">
+            <dt-item-layout class="dt-recipe-feed-item-pill--layout">
               <span class="d-fw-bold">{{ title }}</span>
               <template #left>
                 <dt-icon
                   size="300"
-                  class="dt-feed-item-pill--icon d-pr8 d-box-content"
-                  data-qa="dt-feed-item-pill-icon"
+                  class="dt-recipe-feed-item-pill--icon"
```
juliodialpad:
```suggestion
                  class="dt-recipe-feed-item-pill__icon"
```

recipes/conversation_view/feed_pill/feed_item_pill.vue
```diff
@@ -1,25 +1,25 @@
...
+            :data-qa="DATA_QA.PILL"
+            :class="['dt-recipe-feed-item-pill--button', toggleableClass, buttonClass]"
             @focusin="hover = true"
             @focusout="hover = false"
             @mouseenter="hover = true"
             @mouseleave="hover = false"
             @click="onClick"
           >
-            <dt-item-layout class="d-w100p d-p8">
+            <dt-item-layout class="dt-recipe-feed-item-pill--layout">
```
juliodialpad:
```suggestion
            <dt-item-layout class="dt-recipe-feed-item-pill__layout">
```

recipes/conversation_view/feed_pill/feed_item_pill.vue
```diff
@@ -36,7 +36,7 @@
           </button>
         </template>
         <template #content>
-          <div class="d-jc-center d-d-flex">
+          <div class="dt-recipe-feed-item-pill--content d-jc-center d-d-flex">
```
juliodialpad:
```suggestion
          <div class="dt-recipe-feed-item-pill__content">
```

juliodialpad (changes requested):
Gave it another review, please attend comments.

*note*: We're using [Block Element Modifier](https://getbem.com/naming/) naming convention to name our classes.

recipes/conversation_view/feed_pill/feed_item_pill.mdx
fabiangarga: I feel like the control should be next before the variants? otherwise it feels odd to be "controlling" the props and having to scroll to see the changes in the base style.

recipes/conversation_view/feed_pill/feed_item_pill.stories.js
fabiangarga: Got it, definitely this looks better, thanks. Updated the code.

fabiangarga:
@juliodialpad thanks for the feedback, updated the branch.

recipes/conversation_view/feed_pill/feed_item_pill.mdx
juliodialpad: You're not supposed to use controls to see changes on documentation, that's what the playground is for. Having the props there is so you can read abouth props and what they're meant for.

recipes/conversation_view/feed_pill/feed_item_pill.stories.js
```diff
@@ -62,42 +64,28 @@ const DefaultTemplate = (args, { argTypes }) => createTemplateFromVueFile(
...
-  render: DefaultTemplate,
-  parameters: Default.parameters,
-  args: {
-    borderColor: 'ai',
+  parameters: {
+    options: { showPanel: false },
+    a11y: { disable: true },
+    percy: {
+      args: { seed: 'seed' },
+    },
```
juliodialpad: - The default story should always show the panel, just hide it on the variants story.
- The percy argument `seed` is just for when we're using avatars, so they don't randomly change the color on every visual test, given that, you don't need it on this story.

```suggestion
  parameters: {
    a11y: { disable: true }
  }
```

recipes/conversation_view/feed_pill/feed_item_pill.stories.js
```diff
@@ -62,42 +64,28 @@ const DefaultTemplate = (args, { argTypes }) => createTemplateFromVueFile(
...
-  render: DefaultTemplate,
-  parameters: Default.parameters,
-  args: {
-    toggleable: false,
-    borderColor: 'critical',
+export const Variants = {
+  render: VariantsTemplate,
+  parameters: {
+    ...Default.parameters,
+    controls: { disable: true },
```
juliodialpad: Disable the panel and the controls on variants story.

```suggestion
   parameters: { 
     options: { showPanel: false }, 
     a11y: { disable: true }, 
     controls: { disable: true },
```

juliodialpad (changes requested):
Previously you were showing `DefaultToggledVariant` and `DefaultToggledVariantNonToggleable` and now I cannot see default toggled examples on variants, I think you missed to add those examples, right?

fabiangarga:
> Previously you were showing `DefaultToggledVariant` and `DefaultToggledVariantNonToggleable` and now I cannot see default toggled examples on variants, I think you missed to add those examples, right?

@juliodialpad Correct, due to the conversation here https://dialpad.slack.com/archives/CEDUKPKC1/p1697845368879809
I changed the variants to be more focused in the way we use it in the DP app, by providing some examples of the call/meeting with recaps, which make use of the toggled props

recipes/conversation_view/feed_pill/feed_item_pill.stories.js
fabiangarga: Oh makes sense, thought it was added to allow percy to run the visual test for the component, I was not able to access percy so didn't know that. Thanks.

recipes/conversation_view/feed_pill/feed_item_pill.vue
```diff
@@ -1,25 +1,27 @@
...
             @focusin="hover = true"
             @focusout="hover = false"
             @mouseenter="hover = true"
             @mouseleave="hover = false"
             @click="onClick"
           >
-            <dt-item-layout class="d-w100p d-p8">
-              <span class="d-fw-bold">{{ title }}</span>
+            <dt-item-layout class="dt-recipe-feed-item-pill__layout">
+              <slot name="title">
```
fabiangarga, line 17: @AlexLamDialpad added this so we can use the slot, in case the title needs to be handled differently.

juliodialpad (approved):
Looking good, thanks!

Please update the branch, add the 'visual-test-ready' tag so visual tests run so I can review and approve them before merging.

PR #1266, opened 2023-10-20
feat: add avatar slot to feed-item-row

braddialpad:
# feat: add avatar slot to feed-item-row

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

There is a ton of product side logic that is used to correctly display an avatar currently, so it is best that we are able to pass in a slot here. 

Already tested in product via `npm link`

## :bulb: Context

A lot of display issues with avatar on feed item row since we could not implement all the product side logic into a slot.

## :pencil: Checklist

## :crystal_ball: Next Steps

Update in ubervoice for feed item row changes.

### DIFF
--- a/recipes/conversation_view/feed_item_row/feed_item_row_default.story.vue
+++ b/recipes/conversation_view/feed_item_row/feed_item_row_default.story.vue
@@ -14,9 +14,11 @@
       <template v-if="defaultSlot">
         <span v-html="defaultSlot" />
       </template>
-      <template #avatar>
+      <template 
+        v-if="avatar"
+        #avatar
+      >
         <span
-          v-if="avatar"
           v-html="avatar"
         />
       </template>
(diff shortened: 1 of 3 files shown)


recipes/conversation_view/feed_item_row/feed_item_row_default.story.vue
```diff
@@ -14,6 +14,12 @@
       <template v-if="defaultSlot">
         <span v-html="defaultSlot" />
       </template>
+      <template #avatar>
+        <span
+          v-if="avatar"
+          v-html="avatar"
+        />
+      </template>
```
juliodialpad: Move the v-if to the template to maintain the same logic in all of them.
```suggestion
      <template 
        v-if="avatar"
        #avatar
      >
        <span
          v-html="avatar"
        />
      </template>
```

juliodialpad (approved):
Looks good, just a minor code formatting

PR #1267, opened 2023-10-20
feat: add avatar slot to feed-item-row

braddialpad:
Vue 3 of https://github.com/dialpad/dialtone-vue/pull/1266


juliodialpad (approved):
Same as vue 2 version, just move the v-if to template before merging please

PR #1268, opened 2023-10-20
fix(button): icon sizing

juliodialpad:
# Fix (Button): Icon sizing

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

- Added slot scope data to return the icon size that should be set according to button sizing

## :bulb: Context

Got reports of inconsistent icon sizing on buttons.

## :pencil: Checklist

- [] I have added tests
- [] I have validated components with a screen reader
- [] I have validated components keyboard navigation
- [] I have considered the performance impact of my change
- [] I have checked that my change did not significantly increase bundle size
- [] I am exporting any new components or constants in the index.js in the component directory
- [] I am exporting any new components or constants in the index.js in the root

## :crystal_ball: Next Steps

A migration shouldn't be necessary as the current functionality is maintained and only introduced a way to make sure the icon is congruent with button size.

In case we would like to perform a migration, it shouldn't be too hard as we just need to replace the icon `size` value with the bound `iconSize` and add the slot scope. There are definitely a bunch of this usages but it'd easy to search/replace everything at once.

Screenshot 2023-10-20 at 1 18 44 p m

Can you guys make sure this are the correct icon-button size relations please. 
@francisrupert @yorbi-dp @fede-dp


braddialpad (approved):
Should work well enough

PR #1272, opened 2023-10-23
fix(dt-tooltip-directive): default delay

juliodialpad:
# Fix (Tooltip Directive): Default delay

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

- Moved the external anchor event listener bindings to the tooltip component instead of the directive. This way, the directives and external anchor events are handled directly by the tooltip component so we avoid duplicating code. Also, we can remove the need to handle the events manually.

## :bulb: Context

Tooltips through tooltip directive were not using the default delay.

## :pencil: Checklist

- [ ] I have added tests

### DIFF
--- a/components/tooltip/tooltip.vue
+++ b/components/tooltip/tooltip.vue
@@ -295,7 +295,7 @@ export default {
     },
 
     anchor () {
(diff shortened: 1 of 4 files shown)


braddialpad (approved):
Really clever solution, I like it. Some failing unit tests, probably event related.

PR #1276, opened 2023-10-24
fix: add disabled prop to tooltip

sanjeevdialpad:
# fix: add disabled prop to tooltip

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

this disabled prop will completely ignore whats set on show prop and make sure the tooltip doesnt appear. The general purpose of this disabled prop should be to preserve the hover actions on show prop but to be able to disable tooltip in certain conditions.

### DIFF
--- a/components/tooltip/tooltip.vue
+++ b/components/tooltip/tooltip.vue
@@ -193,6 +193,7 @@ export default {
     /**
      * Controls whether hover/focus causes the tooltip to appear.
      * Cannot be combined with the show prop. show value will be ignored.
+     * by default this is true, if you override with false, the tooltip will never show up.
      */
     enabled: {
       type: Boolean,
@@ -342,6 +343,10 @@ export default {
   },
 
   mounted () {
+    if (!this.enabled && this.show != null) {
+      console.warn('Tooltip: You cannot use both the enabled and show props at the same time.');
+      console.warn('The show prop will be ignored.');
+    }
     this.externalAnchor && this.addExternalAnchorEventListeners();
     this.tip = createTippy(this.anchor, this.initOptions());
 
--- a/components/tooltip/tooltip.vue
+++ b/components/tooltip/tooltip.vue
@@ -21,7 +21,7 @@
     <dt-lazy-show
       :id="id"
       ref="content"
-      :show="isShown && enabled && (!!message.trim() || !!$slots.default)"
+      :show="isVisible"
       role="tooltip"
       aria-hidden="false"
       data-qa="dt-tooltip"
@@ -275,6 +275,11 @@ export default {
   },
 
   computed: {
+    // whether the tooltip is visible or not.
+    isVisible () {
+      return this.isShown && this.enabled && (!!this.message.trim() || !!this.$slots.default);
+    },
+
     tooltipListeners () {
       return {
         ...this.$listeners,
--- a/components/tooltip/tooltip.vue
+++ b/components/tooltip/tooltip.vue
@@ -21,7 +21,7 @@
     <dt-lazy-show
       :id="id"
       ref="content"
-      :show="isShown && !disabled && (!!message.trim() || !!$slots.default)"
+      :show="isShown && enabled && (!!message.trim() || !!$slots.default)"
       role="tooltip"
       aria-hidden="false"
       data-qa="dt-tooltip"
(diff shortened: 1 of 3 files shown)


components/tooltip/tooltip.vue
```diff
@@ -190,6 +190,15 @@ export default {
...
     },
 
+    /**
+     * Controls whether hover/focus causes the tooltip to appear.
+     * Cannot be combined with the show prop. show value will be ignored.
+     */
+    disabled: {
+      type: Boolean,
+      default: false,
+    },
```
braddialpad: hmm maybe not a great name since "disabled" is a valid html attribute on a lot of elements. Maybe we make it "enabled" defaulting to true instead.

components/tooltip/tooltip.vue
```diff
@@ -343,6 +352,13 @@ export default {
     }
   },
 
+  beforeCreate () {
+    if (this.$props.disabled != null && this.$props.show != null) {
+      console.warn('Tooltip: You cannot use both the disabled and show props at the same time.');
+      console.warn('The show prop will be ignored.');
+    }
+  },
```
braddialpad: This is not reactive, if you change the prop after initializing it will not trigger. see `badge.vue` validateProps for an example of how to do it reactively.

sanjeevdialpad: actually for vue2, i cannot even access props in beforeCreate. vue3 it works

components/tooltip/tooltip.vue
sanjeevdialpad: changed.

components/tooltip/tooltip.vue
```diff
@@ -21,7 +21,7 @@
     <dt-lazy-show
       :id="id"
       ref="content"
-      :show="isShown && (!!message.trim() || !!$slots.default)"
+      :show="isShown && !disabled && (!!message.trim() || !!$slots.default)"
```
iropolo: This definetly need a computed property which evaluate this condition...

The name of this computed property should be descriptable to explain what `isShown && !disabled && (!!message.trim() || !!$slots.default)` is doing...

components/tooltip/tooltip.vue
braddialpad: that's because beforeCreate doesn't exist in vue 2 lol

components/tooltip/tooltip.vue
sanjeevdialpad: actually the whole thing can go in isVisible computed poperty ?

components/tooltip/tooltip.vue
iropolo: Could we move this to the `mounted` hook?
I cannot reproduce this warn on `beforeCreate` probably its not being reach.

components/tooltip/tooltip.vue
iropolo: Sounds good.

I would update how the `show` and `isShown` are being used, to work with the new `!disabled && (!!message.trim() || !!$slots.default)` evaluation. 

As far as I can see there is a watcher for `show` prop and there is a watcher for `isShown` data value.
This `isShown` is used in multple places... its being used inside `show` watcher...., huh its kind of a mess this component lol...

components/tooltip/tooltip.vue
sanjeevdialpad: -_- my bad.

sanjeevdialpad: will do.

components/tooltip/tooltip.vue
sanjeevdialpad: update as in documentation ?

iropolo: `isVisible` will replace the `isShown` ?

sanjeevdialpad: No isVisible is basically the entire condition. I probably did not get what you meant by replacing this ? As in just for readability ?

iropolo: Ye I mean... with this addition we will have `show`, `isShown` and `isVisible`... Im wondering whats going on...

iropolo: Update this PR with the `isVisible` approach, then I will take a look further on this, will try to come up with some solution.

sanjeevdialpad: done but am not sure if thats required since the comments on each prop do mention what they do.

also to note, enabled prop is redundant in a way that if message is not provided it does the same thing. we introduced this prop so it will be explicit on that behaviour.

iropolo (approved):
Looks good, thanks

PR #1277, opened 2023-10-24
fix: add disabled prop to tooltip - vue3

sanjeevdialpad:
# fix: add disabled prop to tooltip vue3

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

this disabled prop will completely ignore whats set on show prop and make sure the tooltip doesnt appear. The general purpose of this disabled prop should be to preserve the hover actions on show prop but to be able to disable tooltip in certain conditions.

### DIFF
--- a/components/tooltip/tooltip.vue
+++ b/components/tooltip/tooltip.vue
@@ -21,7 +21,7 @@
     <dt-lazy-show
       :id="id"
       ref="content"
-      :show="isShown && !disabled && (!!message.trim() || hasSlotContent($slots.default))"
+      :show="isVisible"
       role="tooltip"
       aria-hidden="false"
       data-qa="dt-tooltip"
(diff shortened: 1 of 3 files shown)


components/tooltip/tooltip.vue
```diff
@@ -351,6 +360,13 @@ export default {
     }
   },
 
+  beforeCreate () {
+    if (this.$props.disabled != null && this.$props.show != null) {
+      console.warn('Tooltip: You cannot use both the disabled and show props at the same time.');
+      console.warn('The show prop will be ignored.');
+    }
+  },
```
iropolo: Would move this also to `mounted` as vue2 version

iropolo (approved):
Looks good, thanks

PR #1278, opened 2023-10-24
feat: add initial date formatting functions

braddialpad:
# feat: add initial date formatting functions

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Added some initial date formatting functions:

- getDateMedium which returns the date in medium format from our [Writing Guidelines](https://dialpad.design/guides/writing-guidelines/#formats-by-length)
- callDurationToHumanReadable ([currently in DP 2.0](https://github.com/dialpad/web-clients/pull/613/files)) 
- dateRelativeToNow ([currently in DP 2.0](https://github.com/dialpad/web-clients/blob/main/packages/conversation-feed-ui/src/use-relative-date-format.ts))

The story was just to initially add this getDateMedium function, but I decided to include the functions that DP 2.0 wanted to transfer to dialtone as well so they have something that's actually useful to work with. All public functions here are fully i18n compatible and the locale must be set via `setDateLocale` before using any of these functions.

I've documented these functions well in the code so please see the comments for further details about how they work.

Documentation will be coming in a separate PR since I'd like to publish it on dialtone.

## :bulb: Context

Requested by DP 2.0 team. They are having to put a lot of logic for date formatting into DP 2.0 when really this should be handled on the Dialtone side.

## :pencil: Checklist

- [x] I have added tests
- [ ] I have added all relevant documentation (not yet)

## :crystal_ball: Next Steps

- Write public documentation
- Add more date functions
- See if we can extract or use date functions from our current date picker.

## :link: Sources

- https://dialpad.design/guides/writing-guidelines/#formats-by-length
- https://github.com/dialpad/web-clients/pull/613/files
- https://github.com/dialpad/web-clients/blob/main/packages/conversation-feed-ui/src/use-relative-date-format.ts
- https://date-fns.org/

### DIFF
--- a/common/dates.js
+++ b/common/dates.js
@@ -1,0 +1,127 @@
+import {
+  format,
+  formatDuration,
+  intervalToDuration,
+  formatDistance,
+  isToday,
+  isYesterday,
+  isThisWeek,
+  isSameYear,
+} from 'date-fns';
+import { capitalizeFirstLetter } from './utils';
+
+// Base functions just wrap core date-fns functions, but this allows us to do checks and set default options.
+
+function _baseFormat (date, formatString) {
+  _checkLocaleSet();
+  return format(date, formatString, {
+    locale: global.locale,
+  });
+}
+
+function _baseFormatDuration (duration, formatString) {
+  _checkLocaleSet();
+  return formatDuration(duration, {
+    locale: global.locale,
+    format: formatString,
+  });
+}
+
+function _baseFormatDistance (date, baseDate) {
+  _checkLocaleSet();
+  return formatDistance(date, baseDate, {
+    locale: global.locale,
+  });
+}
+
+function _isLocaleSet () {
+  return global.locale !== undefined;
+}
+
+function _checkLocaleSet () {
+  if (!_isLocaleSet()) {
+    throw new Error('Locale not set, please call setDateLocale(locale) and pass ' +
+    'in a datefns locale object as the locale param before calling this function');
+  }
+}
+
+/**
+ * Sets the locale for date-fns. This should be called before any date-fns functions are called.
+ * @param {Locale} locale A date-fns locale object
+ */
+export function setDateLocale (locale) {
+  global.locale = locale;
+}
+
+/**
+ * This formats a date to the Dialtone standard medium date format as shown here:
+ * https://dialpad.design/guides/writing-guidelines/#formats-by-length
+ * @param {Date} date A javascript date object
+ * @returns {string} A string in the format of 'September 2, 2022'
+ */
+export function getDateMedium (date) {
+  return _baseFormat(date, 'MMMM d, y');
+}
+
+/**
+ * Converts a call duration in total number of seconds to a human readable string
+ * such as 'less than a minute' or '4 hours 34 minutes'.
+ * @param {number} durationInSeconds The duration of the call in seconds
+ * @returns {string} A human readable string representing the duration of the call
+ */
+export function callDurationToHumanReadable (durationInSeconds) {
+  if (durationInSeconds < 60) {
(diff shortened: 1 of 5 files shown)


common/dates.js
```diff
@@ -0,0 +1,127 @@
...
+    const rtl = new Intl.RelativeTimeFormat(global.locale.code, { numeric: 'auto' });
+    return capitalizeFirstLetter(rtl.formatToParts(-1, 'day')[0].value, global.locale.code);
+  } else if (isThisWeek(date)) {
+    return _baseFormat(date, 'EEEE');
+  } else if (isSameYear(date, Date.now())) {
+    return _baseFormat(date, 'EEEE, MMMM d');
+  } else {
+    return _baseFormat(date, 'MMMM d, y');
+  }
+}
```
juliodialpad: I don't like too much this if-else statements, it's not very readable. I'd be better if we can use something more readable, if you cannot think on another solution is fine, it's not that bad, but can be improved.

common/utils.test.js
```diff
@@ -475,4 +476,13 @@ describe('Util Tests', () => {
...
   });
+
+  describe('capitalizeFirstLetter', () => {
+    it('should capitalize the first letter of a string', () => {
+      expect(capitalizeFirstLetter('hello')).toBe('Hello');
+    });
+    it('should capitalize the first letter of a string with dutch locale', () => {
+      expect(capitalizeFirstLetter('ĳsselmeer', 'nl')).toBe('Ĳsselmeer');
+    });
+  });
```
juliodialpad: Can we add some test with some Japanese and RLT languages? just to be sure that this will work with those.

juliodialpad (changes requested):
Good job!, couple questions/improvements.

common/dates.js
```diff
@@ -0,0 +1,127 @@
...
+ * @param {Date} date The timestamp of the item's date
+ * @returns {string} A human readable string representing the distance between the date and now
+ */
+export function dateRelativeToNow (date) {
+  if (isToday(date)) {
+    const rtl = new Intl.RelativeTimeFormat(global.locale.code, { numeric: 'auto' });
+    return capitalizeFirstLetter(rtl.formatToParts(0, 'day')[0].value, global.locale.code);
+  } else if (isYesterday(date)) {
+    const rtl = new Intl.RelativeTimeFormat(global.locale.code, { numeric: 'auto' });
+    return capitalizeFirstLetter(rtl.formatToParts(-1, 'day')[0].value, global.locale.code);
```
AlexLamDialpad: No idea if this is an issue, but do we have to do something special for languages that might not have a capitalization system?

common/utils.test.js
braddialpad: Added Japanese.

It seems there is no existing RTL language that contains capital letters so should be fine.

common/dates.js
braddialpad: If it doesn't, it will just not capitalize the letter. Added Japanese to my tests to test this case.

common/dates.js
braddialpad: yeah.. not sure there is an alternative since these are 5 distinct cases with different formats... I can probably at least improve readability by cutting down some of those long statements.

braddialpad:
Updated function names to be more generic as requested by Alex.

common/dates.js
braddialpad: alright still just as many if statements but it's a bit more readable now. Not sure there's much else I can do :shrug:

AlexLamDialpad (approved):
LGTM!

juliodialpad (approved):
Nice improvements, thanks!

PR #1280, opened 2023-10-25
fix(message-input): tooltip should be displayed when character limit has been reached

sanjeevdialpad:
fix(message-input): Tooltip should be displayed when character limit has been reached

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Tooltip showing the character limit message from product should be displayed on the limit count when its negative.


iropolo:
Hmm I cannot see the tooltip, how I can trigger it? seems like the params below are correct

image

braddialpad:
> Hmm I cannot see the tooltip, how I can trigger it? seems like the params below are correct

It's the red -375 in the bottom right.

braddialpad:
> It's the red -375 in the bottom right.

Nevermind you meant the tooltip not the number.. Yeah I also can't see it.

sanjeevdialpad:
> > Hmm I cannot see the tooltip, how I can trigger it? seems like the params below are correct
> 
> 
> 
> It's the red -375 in the bottom right.
> 
> 
> 
> 

Yes is it not triggering ?

braddialpad:
Oh I see, you actually have to mouseover it, it doesn't show up automatically when you hit the limit

sanjeevdialpad:
Screenshot 2023-10-26 at 9 47 41 AM
It does show up on over. Do know that it will only be enabled when its negative but in your screenshot it is. just hover over it it should show up.

iropolo:
ohh.. huh, ye it is working I just need to hover it, I expected it will be shown when it is reached the limit

image

sanjeevdialpad:
> Oh I see, you actually have to mouseover it, it doesn't show up automatically when you hit the limit

oh no, it supposed to just show on hover when the user is puzzled why send button is disabled. The disable of send button should be from product side.

recipes/conversation_view/message_input/message_input.vue
```diff
@@ -145,13 +145,24 @@
         <!-- Right content -->
         <div class="d-d-flex">
           <!-- Optionally displayed remaining character counter -->
-          <p
+          <dt-tooltip
             v-if="displayCharacterLimitWarning"
+            :enabled="characterLimitWarningMessage && (characterLimitCount - inputLength < 0)"
+            placement="top-end"
+            :message="characterLimitWarningMessage"
             class="d-fc-error d-mr16 d-as-center dt-message-input--remaining-char"
```
braddialpad: Should these classes be on the anchor <p> instead of the tooltip itself?

sanjeevdialpad:
the screenshot is in the ticket showing what the behaviour is: https://dialpad.atlassian.net/browse/DP-81811

braddialpad (commented):
Looks pretty good I think, just one question.

braddialpad:
> oh no, it supposed to just show on hover when the user is puzzled why send button is disabled. The disable of send button should be from product side.

Yeah makes sense, I just didn't understand the functionality initally.

iropolo:
I see it okay..., should it have a tooltip also when the warning appear? I mean this case...

image


Something like "You have {x} characters remaining."

iropolo (approved):
Looks good to me.
Pre-approving it. (Review Brad's comments please)

braddialpad:
> oh no, it supposed to just show on hover when the user is puzzled why send button is disabled. The disable of send button should be from product side.

Not really your issue, more of a design issue, but wouldn't it make sense for the tooltip to show when hovering the send button rather than the red text then? I know if I was a user I'd likely hover the button rather than the red text

recipes/conversation_view/message_input/message_input.vue
sanjeevdialpad: i moved everything from <p> as i had an issue with it showing up. didnt have to move classes. Moved it back.

sanjeevdialpad:
> > oh no, it supposed to just show on hover when the user is puzzled why send button is disabled. The disable of send button should be from product side.
> 
> Not really your issue, more of a design issue, but wouldn't it make sense for the tooltip to show when hovering the send button rather than the red text then? I know if I was a user I'd likely hover the button rather than the red text

I will pass on the feedback. i felt the same way too. but then you can argue a disabled send button should not be focusable.

braddialpad:
> I will pass on the feedback. i felt the same way too. but then you can argue a disabled send button should not be focusable.

It's actually a very common case and we have guidance for it in our docs
https://dialpad.design/components/button.html#disabled

This definitely doesn't need to hold up this PR but yeah it's a good idea to shoot this question at design I think. It feels a bit off.

braddialpad (approved):

PR #1282, opened 2023-10-26
docs: document date functions

braddialpad:
# docs: document date functions

- [ ] Fix
- [ ] Feature
- [ ] Refactoring
- [x] Documentation

## :book: Description

Found a pretty cool way to document functions in storybook. Can use a library to generate .md files and then include them in an MDX to be output in storybook. This could also be used for any other functions we want to export in the future.

## :bulb: Context

Needed an automated way of documenting exported functions that are part of dialtone.

## :pencil: Checklist

Screenshot 2023-10-26 at 4 54 12 PM

## :link: Sources

https://typedoc.org/


iropolo:
~~How it will works? could we test it somehow? maybe with a test PR?.~~

Oh sorry I didnt see the image.
That images is the final result of the documented date function?

braddialpad:
ahh something is messed up with the build because deploy preview failed. Let me see...

braddialpad:
Okay, should be fixed now. Checkout the preview: Functions -> Date and Time

iropolo:
`A human readable string representing the duration of the call` Is this writed by... who?

Is this `Date and time` function has sth related by the functions on datepicker, those functions are also exported independently. Should be here?, or maybe not on `Date and time` but inside of `FUNCTIONS` section.

braddialpad:
> `A human readable string representing the duration of the call` Is this writed by... who?

The language text is handled by the library and is based on the locale set.

> Is this `Date and time` function has sth related by the functions on datepicker, those functions are also exported independently. Should be here?, or maybe not on `Date and time` but inside of `FUNCTIONS` section.

Yeah agreed, I have a future task to integrate datepicker with this. We could probably make use of the setLocale function in datepicker as well.

package.json
```diff
@@ -20,8 +21,8 @@
     "start": "npm run storybook",
     "storybook": "STORYBOOK_ENV=development storybook dev -p 9010",
     "storybook:clean": "rm -rf ./storybook-static",
-    "storybook:build": "npm run storybook:clean && STORYBOOK_ENV=production storybook build && touch ./storybook-static/.nojekyll",
-    "storybook:build:docs-only": "STORYBOOK_ENV=production storybook build --docs",
+    "storybook:build": "npm run build:documentation:functions && npm run storybook:clean && STORYBOOK_ENV=production storybook build && touch ./storybook-static/.nojekyll",
```
iropolo, line 24: Will this grow hideously?

braddialpad: haha if we keep getting new processes that have to be run as part of the build, maybe. Hopefully won't be any more.

iropolo:
> Yeah agreed, I have a future task to integrate datepicker with this. We could probably make use of the setLocale function in datepicker as well.

How many others options we have to document using this?

Im trying to find a considerable reason to use this library and this workflow to improve or ease our work doing documentation.

Im sort of expetic with this addition.

Since we are extending as I see:
`.gitignore`

`.storybook/main.js` (this will only be once, since is just `function` section, but for others sections it will be necessary to add it here)

`build:documentation:functions` this looks out of context since we are running a build, but also a particular section of a particular part of the documentation. Maybe we can just separate this to another script and running in the build process, sth like `npm run typedoc:functions` or just `npm run typedoc:generate` idk.

`typedoc.config.js`
This seems like it will only works for functions, I didnt read the documentation of typedoc but it will probably accept an array here if we want to add more.


As I wrote in the beginning, I will be more open to this idea if we can find more usages on this and not only on `FUNCTIONS` .

Of course it is not necessary to do it all in once, but just to know in the future, Im not sure to do all this (adding a new process) just for a couple of things.

Also there is a big chance where Im not seeing this well and my worries are nonsense.

braddialpad:
> > Yeah agreed, I have a future task to integrate datepicker with this. We could probably make use of the setLocale function in datepicker as well.
> 
> How many others options we have to document using this?

There will likely be many more public utility functions added over the future. It will be a core part of the library, we can maybe move to dialpad.design at some point, but right now this is the simplest way.

> `build:documentation:functions` this looks out of context since we are running a build, but also a particular section of a particular part of the documentation. Maybe we can just separate this to another script and running in the build process, sth like `npm run typedoc:functions` or just `npm run typedoc:generate` idk.

yeah true it's a bit confusing, I will update.

> `typedoc.config.js` This seems like it will only works for functions, I didnt read the documentation of typedoc but it will probably accept an array here if we want to add more.

Yes any additional locations can be added via the "entryPoints" array in typedoc.config.js

> As I wrote in the beginning, I will be more open to this idea if we can find more usages on this and not only on `FUNCTIONS` .

What usage would it have other than functions? Our components are documented via native storybook. The only reason we are using this library to generate documentation is because storybook has no support for documenting native js functions. I guess maybe our constants could be documented with this system as well.. is that what you mean?

iropolo:
>The only reason we are using this library to generate documentation is because storybook has no support for documenting native js functions. I guess maybe our constants could be documented with this system as well.. is that what you mean?

Ohhh, okay... sounds good then. Cool.

Yes I like it, then we can move forward with datepicker functions and constants would be cool.

iropolo (approved):
Looks good.

Would be cool to document the needed steps to generate this documentation.
Just to not miss sth.
Maybe sth like "Document with typedoc".

PR #1289, opened 2023-10-30
feat: add description list component

ninamarina:
# feat: add description list component

Jira ticket: https://dialpad.atlassian.net/browse/DLT-1209
Link to preview: https://vue.dialpad.design/deploy-previews/pr-1289/?path=/story/components-description-list--default

Based on this figma, but it still needs design definition: https://www.figma.com/file/HXPNFvkGbcsfU6KpoW0ZPl/Screen-Pops?type=design&node-id=740-186846&mode=design&t=jxVBVwxW69KpzzZc-0

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Adds the new component description list to Dialtone Vue.

## :pencil: Checklist

- [x] I have added tests

## :crystal_ball: Next Steps
Move the CSS to Dialtone

image

### DIFF
--- a/components/description_list/description_list.stories.js
+++ b/components/description_list/description_list.stories.js
@@ -1,0 +1,123 @@
+import { createTemplateFromVueFile } from '@/common/storybook_utils';
+import DtDescriptionList from './description_list.vue';
+import DtDescriptionListDefaultTemplate from './description_list_default.story.vue';
+
+export const argTypesData = {
+  // Props
+  items: {
+    control: 'object',
+    table: {
+      defaultValue: {
+        summary: '{ term: string, description: string }[]',
+      },
+    },
+  },
+  descriptionClass: {
+    control: 'text',
+  },
+  termClass: {
+    control: 'text',
+  },
+};
+
+export const argsData = {
+  direction: 'row',
+  gap: '400',
+  items: [
+    {
+      term: 'Customer Intent',
+      description: `Hello, I'm looking to return my TV`,
+    },
+    {
+      term: 'Reason',
+      description: 'Refound',
+    },
+    {
+      term: 'Country',
+      description: 'England',
+    },
+    {
+      term: 'Random',
+      description: 'Value',
+    },
+  ],
+};
+
+const argsDataLongText = {
+  items: [
+    {
+      term: 'Customer Intent',
+      description: `Lorem ipsum dolor sit amet, consectetur adipiscing elit,
+      sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.
+      Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.`,
+    },
+    {
+      term: 'Three word term',
+      description: ` Duis aute irure dolor in reprehenderit in voluptate velit
+      esse cillum dolore eu fugiat nulla pariatur. Excepteur sint occaecat
+      cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.`,
+    },
+    {
+      term: 'Country',
+      description: 'England',
+    },
+    {
+      term: 'Random',
+      description: 'Value',
+    },
+  ],
+};
+
+const decorator = () => ({
+  template: `<div
+      style="width: var(--dt-size-925);
+      overflow: hidden;
+      resize: horizontal;
+      height: auto;
+      border: 1px solid var(--dt-color-border-subtle);
+      padding: var(--dt-space-450);
+      borderRadius: var(--dt-size-radius-400);"
+      >
+      <story />
+    </div>`,
+});
+
+// Story Collection
+export default {
+  title: 'Components/Description List',
+  component: DtDescriptionList,
+  args: argsData,
+  argTypes: argTypesData,
+  decorators: [decorator],
+  excludeStories: /.*Data$/,
+};
+
+// Templates
+const DefaultTemplate = (args, { argTypes }) => createTemplateFromVueFile(
+  args,
+  argTypes,
+  DtDescriptionListDefaultTemplate,
+);
+
+// Stories
+export const Default = {
+  render: DefaultTemplate,
+  args: {},
+};
+
+export const ColumnDirection = {
+  ...Default,
+  args: {
+    direction: 'column',
+  },
+};
+
+export const LongText = {
+  ...Default,
+  args: { ...Default.args, items: argsDataLongText.items },
+};
+
+export const WithStyles = {
+  ...Default,
+  args: { ...Default.args, termClass: ['d-fw-bold', 'd-fc-disabled'] },
+};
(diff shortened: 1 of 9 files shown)


components/description_list/description_list.stories.js
```diff
@@ -0,0 +1,123 @@
...
+        summary: '{ term: string, description: string }[]',
+      },
+    },
+  },
+  descriptionClass: {
+    control: 'text',
+  },
+  termClass: {
+    control: 'text',
+  },
```
braddialpad: Hmm because this is set to text the array you are passing into `WithStyles` is not showing up in the field. Maybe we set termClass control to object just for that case. Or pass it in as a string rather than an array.

components/description_list/description_list.vue
```diff
@@ -0,0 +1,116 @@
...
+</script>
+
+<style lang="less">
+.dt-description-list {
+  display: flex;
+  flex-wrap: wrap;
+  align-items: flex-start;
+  gap: var(--dt-space-400);
+  flex-direction: row;
+  line-height: var(--lh2);
```
braddialpad: Hmm can we use `var(--dt-font-line-height-200)`? It is slightly different but I think it still works.

components/description_list/description_list.vue
```diff
@@ -0,0 +1,116 @@
...
+  },
+};
+</script>
+
+<style lang="less">
+.dt-description-list {
+  display: flex;
+  flex-wrap: wrap;
+  align-items: flex-start;
+  gap: var(--dt-space-400);
```
braddialpad: can set gap as a var and the modify the var instead of the gap directly:

```less
.dt-description-list {
  --description-list-gap: var(--dt-space-400);
  gap: var(--description-list-gap);
  
  each(range(0, 600, 100), {
    &--gap-@{value} {
     --description-list-gap: ~"var(--dt-space-@{value})";
    }
  });
}
```

We typically use this sort of structure in dialtone, see button.less for example

components/description_list/description_list.vue
```diff
@@ -0,0 +1,116 @@
...
+    },
+  },
+
+  computed: {
+    getDirectionClass () {
+      return `dt-description-list--${this.direction}`;
+    },
+
+    getGapClass () {
+      return DT_STACK_GAP.includes(this.gap) ? `dt-description-list--gap-${this.gap}` : null;
```
braddialpad: Do you need to do this conditional when you already have a validator for the gap prop?

braddialpad (commented):
Nice work here, very clean and complete! Good call on putting the description list in a resizable div so it's easy to see how it works responsively.

components/description_list/description_list.vue
ninamarina: Changed for `var(--dt-font-line-height-300)`, checking the figma that one is closer to what is used there.

components/description_list/description_list.vue
ninamarina: Nice catch, I removed the conditional, thanks!

components/description_list/description_list.vue
```diff
@@ -0,0 +1,122 @@
...
+        {{ item.term }}
+      </dt>
+      <dd
+        :class="['dt-description-list__description', descriptionClass]"
+      >
+        {{ item.description }}
+      </dd>
+    </template>
+  </dl>
+</template>
```
iropolo: Ok, my 2 cents.
Remove `<!-- eslint-disable vue/require-v-for-key -->`

and update

```
<dt
        :key="`dt-${item.term}`"
        :class="['dt-description-list__term', termClass]"
      >
```

and

```
<dd
        :key="`dd-${item.term}`"
        :class="['dt-description-list__description', descriptionClass]"
      >
```

Is this work for you?

iropolo: As I can see we can also use here `index`, since the items cannot be manipulated (removed, add more or reorderer).

It can be like :key="`dt-${index}`"

components/description_list/description_list.vue
```diff
@@ -0,0 +1,122 @@
...
+    <template
+      v-for="item in items"
+    >
+      <dt
+        :class="['dt-description-list__term', termClass]"
+      >
+        {{ item.term }}
+      </dt>
+      <dd
+        :class="['dt-description-list__description', descriptionClass]"
```
iropolo: Would update to sth like a computed propery `ddClass` and `dtClass`

```
dtClass() {
    return ['dt-description-list__term', this.termClass];
  }
```

Result in: `<dt :class="dtClass">`

components/description_list/description_list.vue
```diff
@@ -0,0 +1,122 @@
...
+    color: var(--dt-color-foreground-primary);
+    flex: 1 1 50%;
+    margin-left: 0;
+  }
+}
+each(range(0, 600, 100), {
+  .dt-description-list--gap-@{value} {
+    gap: ~"var(--dt-space-@{value})";
+  }
+});
```
iropolo: Love this

components/description_list/description_list.test.js
```diff
@@ -0,0 +1,148 @@
...
+  describe('Extendability Tests', () => {
+    const customClass = 'my-custom-class';
+
+    describe('When an term class is provided', () => {
+      it('should apply custom class to child', () => {
+        mockProps = { termClass: customClass };
+
+        updateWrapper();
+
+        expect(wrapper.find('dt').classes().includes(customClass)).toBe(true);
```
iropolo: This is not wrong, but we usually create the element and then separately we do the assertion.
Would be something like:

```
const dtElement = wrapper.find('dt')

expect(dtElement.classes().includes(customClass)).toBe(true);
```

It also can be  
```
const dtElementClassess = wrapper.find('dt').classes()

expect(dtElementClassess.includes(customClass)).toBe(true);
```

In this way the assertion is more readable.

components/description_list/description_list.test.js
```diff
@@ -0,0 +1,148 @@
...
+    term: 'Reason',
+    description: 'Refound',
+  }],
+};
+
+let mockProps = {};
+
+describe('DtDescriptionList Tests', () => {
+  let wrapper;
+  let list;
```
iropolo: this should be a `MOCK_LIST`, since it is not sth that is required on the wrapper in initial mount.
Its more like an auxiliary variable you are using to do your stuff.

To be just `list` I will expect sth like:

```
let wrapper
let list

const updateWrapper = () => {
    wrapper = mount(DtDescriptionList, {
      propsData: { ...baseProps, ...mockProps },
    });

    list = wrapper.find(...);
  };
```

iropolo (approved):
Looks good, nice work here.
pre-approving.
Thanks

components/description_list/description_list.vue
ninamarina: That worked, thanks! The term is supposed to be unique so I think it's fine to go with that one.

iropolo (approved):

components/description_list/description_list.vue
```diff
@@ -0,0 +1,131 @@
...
```
braddialpad: I think we can remove this now that we have the inner one right?

braddialpad (approved):
Just one comment, looks great otherwise!

components/description_list/description_list.vue
ninamarina: yes!

PR #1292, opened 2023-11-02
chore: add $attrs to vue2

braddialpad:
# chore: add $attrs to vue2

- [ ] Fix
- [ ] Feature
- [x] Refactoring
- [ ] Documentation

## :book: Description

Attempting to put all storybook props on $attrs in vue 2 repo. For some reason in vue 2 you have to bind argTypes directly to props instead of just using args. Not sure exactly why this is, maybe a difference in how reactivity works in vue 2 vs vue 3.

Did this with a regex and then fixed a bunch of abnormal cases manually so it's definitely possible there are problems. I have not tested well yet. Going to rely on the visual tests first and go from there.

## :bulb: Context

So we don't have to always update mundane things between vue 2 and vue 3


braddialpad:
Alright, lookin good now. Going to get this in because it's quite conflicty.

PR #1294, opened 2023-11-03
refactor(icon): update dialtone icons usage

juliodialpad:
# Refactor (Icon): Dialtone icons usage

- [ ] Fix
- [ ] Feature
- [x] Refactoring
- [ ] Documentation

## :book: Description

Refactored the usage of dialtone-icons now that icons are just exporting SVG's instead of vue components to avoid having to compile vue 2 and vue 3 library.

The usage of the dtIcon remains the same, it's just the implementation under the hood that changes.

This might also impact performance positively due to the icons are being imported just on usage and not at once. 

## :bulb: Context

Dialtone icons v4 had a [breaking change](https://github.com/dialpad/dialtone/pull/8) that made the library export SVG's instead of Vue components, this change is to update its usage. 

## :pencil: Checklist

- [ ] I have added tests

Screenshot 2023-11-02 at 9 00 13 p m


braddialpad:
Some problems in the visual tests it seems.

package.json
```diff
@@ -114,6 +117,7 @@
   },
   "peerDependencies": {
     "@dialpad/dialtone": "^7.30.0 || ^8.21.2",
+    "@dialpad/dialtone-icons": "^4.0.0",
```
braddialpad, line 120: dialtone-icons is a dep and also a peer dep?

juliodialpad: ~That's a left-over when I was trying to reduce bundle size, I'll remove it~

It is a devDependency and a peerDependency, just as `@dialpad/dialtone`, I think that's the right way of doing it, right?

braddialpad: yeah lets do this for now. I think we can maybe improve this but it'll require more research

PR #1304, opened 2023-11-15
fix(collapsible): use correct transition timing variable

ninamarina:
# fix(collapsible): use correct transition timing variable

Jira ticket: https://dialpad.atlassian.net/browse/DLT-1335

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description
We were using `var(--ttf-quint)` as timing function for the transition which doesn't exist. As a result the transition was not applied, so the `height` property was set on enter, but it was never changed back to null because the transition had never ended.
The fix was to change `var(--ttf-quint)` for `var(--ttf-out-quint)` which I think was the intended variable.
This PR also adds a story to reproduce the bug, and then the fix.

As a side effect of this, the transition is now visible.

## :pencil: Checklist

- [ ] I have added tests

| Before | Now |
|--------|-----|
| collapsible-before | collapsible-now |


components/collapsible/collapsible_lazy_show.vue
```diff
@@ -101,8 +101,6 @@ export default {
         if (!element.style.height) {
           element.style.height = '0px';
         }
-
-        element.style.display = null;
```
ninamarina, line 105: I removed this because it does nothing, the display property is not controlled in the transition.

components/collapsible/collapsible_lazy_show.vue
```diff
@@ -160,6 +158,6 @@ export default {
   .enter-active,
   .leave-active {
     overflow: hidden;
-    transition: height var(--td300) var(--ttf-quint);
+    transition: height var(--td300) var(--ttf-out-quint);
   }
 </style>
```
iropolo, line 163: So the real fix is this?
All the other additions you will remove them before merte this PR, right?

iropolo (approved):
Fix in `components/collapsible/collapsible_lazy_show.vue` looks good to me.

Waiting for vue3 fix.

components/collapsible/collapsible_lazy_show.vue
ninamarina: I was thinking about leaving the story but if it adds no real value I can remove it

iropolo: I dont think so... because I see this as an expected behaviour from the collapsible component

braddialpad (approved):
Change looks good. We may want to re-evaluate this component at some time in the future. It feels overly complex to me, I remember there were some flexibility challenges when we worked on it before.

ninamarina:
> Change looks good. We may want to re-evaluate this component at some time in the future. It feels overly complex to me, I remember there were some flexibility challenges when we worked on it before.

I agree it's overly complicated. Right now it has a transition every time the component is mounted, I don't know if that's intended, so when the component is loaded in an open state it shows the transition. If what we want is to only animate the opening / closing I think it could be done with only css, and maybe a little javascript to set the height.

PR #1306, opened 2023-11-15
feat(message-input): edit message mode using the same recipe 

sanjeevdialpad:
feat(message-input): Edit message mode using the same recipe 

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Introduced new props to message input:
- isEdit : edit mode when true
- cancelButtonAriaLabel - aria label for the button
- cancelButtonText - Button text i18n translated

Events:
- cancel - when cancel button is pressed in edit mode

Screenshot 2023-11-15 at 3 18 07 PM


recipes/conversation_view/message_input/message_input.vue
```diff
@@ -191,19 +213,30 @@
...
                   <dt-icon
                     name="send"
                     size="300"
                   />
                 </template>
+                <template
+                  v-if="isEdit"
+                >
+                  <p>{{ 'Save Changes' }}</p>
+                </template>
```
braddialpad: Did we want to center middle the tooltip in this case now that it is wider?

Screenshot 2023-11-15 at 3 29 03 PM

braddialpad (approved):
Looks good, approving with one question.

recipes/conversation_view/message_input/message_input.vue
```diff
@@ -191,19 +213,30 @@
...
+                >
                   <dt-icon
                     name="send"
                     size="300"
                   />
                 </template>
+                <template
+                  v-if="isEdit"
+                >
+                  <p>{{ 'Save Changes' }}</p>
```
braddialpad: Oh yeah.. this needs to be i18n prop?

sanjeevdialpad: thanks, i missed this.,

recipes/conversation_view/message_input/message_input.vue
sanjeevdialpad: when its edit mode, the tooltip will read save changes, so it still will look wider with that tooltip, hence kept the position same.

recipes/conversation_view/message_input/message_input.vue
```diff
@@ -166,19 +167,40 @@
...
             :show="!isSendDisabled && sendButtonFocus"
-            :offset="[6, -4]"
+            :offset="[6, -8]"
           >
             <template #anchor>
               <!-- Right positioned UI - send button -->
               <dt-button
                 data-qa="dt-message-input-send-btn"
                 size="sm"
                 :kind="!isSendDisabled ? 'default' : 'muted'"
```
iropolo: Will suggest to create a computed property here with the name of `buttonKind`

iropolo:
Percy fails with: a new snapshot and some changes if they are intended, ok, there is a chance where they are flaky so its ok to go.

braddialpad:
> Percy fails with: a new snapshot and some changes if they are intended, ok, there is a chance where they are flaky so its ok to go.

I think this will be fixed when he updates branch from staging-vue3

recipes/conversation_view/message_input/message_input.vue
sanjeevdialpad: done

PR #1308, opened 2023-11-17
feat: upgrade vite to v5 and node to v20

braddialpad:
# feat: upgrade vite to v5 and node to v20

## :hammer_and_wrench: Type Of Change

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

Upgraded vite, node and storybook. Everything seems to be running smoothly.

PR #1309, opened 2023-11-17
feat: upgrade vite to v5 and node to v20

braddialpad:
vue3 of https://github.com/dialpad/dialtone-vue/pull/1308


braddialpad:
hmm 1 unit test error.. looking into it

braddialpad:
Looks like we had vitest locked to a specific version for a reason. Reverting back to original version, appears to fix the issue.

PR #1312, opened 2023-11-18
feat(dp-84718): emit skin tone when selected

cjmollitor-dialpad:
# DP-84718: Emit skin tone when selected from the emoji picker in the message input

- [ ] Fix
- [x] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

On DP2.0, we need to be able to save a user's default skin tone when a new one is selected from the emoji picker. Currently, the message input wraps an emoji picker but doesn't emit an event when a new skin tone is selected. This changes the message input to:
1) Accept a `skinTone` prop so that we can populate the emoji picker's skin tone with the user's default skin tone saved in the backend
2) Emit a `select-skin-tone` event when a new skin tone is selected so we can save it as their default skin tone

## :bulb: Context

N/A

## :pencil: Checklist

- [ ] I have added tests

## :crystal_ball: Next Steps

Hook this up on the DP2 side to save default skin tone

image

## :link: Sources

N/A


recipes/conversation_view/message_input/message_input.vue
```diff
@@ -611,11 +619,18 @@ export default {
...
      */
     'cancel',
+
+    /**
+     * Fires when skin tone is selected from the emoji picker
+     *
+     * @event select-skin-tone
+     * @type {String}
+     */
+    'select-skin-tone',
```
braddialpad: since emoji-picker already emits an event called "skin-tone" should we just use the same event name here?

cjmollitor-dialpad: I went with that at first but for some reason it wasn't working when I was testing it in storybook. It was only when I named it something other than `skin-tone` that it'd properly emit the event. I didn't look into _why_ that was happening too deeply though, does that sound like just some weirdness on the storybook side of things?

juliodialpad: This might be only an issue with the naming, make sure to be listening for `onSkinTone` events on the components that need it, you have a prop called `skinTone` too so that might be confusing, maybe changing the property to `selectedSkinTone` 🤔

braddialpad: I thought maybe it was a conflict between the prop and the event both being named skin-tone so I tried it out here: https://github.com/dialpad/dialtone-vue/pull/1313

Seems to work fine though, please confirm. Might have just been something you missed between the naming in the `stories.js` and the `.story.vue`

cjmollitor-dialpad: Huh that's really weird, it's working completely fine now. I was banging my head on it for like an hour, I could've sworn everything was named properly 🤔 I'll merge this in, thanks Brad!

braddialpad (approved):

PR #1315, opened 2023-11-22
refactor(dp-85018): modify msg input emoji picker pass through props structure

cjmollitor-dialpad:
# DP-85018: Consolidate Message Input's Emoji Picker Passthrough Props into One Object

- [ ] Fix
- [ ] Feature
- [x] Refactoring
- [ ] Documentation

## :book: Description

- Consolidates existing emoji picker pass through props into one object and v-binds it on the emoji picker
- "bubbles up" `selected-emoji` event from emoji picker so message input users can listen to it. Needed in DP2 for updating recent emojis on the server.

## :bulb: Context

N/A

## :pencil: Checklist

- [ ] I have added tests

## :crystal_ball: Next Steps

N/A

N/A

## :link: Sources

N/A


recipes/conversation_view/message_input/message_input.vue
```diff
@@ -455,41 +450,40 @@ export default {
...
+          'Most recently used',
+          'Smileys and people',
+          'Nature',
+          'Food',
+          'Activity',
+          'Travel',
+          'Objects',
+          'Symbols',
+          'Flags',
+        ],
```
braddialpad: We probably don't want to give these default values within the component. We can give them default values on the storybook examples specifically, but when a consumer is using the component they should be "forced" to enter strings for i18n reasons. Otherwise most developers will not notice and just leave it as default.

cjmollitor-dialpad: Good call, I'll make this a required prop and keep the validation for required emoji picker props

braddialpad (approved):
Thanks, looks good. Let me check with Julio if this is okay to merge since we are mid monorepo migration right now.

braddialpad:
@juliodialpad Are we okay to merge this? It's a vue 3 only change. Not sure exactly what point the migration is at right now.

juliodialpad:
> @juliodialpad Are we okay to merge this? It's a vue 3 only change. Not sure exactly what point the migration is at right now.

It's ok to merge, I'll make the appropriate changes on the migration to get this in, is only 3 files, I don't want to block this

PR #1318, opened 2023-11-27
fix: updates necessary for feed-item-row product implementation

braddialpad:
# fix: updates necessary for feed-item-row product implementation

- [x] Fix
- [ ] Feature
- [ ] Refactoring
- [ ] Documentation

## :book: Description

- Remove margin from emoji-text-wrapper it is unnecessary and causes lots of spacing issues.
- Add a slot for the emoji picker, since it is not built into the component itself it is very hard to align the anchor correctly without this.
- A few styling changes to better match the designs.
- Trap focus within the feed-item-row. For keyboard navigation purposes the arrow keys will change the selected item and tabbing will tab through the interactive controls within each item. Tabbing to the end should loop around to the first item again rather than going to the next item.

## :bulb: Context

Couple changes that were needed within dialtone-vue when implementing feed-item-row in product

## :pencil: Checklist

## :crystal_ball: Next Steps

Update in firespotter and release private beta for design review.

Might be some more changes to make later but these are most essential for now.

### DIFF
--- a/components/emoji_text_wrapper/emoji_text_wrapper.vue
+++ b/components/emoji_text_wrapper/emoji_text_wrapper.vue
@@ -56,7 +56,7 @@ export default {
       return split.map((item) => {
         if (replaceList.includes(item)) {
           return this.$createElement(DtEmoji, {
-            attrs: { class: 'd-mx4 d-d-inline-block' },
+            attrs: { class: 'd-d-inline-block' },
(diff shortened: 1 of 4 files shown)


iropolo (approved):
Looks, good, thanks

iropolo:
Before:
image

After:
image

J/M/P are a little bit more close vertically to emoji reactions... right?

